PR #149, opened 2015-03-24
Add owner controls and rival columns for user

BatedUrGonnaDie:
Add owner controls and rivals columns when user is signed in on their own profile page.

### DIFF
--- a/app/views/users/show.haml
+++ b/app/views/users/show.haml
@@ -7,4 +7,7 @@
     - @user.games.named.each do |game|
       = link_to game.shortname || game.name.downcase, game, class: 'game-link stealth-link'
 %article
-  = render partial: 'shared/run_table', locals: table_locals(:pbs, user: @user)
+  - if user_signed_in? && current_user.name == @user.name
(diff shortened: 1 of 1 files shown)


glacials:
Owner-specific stuff is intentionally excluded from the profile page.

It's to avoid confusion about who sees what, and prevent thoughts like "I'm not sure if I want to share this link because I don't want people to know my rivals", or worse, "I'm not sure if I want to set any rivals because it looks like anyone will be able to see them".

So there's this distinction between the frontpage, which is more or less a mission control, and the profile page, which is the public view into your account (even if you're you).

Not convinced we should homogenize the two, but I'm open to being convinced.

BatedUrGonnaDie:
Ah, I didn't realize that and it makes a lot of sense.  Is there any reason that rival information should be kept private though?  If anything I feel like knowing that someone considers me to be a rival would make me want to try and top them more.

glacials:
Yeah good point, that could be nice. Maybe a setting for it.

PR #158, opened 2015-04-25
Add urn support

BatedUrGonnaDie:


lib/urn_parser.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def parse_split(segment, prev_split_finish_time)
+    Split.new(
+      best: Split.new(duration: duration_in_seconds(segment["best_segment"])),
+      name: segment["title"].to_s,
+      finish_time: duration_in_seconds(segment["time"])
+    ).tap do |split|
+      split.duration = [split.finish_time - prev_split_finish_time, 0].max
+      split.gold = split.duration > 0 && split.duration.round(5) == split.best.duration.try(:round, 5)
+      split.skipped = split.duration > 0
```
glacials: Was this meant to be

```ruby
split.skipped = split.duration == 0
```

? I actually don't know how Urn handles skips so I don't know the answer here.

glacials:
When given a non-Urn file, the parser should return `nil` and not error. LiveSplitParser handles this [with a global rescue](https://github.com/glacials/splits-io/blob/master/lib/livesplit_parser.rb#L13), which is generally not a good thing to do, but which works and which I'm okay with UrnParser also doing until I do some `run.rb` refactoring.

Looks mostly good though. Can you add Urn to the text for supported programs on `app/views/runs/cant_parse.haml` and `app/views/pages/faq.slim`?

BatedUrGonnaDie:
Yes, that was supposed to be an == and not a <, sorry about that.  I also added the support text to those 2 file as you asked.

glacials:
Cool :+1: We should add Urn as a download format as well, but that can be done independently. This branch looks good.

Issue #57, opened 2014-08-02
Make a file upload endpoint for convert-only, without storing


glacials:
Something like a POST to `/convert_to/:format` should take a file, convert it to `:format` (e.g. `livesplit`), and immediately serve the converted file back without storing it or creating any page for it.

Would be cool to also have this working for versions, e.g. `POST /convert_to/livesplit/1.3`.

PR #160, opened 2015-04-28
Urn download support

### LINKED ISSUE #159: Urn download support
We can now upload Urn runs, but we should be able to download runs as Urn files as well.

BatedUrGonnaDie:
Closes #159

### DIFF
--- a/app/controllers/runs_controller.rb
+++ b/app/controllers/runs_controller.rb
@@ -12,8 +12,6 @@ class RunsController < ApplicationController
   before_action :attempt_to_claim, only: [:show]
   before_action :verify_ownership, only: [:edit, :update, :destroy]
 
-  before_action :set_filetype, only: [:download]
-
   def show
     @run.delay.refresh_from_file if rand < SplitsIO::Application.config.run_refresh_chance
   end
@@ -56,7 +54,7 @@ class RunsController < ApplicationController
       else
         render_to_string(params[:program], layout: false)
       end,
-      filename: "#{@run.to_param}.#{@filetype}",
+      filename: "#{@run.filename(params[:program])}",
       layout: false
     )
   end
@@ -129,9 +127,5 @@ class RunsController < ApplicationController
       @run.destroy
       redirect_to cant_parse_path
     end
-  end
-
-  def set_filetype
-    @filetype = {'livesplit' => 'lss', 'urn' => 'json'}.fetch(params[:program], params[:program])
   end
 end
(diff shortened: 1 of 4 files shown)


CryZe:
Did you guys figure out whether empty times are null, 0 or an empty string in the Urn splits? Looks like this Pull Request assumes it's 0, but were you guys able to verify that?

BatedUrGonnaDie:
From my understanding it should be a string.  It correctly made my skipped split from a live split file into a skipped split when I opened it in urn.

app/controllers/runs_controller.rb
```diff
@@ -128,4 +130,8 @@ def handle_first_visit
       redirect_to cant_parse_path
     end
   end
+
+  def set_filetype
+    @filetype = {'livesplit' => 'lss', 'urn' => 'json'}.fetch(params[:program], params[:program])
+  end
```
glacials: Instead of putting this as a before_action in the controller, can we put it in a method on the `Run` model? In fact, can we just make it a `filename` method that returns the whole string `"#{to_param}.#{extension}"`? I think this would be cleaner than adding another method down here, and would support future cases where we need to give the file a name in another controller (e.g. #57).

This was a good call refactoring this into its own method, it was getting messy even without adding Urn.

app/views/runs/urn.html.erb
```diff
@@ -0,0 +1,15 @@
...
+    "attempt_count": <%= @run.respond_to?(:attempts) ? @run.attempts : 0 %>,
+    "start_delay": "<%= Time.at(@run.offset || 0).utc.strftime("%-H:%-M:%S").strip %>",
+    "splits": [
+        <% @run.splits.each_with_index do |split, index| %>
+        {
+            "title": "<%= split.name %>",
+            "time": "<%= Time.at(split.finish_time || 0).utc.strftime("%H:%M:%S.%6N").strip %>",
+            "best_time": "<%= Time.at(split.finish_time || 0).utc.strftime("%H:%M:%S.%6N").strip %>",
+            "best_segment": "<%= Time.at(split.best.duration || split.duration).utc.strftime("%H:%M:%S.%6N").strip %>"
+        }<%= ',' unless @run.splits.size - 1 == index %>
```
glacials: Instead of doing this manual logic to see if we need a comma, you should be able to instead do something like (notice the `=`):

```erb
<%= @run.splits.map.with_index do |split, index| %>
{
  ...
}
<% end.join(',') %>
```

(disclaimer: I haven't tested this)

glacials:
Just a couple of small things, on the whole looks good. I'm good to merge if these are addressed.

PR #161, opened 2015-05-01
Urn is not mentioned in some parts of splits i/o

CryZe:


BatedUrGonnaDie:
Did not realize that the programs were listed in so many places, thanks cryze

glacials:
ty cryze :3

i will try to dry this up for apiv4 so that we don't have to change it in so many places next time

PR #165, opened 2015-05-09
Add Video link to Twitch panel export

### LINKED ISSUE #148: Add proof URLs to Twitch panel export
Now that we have video proof in runs (see #107), this'd be great to get in the Export to panels feature as an additional column.

PaulSchnau:
Resolves #148 by adding an addition column "Video" with "Watch" links when there is a proof video for the run.

### DIFF
--- a/app/views/users/pbs/export/panels/index.erb
+++ b/app/views/users/pbs/export/panels/index.erb
@@ -2,8 +2,17 @@
 <% longest[:game_name] = (['game'.length] + @user.pbs.categorized.map { |run| run.game.name.length }).max || 'Game'.length %>
 <% longest[:category_name] = (['category'.length] + @user.pbs.categorized.map { |run| run.category.name.length }).max || 'Category'.length %>
 <% longest[:time] = (['pb'.length] + @user.pbs.categorized.map { |run| "[#{time(run.time)}](#{link(run)})".length }).max || 'PB'.length %>
+<% longest[:video_url] = (['video'.length] + @user.pbs.categorized.map { |run| "[Watch](#{run.video_url})".length }).max || 'Video'.length %>
+<% if @user.pbs.categorized.where.not(video_url: nil).count > 0 %>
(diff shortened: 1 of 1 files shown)


PaulSchnau:
Example output http://pastebin.com/raw.php?i=Qt8JwmVp

glacials:
Thanks Paul! A couple of small things:

1: You can assume that `video_url` is either a URL or `nil`. Technically it's possible for it to be `""` right now, but I have a patch ready to fix that that I can merge ahead of this. So instead of checking these two things and storing in `hasURL`, you can just do something like
```ruby
"link text" unless run.video_url.nil?
```
or
```ruby
run_video_url.nil? ? "link text" : ""
```

2: If a person has no runs with videos, we should hide the entire column. In Markdown it looks fine as blank, but when pasted into a Twitch panel it squishes everything to the left just to display an empty column. You can check this with something like
```ruby
user.pbs.categorized.where.not(video_url: nil).count == 0
```
Thanks for the PR! I think this is good to merge once these two things are addressed.

---
As a possibly unrelated side note, I'm open to suggestions for how to make this file cleaner. It's weird writing ERB for Markdown because you're pretty constricted on where you can put line breaks and other code niceties, so we end up with these really long lines without really any other option. Not quite sure what to do about it other than abstract the hell out of it with helpers, but that's for another time.

PaulSchnau:
Thank you for the useful coding hints. I have now included logic for not displaying the Video column if the user does not have any videos. 

I agree that the inline ruby code is becoming cluttered and is somewhat limiting. I built the if logic in a more vertical pattern, but that involved some code duplication. Maybe we should consider building out  panels_controller.rb to include logic for gathering column, run, and spacing information to be sent to the view.

glacials:
Yeah, it'll probably be some combination of PanelsController doing db legwork (looking up the pbs, figuring out if there are videos, etc.) and PanelsHelper doing view-related legwork (finding longest strings, figuring out which columns to display, generating the `--|:------------:|:` and `nbsp;` stuff, etc.).

This is good for now though. :+1: ty!

PR #194, opened 2015-08-16
Switch forms of binary detection

BatedUrGonnaDie:
ptools was not flagging all llanfair files as binary, filemagic seems to
be the more accurate version of checking file types.


glacials:
^ reverted for now, Heroku won't build this because it doesn't have `libmagic`, investigating buildpack stuff or alternatives

PR #196, opened 2015-08-17
Scan uploads for null bytes

BatedUrGonnaDie:
This takes away the libmagic dependency, and all of my llanfair files
have more than one null byte that doesn't change positions between
files.  Should be more accurate than ptools as well.

### DIFF
--- a/app/models/run_file.rb
+++ b/app/models/run_file.rb
@@ -6,10 +6,11 @@ class RunFile < ActiveRecord::Base
 
   def self.for_file(file)
     if file.respond_to?(:read)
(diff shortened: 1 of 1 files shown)


glacials:
sweet :D

glacials:
This works well, in the future I'd like to see if Llanfair files have a more precise way of being identified so we don't have to seek through the whole file to determine that it's not one, but for now this seems to be the best way

BatedUrGonnaDie:
It looks like the beginning of all llanfair run files are the same, maybe it could detect based on that?

PR #197, opened 2015-08-25
Stop scanning the whole file to check if llanfair

BatedUrGonnaDie:
I tested this with every llanfair file I have (around 20) and it worked
fine.  I assume this come from the object serialization of the run
object.


glacials:
great!!!

BatedUrGonnaDie:
Random side effect of this would be that if someone tries to upload splits from another program in the future that has null bytes it won't be picked up by this, and will raise the ArgumentError when trying to parse them (error is thrown when trying to put null bytes in string column of pg).

PR #207, opened 2016-01-08
Add a convert only api endpoint and basic form

### LINKED ISSUE #57: Make a file upload endpoint for convert-only, without storing

BatedUrGonnaDie:
Closes #57.


spec/controllers/api/v4/converts_controller_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  describe '#create' do
+    context 'from Llanfair' do
+      let(:file) do
+        fixture_file_upload('files/llanfair')
+      end
+      subject { post :create, params: {file: file, program: "splits_io", include_history: "on"} }
+      let(:body) { JSON.parse(subject.body) }
+
+      it "retruns a 200" do
```
glacials: "returns"

app/models/run.rb
```diff
@@ -76,6 +76,34 @@ def parses?
...
+        end && result[:splits].map do |split|
+          split.best
+        end.sum.to_f
+      )
+
+      @parse_cache = (@parse_cache || {}).merge(fast => result)
+      return result
+    end
+    {}
+  end
```
glacials: This is a lot of nearly-duplicate code from `parse`, can we DRY these up?

app/controllers/api/v4/converts_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+        run_file = RunFile.new(file: file_text)
+      end
+    end
+
+    @run = Run.new(run_file: run_file, user: nil)
+    @run.convert(fast: params[:include_history] == "on" ? true : false)
+
+    extension = {'livesplit' => 'lss',
+      'urn' => 'json',
+      'splits_io' => 'json'
```
glacials: Prefer just "json" instead of "splits_io", I don't think this format is official enough to give it a name

glacials: Ah, I see that the param is named `program` though, and `program=json` doesn't make much sense. Maybe `program=none` or `format=json` or something.

app/views/converts/new.html.erb
```diff
@@ -0,0 +1,28 @@
...
+<% end %>
+<article>
+  <form id="convert" action="/api/v4/convert" method="post" enctype="multipart/form-data">
+    <div class="form-group col-sm-4">
+      <input type="file" name="file" />
+    </div>
+    <div class="form-group col-sm-2">
+      <select name="program">
+        <% Run.programs.each do |prgrm| %>
+          <% unless prgrm.to_s.downcase == "llanfair" %>
```
glacials: you can accomplish this by using `(Run.programs - Llanfair).each do |prgrm|`

docs/api_v4.md
```diff
@@ -88,3 +88,11 @@ curl https://splits.io/api/v4/categories/123/runs
 Some timers allow users to specify a "category" or similar field. A Category is a collection of information about the
 type of run performed, more specific than a Game. Each Category belongs to a Game. The definition of a category does not
 hold any ties to SRL. Any number of Categories can be associated with a Game.
+
+## Convert
+```bash
+curl https://splits.io/api/v4/convert -F "program=splits_io;include_history=on;file=@/path/to/file
+```
+When converting between timer formats, the file and program parameters must be included.  If you are converting a livesplit
+file and would like to include the history, then include_history parameter can also be included.  The Splits/io format is
```
glacials: Api::V4::RunsController uses `historic=1` to include history, we should be consistent with this param

app/views/converts/new.html.erb
```diff
@@ -0,0 +1,28 @@
...
+            <option value="<%= prgrm.to_s.downcase %>"><%= prgrm.to_s %></option>
```
glacials: instead of `prgrm.to_s.downcase`, use `prgrm.to_sym`

app/views/converts/new.html.erb
```diff
@@ -0,0 +1,28 @@
+<% title "convert" %>
+<% content_for(:header) do %>
+  <h2>convert splits to other cool formats</h2>
+<% end %>
+<article>
+  <form id="convert" action="/api/v4/convert" method="post" enctype="multipart/form-data">
```
glacials: what does the user receive from submitting this form? is it a downloaded file?

BatedUrGonnaDie: Yes, the form is just a way to interface with the convert endpoint for humans.  I thought it would be cool to include it so that people could go and convert their runs to any format supported by splits-io if the timer they use doesn't have built in conversion features.

glacials: cool i am for it, however if a bad file is uploaded the user will see raw json as an error message.

we should either make this a JS submission so that the responses are handled and shown to the user nicely (e.g. like `app/assets/javascripts/upload.coffee` does) or make a separate controller for the user-facing convert feature so that we render html instead of json.

app/views/converts/new.html.erb
```diff
@@ -0,0 +1,26 @@
...
+    </div>
+    <div class="form-group col-sm-2">
+      <input type="checkbox" name="historic" id="historic" />
+      <label for="historic">Incldue History?</label>
+    </div>
+    <div class="form-group col-sm-2">
+      <input type="submit" value="convert" />
+    </div>
+  </form>
+</article>
```
glacials: we use slim instead of erb for most html templates, can we change this file to use it

BatedUrGonnaDie:
I think this should address everything.

app/views/converts/new.slim
```diff
@@ -0,0 +1,18 @@
...
+      select name="format"
+        - (Run.programs - [Llanfair]).each do |prgrm|
+          option value="#{prgrm.to_sym}"
+            = prgrm.to_s
+        option value="json" JSON
+    .form-group.col-sm-2
+      input type="checkbox" name="historic" id="historic"
+      label for="historic" Include History?
+    .form-group.col-sm-2
+      input type="submit" value="convert" id="convert-submit"
```
glacials: Add `class="btn btn-default"` to this to make it look like the button on `/games`.

app/views/converts/new.slim
```diff
@@ -0,0 +1,18 @@
...
+      input type="file" name="file"
+    .form-group.col-sm-2
+      select name="format"
+        - (Run.programs - [Llanfair]).each do |prgrm|
+          option value="#{prgrm.to_sym}"
+            = prgrm.to_s
+        option value="json" JSON
+    .form-group.col-sm-2
+      input type="checkbox" name="historic" id="historic"
+      label for="historic" Include History?
```
glacials: This checkbox looks squished up against the `I`; can we add a space here

screenshot 2016-01-18 18 56 53

app/controllers/api/v4/converts_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    if params[:format] == "json"
+      new_file = render_to_string(json: @run, serializer: Api::V4::Convert::RunSerializer)
+    else
+      new_file = render_to_string("runs/#{params[:format]}", layout: false)
+    end
+    send_data(
+      new_file,
+      filename: "#{params[:file].original_filename.split(".")[0..-1].join}.#{extension}",
+      layout: false
+    )
```
glacials: If I convert a file named `sms-any%.lss` to e.g. JSON, I get a file named `sms-any%lss.json` back, but I expect to get a file called `sms-any%.json`.

glacials:
Can you also add a link to the convert page from below the form on `/upload` or something, so people can discover it?

app/controllers/converts_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  def create
+    params.require(:file)
+    params.require(:format)
+    run_file = RunFile.for_convert(params[:file])
+    @run = Run.new(run_file: run_file, user: nil)
+    @run.parse(fast: ["on", "1"].include?(params[:historic]) ? false : true, convert: true)
+    unless @run.splits
+      redirect_to cant_parse and return
```
glacials: I roughly follow [bbatsov/ruby-style-guide](https://github.com/bbatsov/ruby-style-guide#no-and-or-or) which says to never use `and`. Just use two lines or `&&`.

BatedUrGonnaDie:
I'm not sure how to style the text to the link and text to the convert page to make it pretty, but I feel like putting it on /upload is natural stepping stone for people looking convert functionality.

glacials:
lgtm

PR #210, opened 2016-01-25
Parse speedrun.com run id if present in LS 1.6 file

BatedUrGonnaDie:


app/models/run.rb
```diff
@@ -93,6 +93,7 @@ def parse(fast: true, convert: false)
         name: result[:name],
         program: result[:program],
         attempts: result[:attempts],
+        srdc_id: result[:srdc_id].blank? ? nil : result[:srdc_id],
```
glacials: This can be simplified with `presence`: http://api.rubyonrails.org/v4.2.5/classes/Object.html#method-i-presence

BatedUrGonnaDie: I did not realize that .present? would return false for a blank string, which was why I chose .blank?.  Will update accordingly.

glacials: This is what I use to remember: http://stackoverflow.com/a/20663389/392225

BatedUrGonnaDie: I can see that table being very useful.  Thanks for the link

PR #211, opened 2016-01-27
Flesh out convert endpoints

BatedUrGonnaDie:
More error checking and handling.

### DIFF
--- a/app/controllers/api/v4/converts_controller.rb
+++ b/app/controllers/api/v4/converts_controller.rb
@@ -6,7 +6,7 @@ class Api::V4::ConvertsController < Api::V4::ApplicationController
     run_file = RunFile.for_convert(params[:file])
 
     @run = Run.new(run_file: run_file, user: nil)
-    unless @run.parses?(fast: !(params[:historic] == "1"), convert: true)
+    unless @run.parses?(fast: params[:historic] != "1", convert: true)
       render status: 400, json: {
         status: 400,
         message: "Can't parse that file. We support #{Run.programs.to_sentence}."
--- a/app/controllers/api/v4/converts_controller.rb
+++ b/app/controllers/api/v4/converts_controller.rb
@@ -7,6 +7,13 @@ class Api::V4::ConvertsController < Api::V4::ApplicationController
 
     @run = Run.new(run_file: run_file, user: nil)
     @run.parse(fast: ["on", "1"].include?(params[:historic]) ? false : true, convert: true)
+    unless @run.splits.present?
+      render status: 400, json: {
+        status: 400,
+        message: "Can't parse that file. We support #{Run.programs.to_sentence}."
+      }
+      return
+    end
 
     program_extensions = {'livesplit' => '.lss', 'urn' => '.json'}
     file_name = "#{params[:file].original_filename.split(program_extensions[@run.program])[0]}"
@@ -36,6 +43,12 @@ class Api::V4::ConvertsController < Api::V4::ApplicationController
         message: "Convert supports the following formats: #{supported}"
       }
     end
+  rescue ActionController::ParameterMissing
+    render status: 400, json: {
+      status: 400,
+      message: "Missing 'file' or 'format' parameter. Make sure to include both in your request."
+    }
+    return
   end
 
 end
--- a/app/controllers/api/v4/converts_controller.rb
+++ b/app/controllers/api/v4/converts_controller.rb
@@ -6,8 +6,7 @@ class Api::V4::ConvertsController < Api::V4::ApplicationController
     run_file = RunFile.for_convert(params[:file])
 
     @run = Run.new(run_file: run_file, user: nil)
-    @run.parse(fast: ["on", "1"].include?(params[:historic]) ? false : true, convert: true)
-    unless @run.splits.present?
+    unless @run.parses?(fast: !(params[:historic] == "1"), convert: true)
       render status: 400, json: {
         status: 400,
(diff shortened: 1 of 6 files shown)


app/controllers/api/v4/converts_controller.rb
```diff
@@ -7,6 +7,13 @@ def create
 
     @run = Run.new(run_file: run_file, user: nil)
     @run.parse(fast: ["on", "1"].include?(params[:historic]) ? false : true, convert: true)
```
glacials: I know this is from before but I'm just noticing this -- `a ? false : true` can be simplified to `!a`. Also I'd prefer supporting only `historic=1` and not `historic=on` in order to simplify the code and have consistency with the other endpoint with `historic`.

app/controllers/api/v4/converts_controller.rb
```diff
@@ -7,6 +7,13 @@ def create
...
+    unless @run.splits.present?
```
glacials: `[].present?` returns `false`, but `[]` is a valid outcome for `@run.splits` after parsing.

app/controllers/converts_controller.rb
```diff
@@ -10,7 +10,7 @@ def create
 
     @run = Run.new(run_file: run_file, user: nil)
     @run.parse(fast: ["on", "1"].include?(params[:historic]) ? false : true, convert: true)
-    unless @run.splits
+    unless @run.splits.present?
```
glacials: same as above

app/controllers/converts_controller.rb
```diff
@@ -28,6 +28,9 @@ def create
       filename: file_name,
       layout: false
     )
+  rescue ActionController::ParameterMissing
+    flash.now[:alert] = "Missing 'file' or 'format' form option."
```
glacials: I'm not familiar with `flash.now`, what does this do differently from `flash[:alert] = 'blah'`?

Also, as a user will I see this error message if I submit a form without choosing a file? If so, the error message should be more friendly.

app/controllers/converts_controller.rb
```diff
@@ -28,6 +28,9 @@ def create
...
+    render action: "new"
```
glacials: `render :new` should work

spec/controllers/api/v4/converts_controller_spec.rb
```diff
@@ -27,14 +27,51 @@
...
+        end
+
+        it 'returns an error body' do
+          expect(body['status']).to be_truthy
+          expect(body['message']).to be_truthy
+        end
+      end
+    end
+
+    context 'from Malformed' do
```
glacials: You don't need these double contexts, you can just do the one `'malformed file'` context at the same level as the Llanfair context.

app/controllers/converts_controller.rb
glacials: Ah I see, so then `flash.now` would be used when we `render` and `flash` would be used when we `redirect` typically. Cool, thanks for explaining, this makes sense.

BatedUrGonnaDie: Did you mean the error message should be nicer in terms of language or more specific?

app/controllers/api/v4/converts_controller.rb
glacials: You can store the result of `@run.parse` and see if it's there after parsing:

```ruby
parse_result = @run.parse(:stuff)
unless parse_result.present?
  # do things
end
```

If that's too verbose you could simplify it to

```ruby
unless @run.parse(:stuff).present?
  # do things
end
```

If you want to be more idiomatic you could even change `Run#parses?` to forward its args on to `Run#parse` so that you could do

```ruby
unless @run.parses?(:stuff)
  # do things
end
```

app/controllers/converts_controller.rb
glacials: Nicer in terms of language. If I'm a user and I accidentally forget to pick a file to convert, I should be told that I didn't pick a file to convert. Being told that I'm missing a 'file' or 'format' form option is a bit cryptic and wordy.

If everything on the page behaves correctly and users aren't messing with anything, it's impossible to be missing a format, so I only care about the friendliness of the missing file message. Just something simple like "You didn't select a file." or something like that.

BatedUrGonnaDie: Yeah, that makes more sense.  Trying to cover things that wouldn't normally happen without a user tinkering with the page is kind of silly in retrospect.

app/controllers/api/v4/converts_controller.rb
```diff
@@ -6,7 +6,13 @@ def create
     run_file = RunFile.for_convert(params[:file])
 
     @run = Run.new(run_file: run_file, user: nil)
-    @run.parse(fast: ["on", "1"].include?(params[:historic]) ? false : true, convert: true)
+    unless @run.parses?(fast: !(params[:historic] == "1"), convert: true)
```
glacials: prefer `params[:historic] != "1"`

app/controllers/converts_controller.rb
```diff
@@ -9,8 +9,7 @@ def create
     run_file = RunFile.for_convert(params[:file])
 
     @run = Run.new(run_file: run_file, user: nil)
-    @run.parse(fast: ["on", "1"].include?(params[:historic]) ? false : true, convert: true)
-    unless @run.splits
+    unless @run.parses?(fast: !(params[:historic] == "1"), convert: true)
```
glacials: same

PR #212, opened 2016-02-07
Fix downloading llanfair files incorrectly

BatedUrGonnaDie:


glacials:
:heart_eyes: tests :heart_eyes:

BatedUrGonnaDie:
I'm trying to be better about adding them with stuff I change :+1:

PR #235, opened 2016-06-30
2 small bugfixes

BatedUrGonnaDie:


app/models/concerns/padawan_run.rb
```diff
@@ -11,7 +11,7 @@ def improvements_towards(better_run)
       return improvements if collapsed_splits.size != better_run.collapsed_splits.size
 
       collapsed_splits.each.with_index do |split, i|
-        if better_run.splits[i].duration < split.duration && !split.reduced?
+        if better_run.splits[i].duration < split.duration && !split.reduced? && (split.duration - better_run.splits[i].duration) > 1
```
glacials: +1 i like this change. However this can be simplified by changing the first check to `(better_run.splits[i].duration + 1) < split.duration` rather than having both checks

app/views/layouts/application.slim
```diff
@@ -46,7 +46,7 @@ html
                   = link_to "tools", tools_path
                   = link_to "sign out", session_path(auth_session), method: :delete
             - else
-              a#sign-in.orange href='/auth/twitch' onclick='document.cookie="return_to="+window.location.pathname;' sign in
+              a#sign-in.orange href='/auth/twitch' onclick='document.cookie="return_to="+window.location.pathname+window.location.search;' sign in
```
glacials: +1 this solves #230

glacials:
Coo

PR #236, opened 2016-07-01
pass absolute dates in run table

### LINKED ISSUE #109: Relative uploaded dates are incorrect
"Uploaded at" dates on HTML run tables are often incorrect because the entire table is cached. Not sure how to fix this, but it's a problem. A simple fix would be to stop using relative dates and use absolute ones, but I'm not sure if I want to do this.

BatedUrGonnaDie:
closes #109

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -72,6 +72,7 @@ gem 'numeraljs-rails'
 gem 'spinjs-rails'
 gem 'tipsy-rails'
 gem 'underscore-rails'
+gem 'momentjs-rails'
 
 # models/db
 gem 'has_secure_token'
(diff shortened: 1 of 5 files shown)


BatedUrGonnaDie:
Let me know if you don't like using moment.js for this, I can work on writing something that formats it natively.  The reason I went with moment is because it has a lot of handy time/date formatting, which can be used for various other things (ie: time formatting).

glacials:
yeah moment is fine. In https://github.com/glacials/splits-io/blob/master/app/assets/javascripts/time.coffee we use `.time.needs-formatting`, can we do `.date.needs-formatting` here for consistency?

BatedUrGonnaDie:
Yeah, totally.  For some reason I thought there would be a name conflict because they would both be using .needs-formatting, but now that you say it I realize that it wouldn't matter because the top level selecter.

glacials:
sweet

PR #242, opened 2016-07-24
Update api docs

BatedUrGonnaDie:
Adds new fields reduced and history to split objects.  Adds new fields
attempts and sum_of_best to top object. Changed output to use the
example url.


glacials:
:tada:

PR #243, opened 2016-08-07
Few small changes

BatedUrGonnaDie:


app/controllers/api/v4/converts_controller.rb
```diff
@@ -33,7 +33,8 @@ def create
   private
 
   def check_parameters
-    params.require(:file, :format)
+    params.require(:file)
+    params.require(:format)
```
glacials: what's this change do?

lib/parsers/livesplit.rb
```diff
@@ -85,6 +85,7 @@ def v1_6(xml, fast)
         raise "Not a proper LiveSplit run file"
       end
 
+      run[:history].unshift(run[:time]) unless fast || run[:time].in?(run[:history])
```
glacials: Long term, faking history entries might get weird since LiveSplit assigns IDs to each attempt history so you can look up the corresponding segment histories, and we probably want to start scraping and returning those IDs at some point. When that happens faking histories will be a no-no. 

What specific problem does this solve? Is there a different way to solve it?

app/controllers/api/v4/converts_controller.rb
BatedUrGonnaDie: The tests were failing because require only takes one parameter.

glacials: Weird, does an array work? [The docs]http://api.rubyonrails.org/classes/ActionController/Parameters.html#method-i-require) say that an array works but their example is of two arguments, so maybe the example is wrong.

BatedUrGonnaDie: I honestly think something might have changed because I know the test used to pass just fine otherwise I would've fixed it when I added it.

lib/parsers/livesplit.rb
BatedUrGonnaDie: I was just trying to standardize the histories a little bit more so that there isn't a loss of information.

app/controllers/api/v4/converts_controller.rb
glacials: Alright, probably some weird Rails 5 thing. Don't worry about it, this is fine.

lib/parsers/livesplit.rb
glacials: Alright, if that's as far as the problem goes and there's not some specific feature that depends on PB being in history, I'd prefer to not do this. If PB isn't in history, then putting it there doesn't necessarily complete the incomplete information, and it likely makes it inaccurate. If real-world history looks like

    0 1 2 3 4 5 6 7

and LiveSplit's account of real-world history looks like

    5 6 7

and we turn it into

    2 5 6 7

then we're misleading people into thinking that that's a contiguous segment of runs when really there could have been any number of runs between the first and the second. The information is *slightly* more complete, but I'd argue much more misleading, considering histories in our API have always been contiguous.

A PB time not being in a history is shitty but simple to understand once you realize LiveSplit didn't always record histories. But a buttload of missing times immediately before and after the first element in history -- and a number of run history entries that's sometimes-but-not-always one element greater than the number of split history entries in the last split -- is not as simple to understand.

BatedUrGonnaDie: Ah, that makes a lot of sense.  I'll remove these tonight.  I didn't think about the implications of other runs missing and agree that it would do more harm then good.  Do you know if LiveSplit has always tracked split history?  If it did would adding all completed runs to the history by going through the split ID's be a better solution?

glacials: Not sure if LiveSplit has always tracked segment history. @CryZe / @wooferzfg?

If it has, then yeah that sounds like a reasonable solution if we can navigate around the edge cases like missed splits and resets.

wooferzfg:
yeah it always has

BatedUrGonnaDie:
Do you by chance have any 1.2 or 1.1 files lying around?

lib/parsers/livesplit.rb
BatedUrGonnaDie: For now I reverted the commit for that change.  It can be a future issue/pr.

glacials: Cool sounds good

PR #253, opened 2016-10-13
2 small bugfixes.

BatedUrGonnaDie:


app/controllers/runs/application_controller.rb
```diff
@@ -6,6 +6,6 @@ def set_run
     gon.run = {id: @run.id, id36: @run.id36, splits: @run.collapsed_splits}
     gon.scale_to = @run.time
   rescue ActionController::UnknownFormat, ActiveRecord::RecordNotFound
-    not_found
```
glacials: Why did this 500? I can't reproduce: https://splits.io/boop/stats

BatedUrGonnaDie: That's a very good question. On my local box it would throw a 500 but it seems to have different results there.  I'll revert all my changes locally and see what happens.

BatedUrGonnaDie: This change has been reverted.

PR #265, opened 2016-12-16
Fix all failing test cases

BatedUrGonnaDie:


app/controllers/api/v4/converts_controller.rb
```diff
@@ -14,7 +14,7 @@ def create
     end
 
     old_extension = Run.program(@run.program).file_extension
-    new_extension = Run.program(params[:format]).file_extension
+    new_extension = Run.program(params[:format]).try(:file_extension) || params[:format]
```
glacials: when does `Run.program` return nil? afaict `check_parameters` takes care of unexpected values

app/controllers/runs_controller.rb
```diff
@@ -39,8 +39,8 @@ def update
       redirect_to edit_run_path(@run), alert: "There was an error saving that data. Please try again."
     end
 
-    if params[@run.id36][:category]
-      @run.update(category: Category.find(params[@run.id36][:category]))
+    if params[@run.id36][:category][:category]
+      @run.update(category: Category.find(params[@run.id36][:category][:category]))
       redirect_to edit_run_path(@run), notice: 'Game/category updated.'
       return
     end
```
glacials: seems like this works on the website, so if this is required to fix a test, i think the test is broken and not the website

app/controllers/api/v4/converts_controller.rb
BatedUrGonnaDie: When the option is json, which does not have a program attached to it.

glacials: ah right, cool. can we make that more clear in the code, i prefer clarity over flexibility

app/controllers/runs_controller.rb
BatedUrGonnaDie: Ah, was trying to make the tests pass assuming they were correct in the past.

app/controllers/api/v4/converts_controller.rb
BatedUrGonnaDie: Makes sense.  I'll try not to do weird tricks in the future.

app/controllers/api/v4/converts_controller.rb
```diff
@@ -14,7 +14,10 @@ def create
     end
 
     old_extension = Run.program(@run.program).file_extension
-    new_extension = Run.program(params[:format]).file_extension
+    if Run.program(params[:format])
+      new_extension = Run.program(params[:format]).file_extension
+    else
+      new_extension = params[:format]
```
glacials: does this work? is there supposed to be an `end` here?

BatedUrGonnaDie: yeah that was a mistake, thought i ran all the tests again before i pushed but it looks like i just ran one folder that didn't include convert.  i ran them all this time for sure though

glacials:
cool. travis ci, you're next

PR #266, opened 2016-12-17
properly delete files from s3 upon run deletion

BatedUrGonnaDie:


glacials:
👍

BatedUrGonnaDie:
as a side note it still seems to leave the meta data in fakes3, not sure if that will be reflected in real s3 or not but the main splits file is removed which i assume is the important bit.

glacials:
Yeah that's fine, I'm not too worried about proper cleanup, there's more
important stuff to worry about.

PR #276, opened 2017-02-09
Use display name throughout the app

### LINKED ISSUE #246: Use Twitch display names
Right now we just use logins which are all lowercase. We should use the capitalization that the user's Twitch account has. We also need to sync these once in a while with Twitch.

BatedUrGonnaDie:
Closes #246

### DIFF
--- a/db/migrate/20170209005256_add_display_name_to_users.rb
+++ b/db/migrate/20170209005256_add_display_name_to_users.rb
@@ -1,0 +1,5 @@
+class AddDisplayNameToUsers < ActiveRecord::Migration[5.0]
+  def change
+    add_column :users, :display_name, :string
+  end
+end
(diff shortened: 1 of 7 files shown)


db/migrate/20170209005256_add_display_name_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddDisplayNameToUsers < ActiveRecord::Migration[5.0]
+  def change
+    add_column :users, :display_name, :string
```
glacials: Can we call this `twitch_display_name`? I regret calling `name` `name`, as adding other sign in methods later just makes that confusing.

BatedUrGonnaDie:
Looks like to_s was already overridden, so adding display_name onto the front should be ok, yes?

glacials:
Yeah, for sure. Also, can you make sure to test with Chinese, Japanese, or
Korean characters? These are allowed in Twitch display names, so just want
to make sure PostgreSQL handles them fine.

BatedUrGonnaDie:
chrome_2017-02-08_22-40-56

Works ok with everyone's favorite toobou

PR #280, opened 2017-04-13
Travis-CI integration working

BatedUrGonnaDie:


BatedUrGonnaDie:
As a heads up if you do accept this your DATABASE_URL in the travis-ci config is incorrectly set to dev instead of test.

glacials:
Yo sick. Although I just tried changing DATABASE_URL and rebuilding this PR and it still failed. https://travis-ci.org/glacials/splits-io/builds/221680885

BatedUrGonnaDie:
So the AWS stuff are marked as secure so they are removed from the build process (Line 132 and 133) for pull requests.  Since the test AWS credentials are just stubs you could unsecure them, or there might be a setting to allow them in pull requests.  Here is the build from my test https://travis-ci.org/BatedUrGonnaDie/splits-io (although i just noticed that they are secure in mine, but the values i used were taken straight from the application_config.yml.example

glacials:
Ok interesting. If it's just the fact that it's a PR causing it to fail then I'll go ahead and merge, we can solve that later. Just wanted to make sure it was behaving as expected. Let's try this out!

Issue #275, opened 2017-02-03
Darkmode/Lightmode preference should be saved to account


jhobz:
Unless you think there are specific use cases for using a different mode on different devices.

glacials:
Closed by #281.

Issue #255, opened 2016-10-18
Stats graphs need dark mode styles


jhobz:
Just the text is probably minimum, although the graph lines would be nice as well.

![image](https://cloud.githubusercontent.com/assets/1765799/19463722/6c7bd66e-94c4-11e6-8aea-60df18a334f3.png)

glacials:
Thanks, +1 reason for #221.

jhobz:
Tooltips will probably need to be updated with #221 as well:
![image](https://cloud.githubusercontent.com/assets/1765799/19499681/b1dd660e-9565-11e6-89a8-8adbebf74c69.png)

BatedUrGonnaDie:
Should be fixed with permadarkmode.

PR #281, opened 2017-04-14
Redesign + permadarkmode

### LINKED ISSUE #221: Refactor darkmode
As shown in #219 darkmode is hard to program around because you don't really remember that it exists, so when you make a new feature it's easy for that feature to look bad in darkmode.

Options:
1. Gather darkmode usage data and maybe experiment with making it default & removing lightmode
2. Refactor styles in a way that puts lightmode & darkmode style definitions next to each other, making it super obvious that it exists
3. Make testing things in lightmode and darkmode locally easy and obvious (is this possible?)
4. Programmatic darkmode (not convinced this will ever look good)
5. Don't care about it and don't do anything

glacials:
Giving a chance for people to bring up things to either validate or destroy my hesitance merging this. This branch will close #221 (along with #255 and #275 incidentally). The branch is live at [beta.splits.io][1]:

[1]: https://beta.splits.io/

splits i o hello

This is the Superhero bootswatch theme. There are still many minor things I want to clean up (not happy with the reds/oranges on this background for instance) but I've been tweaking and toying with this for a couple of weeks and am reaching the point where I want to just get it out there, knowing my own habits with delaying branches indefinitely and the consequences that follow.

Comparison with current live site theme:

splits i o hello 2

The only thing on trial here is the theme itself, there are some other things changed (like the following list) purely functionally which can adapt to any theme so I'm not concerned with those right now.


CryZe:
The split bar colors need to be changed for this. They don't seem to fit the dark background well. Also the font is kind of small in some places. Overall I like this change though :)

CryZe:
This is what it looks like after fiddling around with the CSS a bit:

glacials:
Cool, thanks for the feedback. I did change the timeline colors a little bit (they're more blue) but I agree it still looks weird.

For the rest of the colors, just to be transparent about this: I prefer very heavily to go with an out-of-the-box Bootstrap theme. There is loads less maintenance if we avoid tweaking individual colors to our liking. I'm super open to any of the ones here: https://bootswatch.com/ or really anywhere on the internet. I'd rather have the site look 20% worse in exchange for not having any design maintenance work get in the way of features.

glacials:
Merged and deployed the general changes, but went with a different (still dark) Bootswatch theme based on comments here and elsewhere. Merged doesn't mean closed for discussion, so still open to opinions.

BatedUrGonnaDie:
I personally like the changed theme better than the original, the text seems much easier to read and the nav bar looks cleaner as well.

PR #282, opened 2017-04-24
Add env variables to travis.yml

BatedUrGonnaDie:
Also use quiet option in fake-s3, which silences some output.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,15 @@ language:
 cache: bundler
 rvm:
  - 2.3.1
+env:
+  AWS_ACCESS_KEY_ID=boop
+  AWS_REGION=local
+  AWS_SECRET_KEY=beep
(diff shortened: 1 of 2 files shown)


BatedUrGonnaDie:
As a side note, I have a branch of fakes3 that respects the -q option for 404 file not founds that cleans up the logs quite a bit.  Not sure if you wanna use a non-source for it but you can see the build log here https://travis-ci.org/BatedUrGonnaDie/splits-io/builds/224505318 .

glacials:
Cool, thanks!

For your fakes3 branch: I'd rather not switch to a fork for something that small because running off a fork will be more work in the future to stay up to date, but I recommend you PR the original project because that seems like a good change.

BatedUrGonnaDie:
I figured as much.  I already introduced a PR for that but it seems like there hasn't been much activity in the main repo for some months now.

PR #283, opened 2017-04-27
Add total playtime to the stats page

### LINKED ISSUE #185: Stats: Total life playtime
See #38.

BatedUrGonnaDie:
Would close #185

### DIFF
--- a/app/assets/javascripts/segment_history_graph.coffee
+++ b/app/assets/javascripts/segment_history_graph.coffee
@@ -15,7 +15,7 @@ $ ->
   hidden_data = []
   first_non_null_attempt = Infinity
 
-  $.each gon.run.splits, (i, split) ->
+  $.each gon.run.raw_splits, (i, split) ->
     non_null_time_inserted = false
     attempt_durations = []
     $.each split.history, (j, attempt) ->
--- a/app/assets/javascripts/reset_graph.coffee
+++ b/app/assets/javascripts/reset_graph.coffee
@@ -13,7 +13,7 @@ $ ->
 
   reset_counter = 0
   reset_data = []
-  $.each gon.run.splits, (i, segment) ->
+  $.each gon.run.raw_splits, (i, segment) ->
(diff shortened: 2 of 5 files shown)


app/assets/javascripts/segment_history_graph.coffee
```diff
@@ -15,7 +15,7 @@ $ ->
   hidden_data = []
   first_non_null_attempt = Infinity
 
-  $.each gon.run.splits, (i, split) ->
+  $.each gon.run.raw_splits, (i, split) ->
```
glacials: What's this change for/do?

BatedUrGonnaDie: gon.run.raw_splits is supposed to have the complete splits with the history.  When I moved the full_splits loop in reset_graph.js it broke all of the splits due to trying to .filter a null.

glacials: I don't understand why it broke, from what I can tell the old `full_segments` code in StatsController produces the same result as the new `segments_with_history` code in CompletedRun, so the swap you did on StatsController line 24 (now 19) shouldn't cause anything that uses `gon.run.raw_splits` to notice a difference. Am I missing something?

BatedUrGonnaDie: I'm not exactly sure why it changes what populates gon.run.splits to be quite honest.  Checking the contents of it pre-commits shows that the history is present in gon.run.splits, but when the logic is moved out of the controller it no longer puts the history in gon.run.splits.  I'm not sure if there is a call to some function somewhere that I'm just missing.

glacials: ooo, yeah that's it :) nice catch! It was incorrect to begin with.

```
$ irb
>> b = [1, 2, 3]
=> [1, 2, 3]
>> a = b
=> [1, 2, 3]
>> a[0] = 2
=> 2
>> b
=> [2, 2, 3]
```

PR #284, opened 2017-04-29
Prevent .sum from returning nil on empty arrays

BatedUrGonnaDie:


app/models/concerns/completed_run.rb
```diff
@@ -67,7 +67,7 @@ def total_playtime
       time = 0
       return time unless program == "livesplit"
       segments_with_history.each do |segment|
-        time += segment.history.sum { |h| h[:duration_seconds].nil? ? 0 : h[:duration_seconds] }
+        time += segment.history.sum { |h| h[:duration_seconds].blank? ? 0 : h[:duration_seconds] } || 0
```
glacials: Fix seems logical, but this line is getting to be a mouthful. Can we make this more readable somehow?

BatedUrGonnaDie: Can do, let me know if this next change is more readable.

PR #286, opened 2017-05-01
Return nil instead of raising error

BatedUrGonnaDie:
Runs will redirect to cant_parse_path instead of 500'ing


glacials:
Thanks for catching this!

PR #288, opened 2017-05-03
Update migrations to remove depreciation warning

BatedUrGonnaDie:
Adds target rails version identifiers to migrations.


glacials:
Interesting, never seen this syntax before. Can you point me to docs for this?

BatedUrGonnaDie:
The best explanation I can find is this blog post on it: http://blog.bigbinary.com/2016/03/01/migrations-are-versioned-in-rails-5.html .  If you look into the past builds on travis though you can see the depreciation warnings for each migration that doesn't have the version identifier.

BatedUrGonnaDie:
All new migrations generated will have a [5.0] added to the end.

glacials:
Cool, that's good enough for me. Thanks for caring about all this stuff, it's super nice getting these deprecation warnings killed!

PR #289, opened 2017-05-16
Don't show segment history download if attempts are 0

BatedUrGonnaDie:


glacials:
merge at your convenience 👍

PR #294, opened 2017-07-07
Become an OAuth provider

### LINKED ISSUE #227: Become an OAuth provider
#### Problem

The current [claim token system](https://github.com/glacials/splits-io/blob/master/docs/api.md#post-runs) is not dependable. For a third party program to upload a run on behalf of a user, it needs to send the user there in a web browser after uploading. If it does not, the run is unowned.

This is wonky because it depends
1. on the program correctly sending the user there,
2. on the web browser correctly opening,
3. on Splits I/O correctly loading the page,
4. on the user being logged in to Splits I/O already.

Even if there are no issues with these things, it's still awkward that the user is required to open the page, which they may not want (e.g. when uploading splits en masse). This also makes it impossible to upload splits for a user in a headless process, like a background job on an SRDC server.
#### Solution

"Upload a run" and "claim a run" should be one atomic action, not two actions that can fail independently.

We should probably become an OAuth provider so that users can authorize applications to upload splits (or optionally do other things) on their behalf. Potential gem to allow this is [doorkeeper-gem/doorkeeper](https://github.com/doorkeeper-gem/doorkeeper).

glacials:
Become an OAuth provider, letting other programs (like timers) request specific permissions from users (like uploading runs to their account) then receive a token they can use to do just that.

Very high level overview:

- Lets users create and delete applications from their settings page
- Lets users authorize and deauthorize applications to view who they are on Splits I/O (no scope) and upload runs to their account (`upload_runs` scope) by going through an OAuth flow
- Lets API requests to the v3 and v4 run upload endpoints be authenticated with an OAuth token, which will cause the created run to be owned by the user who granted that token

Closes #227.

screenshot 2017-07-07 02 39 30screenshot 2017-07-07 02 41 25screenshot 2017-07-07 02 44 38

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -368,9 +368,9 @@ on their behalf. If they accept, you will receive an OAuth token which you can i
 order to create the run as that user.
 
 The following instructions go into naive-case details about implementing this OAuth support in your application. If you
-want to learn more about OAuth or need general OAuth troubleshooting help, you can research OAuth2 online. Especially if
-your application is a website, it's likely that the language you're using has well-established libraries that handle
-much of the below OAuth flow for you.
+want to learn more about OAuth or need general OAuth troubleshooting help, you can [research OAuth2
+online][oauth2-simplified]. Especially if your application is a website, it's likely that the language you're using has
+well-established libraries that handle much of the below OAuth flow for you.
 
 In all cases, you'll need to first go to your Splits I/O account's [settings page][1] and create an application, then
 refer to the relevant section below.
@@ -379,6 +379,8 @@ refer to the relevant section below.
 ```http
 GET https://splits.io/oauth/token/info?access_token=YOUR_TOKEN
 ```
+
+[oauth2-simplified]: https://aaronparecki.com/oauth-2-simplified/
 
 #### Example 1: My application is a local program that runs on the user's computer
 If your application runs locally as a program on a user's computer, you should use OAuth's **authorization code grant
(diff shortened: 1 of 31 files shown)


docs/api.md
```diff
@@ -1,6 +1,18 @@
 # splits i/o API
 These docs are for the v3 API.
 
+## Jump to section
+
+### [Game Endpoints][game-endpoints]
```
BatedUrGonnaDie: The anchors here seem to be pointing at themselves for me.

docs/api.md
```diff
@@ -334,3 +346,212 @@ curl -iX POST --form file=@/path/to/splits_file.lss https://splits.io/api/v3/run
...
+      "expires_in": 7200,
+      "refresh_token": "0e23b095e0d0104ff0642cde71b61d236f3b3865734a0e734714ecd45b25106c",
+      "scope": "upload_run",
+      "created_at": 1499314941
+    }
+    ```
+    Success! `access_token` is your OAuth token for the user. Use it in your API requests to act on behalf of the user
+    by including this header in your requests:
+    ```http
+    Bearer: OAuth YOUR_ACCESS_TOKEN
```
BatedUrGonnaDie: In the api v3 and v4 code it checks for the "Authorization" header, should this be "Authorization" as well or does it check bearer automatically?

BatedUrGonnaDie:
Other than the questions about the docs, looks good to me with the 2 new commits :+1:

docs/api.md
glacials: Oops, yeah this is a typo, should be `Authorization: Bearer <token>`, not `Bearer: OAuth <token>`. Nice catch

glacials:
Found one bug where users can't revoke authorizations, will fix later today then merge this.

PR #295, opened 2017-07-11
Allow claiming anonymous runs

glacials:
Allows users to self-claim runs they uploaded while logged out, which is the most common support request I get from people that I have to fulfill manually.

Somehow I had the foresight 2 years ago to add [a single line of code][1] that records claim tokens in localStorage when a run is uploaded, but simultaneously the laziness to waste two years before putting it to any use. So theoretically you can use this feature to claim runs uploaded anonymously ever since.

Would be nice to also have a page that lists claimable runs, but that's more complex.

screenshot 2017-07-11 00 46 36

[1]: https://github.com/glacials/splits-io/commit/191af4635d3d71f4f6cfe8e198ac67e3db676e99#diff-471b2f654e508c40fc08adf4ea19ea00R13

### DIFF
--- a/app/assets/javascripts/run_claim.coffee
+++ b/app/assets/javascripts/run_claim.coffee
@@ -11,7 +11,7 @@
 activateClaimLink = (claimToken) ->
   if gon.user != null
     claimLink = document.getElementById('claim-nav-link')
-    claimLink.href = window.location.pathname + '?claim_token=' + claimToken
+    claimLink.href = '/' + gon.run.id + '?claim_token=' + claimToken
 
   claimLinkContainer = document.getElementById('claim-nav-link-container')
   claimLinkContainer.style.display = 'inline'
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/run_claim.coffee
```diff
@@ -0,0 +1,51 @@
...
+activateClaimPrompt = (claimToken) ->
+  if gon.user != null
+    claimPromptButton = document.getElementById('claim-prompt-button')
+    claimPromptButton.href = window.location.pathname + '?claim_token=' + claimToken
+
+  claimPrompt = document.getElementById('claim-prompt')
+  claimPrompt.style.display = 'block'
+
+$ ->
+  claimTokenKey = 'claim_tokens' + window.location.pathname
```
BatedUrGonnaDie: A check for gon.run being defined solves the errors I found.  L34 will error on pages where it isn't defined, and subsequently L39 will error on pages that don't have the 'dismiss-claim-prompt' id present on the page.

BatedUrGonnaDie:
Everything looks good to me now.

PR #301, opened 2017-07-29
Add gametime support

### LINKED ISSUE #146: Show in-game times
There should be a toggle on the run page to switch between RTA and game time (if game time is present).

### LINKED ISSUE #261: Can only upload RTA times?
The run I'd like to upload uses LiveSplit under Game Time format, but no matter what I do the site only wants to read the RTA section of the file (even if the RTA section is empty and the SS section has complete splits). Yes, the .lss is already associated with the "SS" catagory, so that isn't the problem.

glacials:
Add support for gametimes. Closes #146 and closes #261.

screenshot 2017-07-29 12 07 07
screenshot 2017-07-29 12 15 18
screenshot 2017-07-29 12 07 40

This was a big change. High-level overview:

- All time-holding db columns (`start_milliseconds`) have been split into two (`realtime_start_ms` and `gametime_start_ms`).
- All functions that operate on or return times (like `shortest_segment`) must now be passed a `timing` argument which can be `Run::REAL` or `Run::GAME`.
- Generally, all places which call said functions will use `params[:timing] || @run.default_timing` (a new column, editable from the run edit page).

This diff is long because I forced myself to add tests for anything that broke in development but didn't trigger a corresponding test failure, and because there are just a lot of function calls to edit in small ways. Feel free to skip over both those types of things.


bepstein111:
Hot damn thank you so much. And thanks for showing off my split titles for DH =)

bepstein111:
Is it live yet? Do I need to reupload? I tried re-parsing, but the toggle switch still isn't there on my runs. Will try reuploading in a few hours when I get to my gaming PC.

glacials:
Yep, it's live as of ~15 minutes ago (had to work through some deploy bugs). A reparse should activate the toggle switch for any run with game times recorded for all segments. Example: https://splits.io/1n2p

If this isn't working for you, can you link me the run?

bepstein111:
Got it. The first time I tried re-parsing, it was on my phone, so you might want to double check compatibility with mobile. When I tried it on a PC, it worked fine. link: https://splits.io/1ej9/

Thanks for all your hard work!

glacials:
No prob, if there's anything you notice that can be better about it let me know, as I don't currently run any games that use game time so I don't have first-hand knowledge of what works well and what doesn't.

PR #304, opened 2017-08-10
Use twitch api v5 and twitch ids instead of usernames

### LINKED ISSUE #302: Move Twitch api calls to v5

BatedUrGonnaDie:
Closes #302

### DIFF
--- a/app/models/concerns/twitch_user.rb
+++ b/app/models/concerns/twitch_user.rb
@@ -12,7 +12,7 @@ module TwitchUser
     end
 
     def twitch_sync!
-      body = Twitch::User.get(name)
+      body = Twitch::User.get(self)
 
       basic_info = JSON.parse(body)
 
--- a/app/models/concerns/twitch_user.rb
+++ b/app/models/concerns/twitch_user.rb
@@ -12,13 +12,14 @@ module TwitchUser
     end
 
     def twitch_sync!
-      body = Twitch::User.get(self)
+      body = Twitch::User.get(twitch_id)
 
       basic_info = JSON.parse(body)
 
       twitch_id = basic_info['_id']
       name = basic_info['name']
       avatar = basic_info['logo']
+      display_name = basic_info['display_name']
 
       if avatar.nil?
         avatar = 'https://static-cdn.jtvnw.net/jtv_user_pictures/xarth/404_user_150x150.png'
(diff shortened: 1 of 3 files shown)


app/models/concerns/twitch_user.rb
```diff
@@ -12,7 +12,7 @@ def follows
     end
 
     def twitch_sync!
-      body = Twitch::User.get(name)
+      body = Twitch::User.get(self)
```
glacials: For code in `lib/` it's generally good to keep inputs super simple (ids rather than whole ActiveRecord objects) so that the library stays hyperfocused on abstracting the communication part and doesn't accidentally take on other responsibilities.

app/models/concerns/twitch_user.rb
```diff
@@ -12,7 +12,7 @@ def follows
...
 
       basic_info = JSON.parse(body)
```
glacials: Can you also add `twitch_display_name` to the attributes we update in this section? I just realized we don't do it here but we do in `SessionsController#create`.

app/models/concerns/twitch_user.rb
BatedUrGonnaDie: Gotcha, that would explain why it passed name before instead of a User object.

lib/twitch.rb
```diff
@@ -33,7 +33,7 @@ def self.followed_ids(login)
 
     class << self
       def get(login)
-        route(login).get(Twitch.headers)
+        route(login.twitch_id).get(Twitch.headers)
```
glacials: Is this right? Also can we change `login` to `id` if these are all IDs now?

BatedUrGonnaDie: Should I change that to only pass id as well?  That one was passing a user object in from the beginning, and I didn't change how the function was called.  Previously it would implicitly .to_s the user object.  The only place I could see it called was from twitch_user.rb on line 10.

glacials: Wow haha ok that was totally messed up then, even though it technically still worked. Yeah if anything is a user object but is called `login` something is definitely amiss. Let's change to ID.

BatedUrGonnaDie: Okie dokie, I'll change the calls to just pass in the twitch id, and then in the functions rename the parameters to just id.

BatedUrGonnaDie:
No problem.  A side effect of this change should fix the unicode errors for users that have non ascii display names since it's no longer trying to call .to_s on a user object which defaults to display_name since I added that in the past.

PR #308, opened 2017-08-26
Fix uncaught sum_of_best call and add test

BatedUrGonnaDie:


app/serializers/api/v3/run_serializer.rb
```diff
@@ -7,7 +7,7 @@ class Api::V3::RunSerializer < Api::V3::ApplicationSerializer
   attributes :id, :path, :name, :program, :image_url, :created_at, :updated_at, :video_url, :attempts, :sum_of_best
 
   def sum_of_best
-    object.sum_of_best.to_f
+    object.realtime_sum_of_best_s.to_f
```
glacials: can we use _ms/1000 instead? been wanting to eventually get rid of the _s calls so we can delete that column

BatedUrGonnaDie: Sure thing.  For future reference are all seconds methods to be considered depreciated?

glacials: Yeah.

PR #309, opened 2017-08-27
Add ability to specify accept header to retrieve runs from api

### LINKED ISSUE #74: Download runs via API
It'd be cool to be able to specify a `Content-Type` to get runs in a certain format via hits to `GET splits.io/api/v2/runs/:id`, so that you don't have to read in `path` and hit `GET splits.io + path + /download/livesplit` or whatever

BatedUrGonnaDie:
Closes #74 .  Can also depreciate the CORS headers on `/:run_id/download/:timer`.

### DIFF
--- a/app/controllers/api/v4/runs_controller.rb
+++ b/app/controllers/api/v4/runs_controller.rb
@@ -2,6 +2,8 @@ class Api::V4::RunsController < Api::V4::ApplicationController
   before_action :set_run, only: [:show, :update, :destroy]
   before_action :verify_ownership!, only: [:update, :destroy]
 
+  before_action :find_accept_header, only: [:show]
+
   before_action :set_link_headers, if: -> { @run.present? }
 
   before_action only: [:create] do
@@ -14,12 +16,8 @@ class Api::V4::RunsController < Api::V4::ApplicationController
   end
 
   def show
-    accept = request.headers.fetch('HTTP_ACCEPT', 'json').downcase
-    if accept == "original_timer"
-      accept = @run.program
-    end
-    timer = Run.program(accept)
-    if accept == 'json' || timer.nil? || !Run.exportable_programs.include?(timer)
+    timer = Run.program_from_content_type(@accept_header)
+    if timer.nil?
       if params[:historic] == '1'
         @run.parse(fast: false)
         render json: @run, serializer: Api::V4::RunWithHistorySerializer
@@ -27,14 +25,11 @@ class Api::V4::RunsController < Api::V4::ApplicationController
         render json: @run, serializer: Api::V4::RunSerializer
       end
     else
+      rendered_run = render_run_to_string(timer)
       send_data(
-        if timer == Run.program(@run.timer)
-          @run.file
-        else
-          render_to_string(file: "runs/#{accept}.html.erb", layout: false)
-        end,
+        rendered_run,
         layout: false,
-        type: "application/#{accept}",
+        type: "#{@accept_header}",
         disposition: 'inline',
         filename: "#{@run.filename(timer: timer)}"
       )
@@ -110,6 +105,30 @@ class Api::V4::RunsController < Api::V4::ApplicationController
        "#{headers['Link']}, #{links}"
     else
       links
+    end
+  end
+
+  def find_accept_header
+    @accept_header = request.headers.fetch('HTTP_ACCEPT', 'application/json').downcase
+    @accept_header = 'application/json' if @accept_header.blank?
+    @accept_header = 'application/json' if @accept_header.include?('text/html')
+    @accept_header = "#{Run.program(@run.timer).content_type}" if @accept_header == 'application/original-timer'
+    valid_accepts = Run.exportable_programs.map(&:content_type) << 'application/json'
+    unless valid_accepts.include?(@accept_header)
+      render status: 406, json: {
+        status: 406,
+        error: "Accept mime type #{@accept_header} not supported",
+        valid_accepts: valid_accepts
+      }
+      return
+    end
+  end
+
+  def render_run_to_string(timer)
+    if timer == Run.program(@run.timer)
+      @run.file
+    else
+      render_to_string(file: "runs/#{@run.program}.html.erb", layout: false)
     end
   end
 
--- a/app/controllers/api/v4/runs_controller.rb
+++ b/app/controllers/api/v4/runs_controller.rb
@@ -14,11 +14,27 @@ class Api::V4::RunsController < Api::V4::ApplicationController
   end
 
   def show
-    if params[:historic] == '1'
-      @run.parse(fast: false)
-      render json: @run, serializer: Api::V4::RunWithHistorySerializer
+    accept = request.headers.fetch('HTTP_ACCEPT', 'json').downcase
+    timer = Run.program(accept)
+    if accept == 'json' || timer.nil? || !Run.exportable_programs.include?(timer)
+      if params[:historic] == '1'
+        @run.parse(fast: false)
+        render json: @run, serializer: Api::V4::RunWithHistorySerializer
+      else
+        render json: @run, serializer: Api::V4::RunSerializer
+      end
     else
-      render json: @run, serializer: Api::V4::RunSerializer
+      send_data(
+        if timer == Run.program(@run.timer)
+          @run.file
+        else
+          render_to_string(file: "runs/#{accept}.html.erb", layout: false)
+        end,
+        layout: false,
+        type: "application/#{accept}",
+        disposition: 'inline',
+        filename: "#{@run.filename(timer: timer)}"
+      )
     end
   end
 
--- a/app/controllers/api/v4/runs_controller.rb
+++ b/app/controllers/api/v4/runs_controller.rb
@@ -128,7 +128,7 @@ class Api::V4::RunsController < Api::V4::ApplicationController
     if timer == Run.program(@run.timer)
       @run.file
     else
-      render_to_string(file: "runs/#{@run.program}.html.erb", layout: false)
+      render_to_string(file: "runs/#{timer.to_sym}.html.erb", layout: false)
     end
   end
 
--- a/app/controllers/api/v4/runs_controller.rb
+++ b/app/controllers/api/v4/runs_controller.rb
@@ -16,7 +16,8 @@ class Api::V4::RunsController < Api::V4::ApplicationController
   def show
     accept = request.headers.fetch('HTTP_ACCEPT', 'json').downcase
     timer = Run.program(accept)
-    if accept == 'json' || timer.nil? || !Run.exportable_programs.include?(timer)
+    json_out = accept == 'json' || timer.nil? || !Run.exportable_programs.include?(timer)
+    if json_out && accept != "original_timer"
       if params[:historic] == '1'
         @run.parse(fast: false)
         render json: @run, serializer: Api::V4::RunWithHistorySerializer
@@ -25,7 +26,7 @@ class Api::V4::RunsController < Api::V4::ApplicationController
       end
     else
       send_data(
-        if timer == Run.program(@run.timer)
+        if timer == Run.program(@run.timer) || accept == 'original_timer'
           @run.file
         else
           render_to_string(file: "runs/#{accept}.html.erb", layout: false)
--- a/app/controllers/api/v4/runs_controller.rb
+++ b/app/controllers/api/v4/runs_controller.rb
@@ -15,9 +15,11 @@ class Api::V4::RunsController < Api::V4::ApplicationController
 
   def show
     accept = request.headers.fetch('HTTP_ACCEPT', 'json').downcase
+    if accept == "original_timer"
+      accept = @run.program
+    end
     timer = Run.program(accept)
-    json_out = accept == 'json' || timer.nil? || !Run.exportable_programs.include?(timer)
-    if json_out && accept != "original_timer"
+    if accept == 'json' || timer.nil? || !Run.exportable_programs.include?(timer)
       if params[:historic] == '1'
         @run.parse(fast: false)
         render json: @run, serializer: Api::V4::RunWithHistorySerializer
@@ -26,7 +28,7 @@ class Api::V4::RunsController < Api::V4::ApplicationController
       end
     else
       send_data(
-        if timer == Run.program(@run.timer) || accept == 'original_timer'
+        if timer == Run.program(@run.timer)
           @run.file
         else
           render_to_string(file: "runs/#{accept}.html.erb", layout: false)
--- a/docs/api_v4.md
+++ b/docs/api_v4.md
@@ -46,10 +46,22 @@ If a `historic=1` param is included in the request, one additional field will be
 |-------------:|:---------------------------------------------|:------------------------------------------------------|:----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `history`    | array of numbers                             | never                                                 | Ordered durations of all previous runs. The first item is the first run recorded by the runner's timer into the source file. The last item is the most recent one. This field is only nonempty if the source timer records history.           |
 
-If an `Accept` header is present, splits.io will try to render the run file natively not in JSON of the type specified.  The format for the header is `application/timer`, with timer being any of the timers that can be exported by splits.io.
-If the `Accept` header is valid, the `Content-type` header in the response will be in the format `application/timer`, with timer being the specified timer.  If the `Accept` header is not valid, the `Content-type` header of the response
-will be `application/json`, and the status code will be a 406.  In the reponse there will be an array of values that can be rendered.  If no `Accept` header is present or `text/html` is contained in the header, JSON will be rendered.
-You may also specify `application/original-timer`, which will output the run file as is with a `Content-type` header with the run's original timer.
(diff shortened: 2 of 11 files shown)


CryZe:
Would it be possible to specify Accept as the "original timer"? At the moment we have to do a request to the API to retrieve that information and then yet another request to download it.

BatedUrGonnaDie:
@CryZe Do you mean having a parameter that just dumps the original file back out no matter what the timer is?

CryZe:
Yeah

BatedUrGonnaDie:
I'm sure something can be worked up.  Is there a specific use case for needing the original file back specifically and not just in one of the timer formats?

CryZe:
livesplit-core's conversion code is less lossy, so we prefer downloading the original file, instead of downloading as a LiveSplit file.

BatedUrGonnaDie:
Ah, that makes sense, I'll see what I can do.

app/controllers/api/v4/runs_controller.rb
```diff
@@ -14,11 +14,28 @@ class Api::V4::RunsController < Api::V4::ApplicationController
...
     else
-      render json: @run, serializer: Api::V4::RunSerializer
+      send_data(
+        if timer == Run.program(@run.timer) || accept == 'original_timer'
+          @run.file
+        else
+          render_to_string(file: "runs/#{accept}.html.erb", layout: false)
+        end,
+        layout: false,
+        type: "application/#{accept}",
```
CryZe: In case of Accept being `original_timer`, this would be `application/original_timer`. I think it may make sense to adjust it properly in that case. It's not important for us, but I think it makes more sense.

BatedUrGonnaDie: That would make much more sense, you are quite right.

app/controllers/api/v4/runs_controller.rb
```diff
@@ -14,11 +14,30 @@ class Api::V4::RunsController < Api::V4::ApplicationController
...
+        disposition: 'inline',
+        filename: "#{@run.filename(timer: timer)}"
+      )
     end
   end
```
glacials: The extra conditionals and variables make this method a bit harder to follow now. Have you considered splitting it up into smaller functions?

app/controllers/api/v4/runs_controller.rb
```diff
@@ -14,11 +14,30 @@ class Api::V4::RunsController < Api::V4::ApplicationController
...
-      @run.parse(fast: false)
-      render json: @run, serializer: Api::V4::RunWithHistorySerializer
+    accept = request.headers.fetch('HTTP_ACCEPT', 'json').downcase
+    if accept == "original_timer"
+      accept = @run.program
+    end
+    timer = Run.program(accept)
+    if accept == 'json' || timer.nil? || !Run.exportable_programs.include?(timer)
+      if params[:historic] == '1'
+        @run.parse(fast: false)
```
glacials: We shouldn't need to do this sort of thing anymore now that history is in the database. If you copied this from somewhere else, that should be safe to remove as well (or you can wait until I get around to it).

app/controllers/api/v4/runs_controller.rb
```diff
@@ -14,11 +14,30 @@ class Api::V4::RunsController < Api::V4::ApplicationController
...
   def show
-    if params[:historic] == '1'
-      @run.parse(fast: false)
-      render json: @run, serializer: Api::V4::RunWithHistorySerializer
+    accept = request.headers.fetch('HTTP_ACCEPT', 'json').downcase
+    if accept == "original_timer"
+      accept = @run.program
+    end
+    timer = Run.program(accept)
+    if accept == 'json' || timer.nil? || !Run.exportable_programs.include?(timer)
```
glacials: If someone passes in an Accept header we don't understand like `application/xml` or `application/livesolit`, I think it would be clearer to 4XX than to succeed and return JSON.

It looks like the proper way to do this is the [406 error code][1]:
> indicates that the target resource does not have a current representation that would be acceptable to the user agent, according to the proactive negotiation header fields received in the request (Section 5.3), and the server is unwilling to supply a default representation.
>
> The server SHOULD generate a payload containing a list of available representation characteristics and corresponding resource identifiers from which the user or user agent can choose the one most appropriate.

[1]: https://tools.ietf.org/html/rfc7231#section-6.5.6

app/controllers/api/v4/runs_controller.rb
```diff
@@ -14,11 +14,30 @@ class Api::V4::RunsController < Api::V4::ApplicationController
...
 
   def show
-    if params[:historic] == '1'
-      @run.parse(fast: false)
-      render json: @run, serializer: Api::V4::RunWithHistorySerializer
+    accept = request.headers.fetch('HTTP_ACCEPT', 'json').downcase
+    if accept == "original_timer"
+      accept = @run.program
+    end
+    timer = Run.program(accept)
```
glacials: It looks like we're specifying our outgoing Content-Types as `application/<format>` but expecting consumers to pass us `<format>`; imo we should be using the former for both cases.

This might be a good place to add a new method like `content_type` to each of the parsers, like we do for `file_extension` and `website`, so each parser can be in charge of how its name gets translated into a Content-Type header.

spec/controllers/api/v4/runs_controller_spec.rb
```diff
@@ -111,6 +111,23 @@
...
+
+        expect(subject).to have_http_status 200
+      end
+
+      it 'returns the correct content-type header' do
+        request.headers["Accept"] = 'wsplit'
+
+        expect(response.content_type).to eq('application/wsplit')
+      end
+    end
```
glacials: Let's also test expected failure cases, like passing a bogus Content-Type (see above).

app/controllers/api/v4/runs_controller.rb
```diff
@@ -14,11 +14,30 @@ class Api::V4::RunsController < Api::V4::ApplicationController
...
+        if timer == Run.program(@run.timer)
+          @run.file
+        else
+          render_to_string(file: "runs/#{accept}.html.erb", layout: false)
+        end,
+        layout: false,
+        type: "application/#{accept}",
+        disposition: 'inline',
+        filename: "#{@run.filename(timer: timer)}"
+      )
```
glacials: This `send_data` looks similar to the code we use to make a user's browser initiate a download (rather than rendering text in the browser). Do we know what the technical difference is that causes browsers to treat it differently, and do we know that it won't throw API consumers?

app/controllers/api/v4/runs_controller.rb
BatedUrGonnaDie: So have a class method that just returns something like "application/wsplit", correct?

app/controllers/api/v4/runs_controller.rb
BatedUrGonnaDie: Wasn't sure what the best way to handle bad accept headers were, are there some that should render the default json schema besides `application/json`?

app/controllers/api/v4/runs_controller.rb
glacials: Yeah exactly.

app/controllers/api/v4/runs_controller.rb
BatedUrGonnaDie: Yeah, I used the base runs controller as a base for the code here.  Should everything always return the history by default and not take the parameter anymore?

app/controllers/api/v4/runs_controller.rb
BatedUrGonnaDie: From what I was able to find, `send_data` allows us to send arbitrary data down the pipe.  What do you mean by throw api consumers?

app/controllers/api/v4/runs_controller.rb
BatedUrGonnaDie: For case formatting should it be all down case or with the custom casing like LiveSplit?

glacials: There's probably some RFC somewhere that has a standard in it. From what I've seen it's all lowercase with hyphens for spaces. So something like `application/livesplit` and `application/time-split-tracker` etc.

app/controllers/api/v4/runs_controller.rb
glacials: Just `application/json` and a missing Accept header should render JSON I think. I can't think of any case other than those two.

app/controllers/api/v4/runs_controller.rb
glacials: I don't think so. Even though history is much easier to retrieve now, there might still be thousands or hundreds of thousands of entries, which would bloat and slow down responses. Definitely still worth reevaluating now that our setup is better though, perhaps the way we're doing it now is too conservative. But we can figure that out outside of this PR.

app/controllers/api/v4/runs_controller.rb
BatedUrGonnaDie: Gotcha, I'll look around for some kind of specification when I have some time.

app/controllers/api/v4/runs_controller.rb
BatedUrGonnaDie: Then I'll leave it as is since I didn't change any of the JSON rendering besides nesting it in a conditional.

app/controllers/api/v4/runs_controller.rb
glacials: By throw I mean mess up how they expect their data to arrive. I looked it up and it's a `Content-Disposition` header; if it's `inline`, the browser will render the text, if it's `attachment`, the browser will download the text as a file.

So it looks like there's no actual difference in the body itself, and I can't immediately find anything online saying if/how APIs are expected to set this header, so I don't think we need to worry about anything here.

BatedUrGonnaDie: I'll write a quick script to download from the endpoint and make 100% sure that the programs can open them without any errors.

BatedUrGonnaDie:
I always forget to add stuff to docs.  Are you ok with using `original_timer` as the name for native file?  Wasn't sure what a good name would be for that.

app/controllers/api/v4/runs_controller.rb
BatedUrGonnaDie: Would `text/html` be a good one to catch for people testing stuff it it browsers?

app/controllers/api/v4/runs_controller.rb
BatedUrGonnaDie: Is taking the if/else and filename call outside of the function call succinct enough?  Filename could probably be removed but I left it in case someone is testing something in browser and wanted to save the page, same with disposition.  As you said in another reply they most likely aren't completely needed, just kept for ease of use.

app/controllers/api/v4/runs_controller.rb
glacials: Ah, I forgot to think about browsers. The endpoint definitely needs to work and return JSON if accessed in a browser. If `text/html` is what browsers send, then yeah let's special case that.

app/controllers/api/v4/runs_controller.rb
glacials: Yeah filename should remain, I think the core readability issue with lots of functions in Splits I/O is having some inputs and then branching code below where multiple code paths use the inputs.

Inevitably the branches grow and when you're skimming logic that's 20+ lines away from the variable assignments that it's depending on, it's hard to understand at a glance what the values could be.

One possible solution is to condense the bigger of the branching code paths into methods, to end up with something like
```ruby
timer = format_from_headers(request.headers, run: @run)
if timer.present?
  render_as_format(timer) # a simple method that handles the send_data mess
  return
end

if params[:historic] == '1'
  @run.parse(fast: false)
  render json: @run, serializer: Api::V4::RunWithHistorySerializer
else
  render json: @run, serializer: Api::V4::RunSerializer
end
```

glacials:
Yeah that's fine with me, just use `-` over `_` if we discover / decide on that from that other comment related to casing.

BatedUrGonnaDie:
Ok, it will be `-` then because I found a nice MDN article on mime types and it says case insensitive, but try to use lower case, and that no spaces are allowed, so use -'s instead.

BatedUrGonnaDie:
I think I was able to clean it up pretty well.  Definitely make sure the documentation I added is up to scratch.  I'm not very good at explaining stuff usually.

docs/api_v4.md
```diff
@@ -46,6 +46,11 @@ If a `historic=1` param is included in the request, one additional field will be
 |-------------:|:---------------------------------------------|:------------------------------------------------------|:----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `history`    | array of numbers                             | never                                                 | Ordered durations of all previous runs. The first item is the first run recorded by the runner's timer into the source file. The last item is the most recent one. This field is only nonempty if the source timer records history.           |
 
+If an `Accept` header is present, splits.io will try to render the run file natively not in JSON of the type specified.  The format for the header is `application/timer`, with timer being any of the timers that can be exported by splits.io.
```
glacials: splits.io is the domain, Splits I/O is the website, so usually use the latter when talking about it in a sentence. (splits-io or splitsio is the code base, still undecided on that one but not relevant here)

docs/api_v4.md
```diff
@@ -46,6 +46,11 @@ If a `historic=1` param is included in the request, one additional field will be
...
+If the `Accept` header is valid, the `Content-type` header in the response will be in the format `application/timer`, with timer being the specified timer.  If the `Accept` header is not valid, the `Content-type` header of the response
```
glacials: I suggest using "`application/<timer>`" over "`application/timer`" to avoid confusion about if the literal `application/timer` is an acceptable header. Then you can refer to `<timer>` later in the sentence to describe what it should be replaced with (like you're doing).

docs/api_v4.md
```diff
@@ -46,6 +46,11 @@ If a `historic=1` param is included in the request, one additional field will be
...
+will be `application/json`, and the status code will be a 406.  In the reponse there will be an array of values that can be rendered.  If no `Accept` header is present or `text/html` is contained in the header, JSON will be rendered.
+You may also specify `application/original-timer`, which will output the run file as is with a `Content-type` header with the run's original timer.
+
```
glacials: Although HTTP headers are case-insensitive, `Content-type` is usually referred to as `Content-Type`.

docs/api_v4.md
glacials: "natively not in JSON of the type specified" is a little confusing, but it sounds like you mean "in the type specified rather than in JSON".

glacials: We should list all the valid Accept headers. If we add support for another timer it'll be more work to update this (and possibly confusing if we forget to), but the work saved for readers by having a usually-complete list here will far outweigh that.

docs/api_v4.md
```diff
@@ -46,6 +46,11 @@ If a `historic=1` param is included in the request, one additional field will be
...
 | `history`    | array of numbers                             | never                                                 | Ordered durations of all previous runs. The first item is the first run recorded by the runner's timer into the source file. The last item is the most recent one. This field is only nonempty if the source timer records history.           |
 
+If an `Accept` header is present, splits.io will try to render the run file natively not in JSON of the type specified.  The format for the header is `application/timer`, with timer being any of the timers that can be exported by splits.io.
+If the `Accept` header is valid, the `Content-type` header in the response will be in the format `application/timer`, with timer being the specified timer.  If the `Accept` header is not valid, the `Content-type` header of the response
+will be `application/json`, and the status code will be a 406.  In the reponse there will be an array of values that can be rendered.  If no `Accept` header is present or `text/html` is contained in the header, JSON will be rendered.
```
glacials: The `text/html` explanation is unnecessary. As long as the endpoint works in a browser like the others, readers will never need to know or care that this is a special case. If they do, well, we're open source.

docs/api_v4.md
glacials: In splitsio we use one space after sentences, not two.

docs/api_v4.md
BatedUrGonnaDie: Do you have a preferred format for the list of acceptable values?

BatedUrGonnaDie: I tried adding a table cause it looked snazzy to me, let me know if it's actually good looking or if I'm crazy.

glacials: Yeah this looks good. There's a couple of sentences above it that restate what the table says that can probably be removed, overall this is a good format for this info though.

docs/api_v4.md
```diff
@@ -46,6 +46,23 @@ If a `historic=1` param is included in the request, one additional field will be
...
+| `Accept` Header                  | Return Format      | Return `Content-Type`            |
+|---------------------------------:|:-------------------|:---------------------------------|
+| None                             | JSON               | `application/json`               |
+| `application/json`               | JSON               | `application/json`               |
+| `application/wsplit`             | WSplit             | `application/wsplit`             |
+| `application/time-split-tracker` | Time Split Tracker | `application/time-split-tracker` |
+| `application/splitterz`          | SplitterZ          | `application/splitterz`          |
+| `application/livesplit`          | LiveSplit          | `application/livesplit`          |
+| `application/urn`                | Urn                | `application/urn`                |
+| `application/original-timer`     | Original Run File  | The above timers `Content-Type`  |
```
glacials: timer's

BatedUrGonnaDie:
Thanks for all the help writing the docs and code input 👍

PR #311, opened 2017-08-29
fix the bug of text overlapping on download link

yutingguo:


glacials:
for posterity this is the bug in question:

untitled

app/views/runs/_title.slim
```diff
@@ -47,7 +40,15 @@ h6
           = link_to "claim", '#', id: 'claim-nav-link', data: {confirm: "Claim this run as #{current_user.twitch_display_name}?"}
         - else
           = link_to "sign in to claim", '/auth/twitch'
-
+  br
```
glacials: Generally `br` isn't the desired way to separate content vertically, see https://stackoverflow.com/a/1726103 for why.

We can probably get the same effect by changing the `#download-menu` from a span into a div, and changing the `.run-options` CSS from `display: inline` to `display: block`.

glacials:
Looks good to me! Thanks!

PR #312, opened 2017-09-10
Use livesplit-core for parsing

### LINKED ISSUE #277: Use livesplit-core for Parsing
Continuing from our discussion at https://github.com/glacials/splits-io/issues/270

There are builds available in our repository for 23 common platforms, as both shared and static library and with the Ruby bindings needed for splits.io. So it should be somewhat easy to start using this now.

https://github.com/CryZe/livesplit-core/releases

BatedUrGonnaDie:
Reclassify all old parsers as programs now, and give them the exportable class method to replace the manual list in Run model.  Programs added for all those supported by livesplit-core.

Edit: Closes #277


BatedUrGonnaDie:
Currently the parser has a lot of duplicate information to make sure that all the tests pass, and I wasn't 100% sure how it should be reduced.  Didn't want to remove stuff that would break api outputs that should be fixed and all that jazz.

app/models/run.rb
```diff
@@ -53,9 +57,9 @@ def program(string_or_symbol)
       h[string_or_symbol.to_s]
     end
 
-    def program_from_content_type(content_type_string)
-      Run.exportable_programs.each do |prg|
-        return prg if prg.content_type == content_type_string
+    def program_from_attribute(func_name, value)
+      Run.programs.map do |prg|
+        return prg if prg.send(func_name) == value
```
glacials: I'm in favor of the abstraction, but `send` makes me very nervous. This has strong potential to be an attack vector if some future code passes user input in without realizing what the function is doing.

I would rather have a bit of duplicated code such as one function per `func_name`. Or, you can have this function switch/case on a specific set of `func_name` values. Either way we should not allow any arbitrary method to be called (all modules already come with [a lot of dangerous ones that we don't think about][1]).

[1]: https://ruby-doc.org/core-2.4.1/Module.html

glacials:
As for the outputs, this is fine for now because it matches existing behavior, and that's the most important thing when switching out massive backends like this.

We can go in after it's merged and remove a lot of the translation code between livesplit-core and our own models, by killing `Split` and the `run_object` hash that `Run#parse_into_activerecord` expects. This code was nice when communicating with multiple parsers, but now that we have only one parser and one output format we can ditch a lot of the middleman code.

But let's keep this branch as simple as we can and do that stuff later :)

app/models/run.rb
BatedUrGonnaDie: Is send considered one of those methods you should really never use like in JS with eval?

BatedUrGonnaDie:
Should parsing.md be removed as well and just put upgrade info for livesplit-core in the main readme?

glacials:
Yeah that's a good move, forgot about that file.

app/controllers/api/v4/runs_controller.rb
```diff
@@ -16,7 +16,7 @@ class Api::V4::RunsController < Api::V4::ApplicationController
   end
 
   def show
-    timer = Run.program_from_content_type(@accept_header)
+    timer = Run.program_from_attribute('content_type', @accept_header)
```
glacials: when passing around internal-only things such as method names, generally prefer symbols over strings

app/models/run.rb
```diff
@@ -36,11 +37,14 @@ class Run < ApplicationRecord
 
   class << self
     def programs
-      [LlanfairGered, Llanfair, Urn, LiveSplit, SplitterZ, TimeSplitTracker, WSplit]
+      [
+        ShitSplit, Splitty, Llanfair2, SplitterZ, FaceSplit, Llanfair2, Portal2LiveTimer,
+        LlanfairGered, Llanfair, Urn, LiveSplit, SplitterZ, TimeSplitTracker, WSplit
+      ]
```
glacials: 1. Llanfair2 and SplitterZ are in this list twice
2. Be sure to modify the top of the README, which specifies what timers we accept, and the v4 API doc, which specifies which timers we can output when given `Accept: application/original-timer`

app/models/run.rb
```diff
@@ -53,9 +57,23 @@ def program(string_or_symbol)
...
+        when 'website'
+          prg_value = prg.website
+        when 'content_type'
+          prg_value = prg.content_type
+        when 'exportable'
+          prg_value = prg.exportable
+        end
+        return prg if prg_value == value
       end
       return nil
```
glacials: this `return nil` case should be a super obvious and unmissable error, like a `raise`, because it means there is a consumer calling this function incorrectly

lib/parser/livesplit_core_parser.rb
```diff
@@ -0,0 +1,107 @@
...
+      run_object[:realtime_history] = []
+      (0...run.attempt_history_len()).each do |i|
+        attempt = run.attempt_history_index(i)
+        attempt_id = attempt.index().to_i
+        time = attempt.time()
+
+        run_object[:history][attempt_id] = {
+          gametime: time.game_time().try(:total_seconds) || 0,
+          realtime: time.real_time().try(:total_seconds) || 0
+        }
```
glacials: if nothing is using this new history field yet, can we put milliseconds in it instead? that way when we start using it later we don't have to convert anything.

lib/parsers/livesplit.rb
```diff
@@ -1,366 +0,0 @@
...
-    end
-
-    def duration_in_seconds_of(time)
-      return 0 if time.blank?
-
-      time.sub!('.', ':') if time.count('.') == 2
-      ChronicDuration.parse(time, keep_zero: true)
-    end
-  end
-end
```
glacials: goodbye old friend. you served us well

ish

app/models/run.rb
BatedUrGonnaDie: The output timers haven't changed AFAIK because core can only save as .lss and not any other formats.  So we would still need to write templates for any of those timers if we wanted to support outputting to them.

lib/parser/livesplit_core_parser.rb
BatedUrGonnaDie: Totally, should I also change the new split.history hash to also convert down to milliseconds?

glacials: yeah, if it's not an unreasonable amount of work

app/models/run.rb
glacials: i believe i understand that part, but we still can potentially return those formats if the original upload was in one of those unexportable formats, right?

app/models/run.rb
glacials: yeah pretty much. i would only ever consider using it to perform magic in a library that was well-guarded against the outside world abusing it.

README.md
```diff
@@ -79,8 +79,15 @@ xcode-select --install
 See the full API documentation in the [API readme](./docs/api.md) file.
 
 ## Parsing
-If you're interested in how we parse run files from different programs, check out the [parsing
-readme](./docs/parsing.md).
+Split I/O uses [LiveSplit Core][livesplit-core] for parsing. The parser is located in `lib/parser/*`.
```
CryZe: Slight nitpick: I prefer livesplit-core.

lib/parser/livesplit_core_parser.rb
BatedUrGonnaDie: Turns out indexed_history is used by other methods, and I would have to add a new field to the db to achieve this which seems a bit out of scope for this PR.

BatedUrGonnaDie:
Okay, I think I got everything changed that was requested. Are the instructions for updating livesplit-core clear enough?

README.md
BatedUrGonnaDie: Makes sense, wasn't sure what the exact capitalization should be since some places use LiveSplit and other livesplit.

app/controllers/api/v4/runs_controller.rb
```diff
@@ -33,6 +26,13 @@ def show
...
         filename: "#{@run.filename(timer: timer)}"
       )
+    rescue
+      if params[:historic] == '1'
+        @run.parse(fast: false)
+        render json: @run, serializer: Api::V4::RunWithHistorySerializer
+      else
+        render json: @run, serializer: Api::V4::RunSerializer
+      end
     end
```
glacials: why the begin/rescue?

app/models/run.rb
```diff
@@ -53,11 +57,25 @@ def program(string_or_symbol)
...
+          prg_value = prg.website
+        when :content_type
+          prg_value = prg.content_type
+        when :exportable
+          prg_value = prg.exportable
+        end
+        return prg if prg_value == value
       end
-      return nil
+      raise 'timer not found'
```
glacials: oops, my previous comment was actually referring to the situation of `func_name` being unexpected, like `program_from_attribute(:blontent_type, 'application/livesplit')`. in that situation, we should `raise`.

i didn't notice that this line could be reached in an expected way, like `program_from_attribute(:content_type, 'some-new-timer')`.

we should `raise` only in the former case, the latter case should still be a `nil` return value like you had. sorry, my mistake.

docs/api_v4.md
```diff
@@ -51,16 +51,31 @@ If the `Accept` header is valid, the `Content-Type` header in the response will
...
+* `application/llanfair2`
+* `application/facesplit`
+* `application/portal-2-live-timer`
+* `application/llanfair-gered`
+* `application/llanfair`
+* `application/urn`
+* `application/livesplit`
+* `application/splitterz`
+* `application/time-split-tracker`
+* `application/wsplit`
```
glacials: 👍

lib/parser/livesplit_core_parser.rb
```diff
@@ -36,9 +36,10 @@ def self.parse(run_string, fast: false)
         time = attempt.time()
 
         run_object[:history][attempt_id] = {
-          gametime: time.game_time().try(:total_seconds) || 0,
-          realtime: time.real_time().try(:total_seconds) || 0
+          gametime: time.game_time().try(:total_seconds).try(:*, 1000) || 0,
+          realtime: time.real_time().try(:total_seconds).try(:*, 1000) || 0
```
glacials: is it actually possible for `total_seconds` to return nil?

glacials:
Yeah those instructions are great.

app/controllers/api/v4/runs_controller.rb
BatedUrGonnaDie: application/json is not a valid program content type, and with my misunderstanding of the raise from the attribute getter this would throw an exception if you selected application/json, which it then catches and renders the serialzier.

app/models/run.rb
BatedUrGonnaDie: Oh, ok.  I'll change it so the default of the switch statement is a raise, and then the last line will be return nil then.  That way if it's called with an invalid attribute it throws, but otherwise it can return nil if not program is found for that attribute?

glacials:
Hmm so including a precompiled binary in the repository is going to be a no-no -- since it's been compiled on and for your architecture, other architectures (including my laptop, which I just tested) will not be able to run using that binary.

The error I get is 
```
Could not open library '/Users/glacials/pj/splits-io/lib/parser/liblivesplit_core.so': dlopen(/Users/glacials/pj/splits-io/lib/parser/liblivesplit_core.so, 5): no suitable image found.  Did find:
	/Users/glacials/pj/splits-io/lib/parser/liblivesplit_core.so: unknown file type, first eight bytes: 0x7F 0x45 0x4C 0x46 0x02 0x01 0x01 0x00
```
when accessing any page on splits-io locally while running this branch.

We will probably have to either include a Makefile and/or script that compiles the library and ask people to run it before their first run (and do the production equivalent for Elastic Beanstalk), or adapt this branch to work in Docker which is 90% complete in the `docker` branch.

I'll leave the selection to you, but I can help with either path if you need it.

BatedUrGonnaDie:
How much more work (and what work) needs to be done for the docker branch to be completed?  It seems like a bit of a waste to be writing scripts if that is almost completed.

glacials:
The branch works locally, the work is all in getting it running in AWS. I've been taking a few-day break from working on it out of frustration, but will start again now that this branch may be using it. I'll timebox it to a week. After that, if it's not ready I'll probably abandon it for now.

glacials:
Code changes all look good. I consider this approved but currently unmergable because of the binary situation.

BatedUrGonnaDie:
I'll probably let this chill until next weekend then, and if docker isn't done I'll look into writing a script to either generate everything or instructions to compile everything beforehand.

CryZe:
I just pushed a new commit, which allows specifying the path of the shared library for the ruby bindings. You can call it like so:
```
cargo run -- --ruby-lib-path "../liblivesplit_core.so"
```
resulting in
```ruby
ffi_lib File.expand_path('../liblivesplit_core.so', __FILE__)
```
in the Ruby code. This also now means that you don't need to rename the library anymore.

BatedUrGonnaDie:
Rebased the branch on top of the docker branch and changed the PR target to docker branch.

BatedUrGonnaDie:
@glacials Re-review whenever you get a chance.  This PR is still blocked by livesplit-core needing to have a new release put out so don't merge.

glacials:
Since this depends on docker I'm going to wait until that's merged and this is rebased to review again

README.md
```diff
@@ -71,5 +71,5 @@ docker-compose build
 to rebuild the Docker image for your changes to apply.
 
 ## Parsing
-If you're interested in how we parse run files from different programs, check out the [parsing
-readme](./docs/parsing.md).
+Split I/O uses [livesplit-core][livesplit-core] for parsing. The parser is located in `lib/parser/*`.
```
glacials: This link ref is missing

README.md
```diff
@@ -22,7 +23,7 @@ dependencies. The one downside is that you must first install Docker!
 
 ### First run
 The first time you run Splits I/O with Docker is the best, because you'll have time to get a coffee! Yum! After the
-first run, it will be much quicker.
+first run, it will be much quicker (on Linux, docker commands may need to be prefixed with sudo).
```
glacials: What is root needed for? I'd prefer to not require users to need it, so if there's something else we can do to let people run docker as themselves, we should probably do it.

BatedUrGonnaDie: Docker attaches itself to a unix socket, owned by root so in order to talk to it you need to sudo it.  The alternative is to create a docker group and add your user to that, but that docker group is essentially a root priviledged group.
https://docs.docker.com/engine/installation/linux/linux-postinstall/

glacials: Gotcha, ok. I wonder why this isn't required on Mac, but the README change looks fine given that situation.

PR #316, opened 2017-09-29
Change how to fill in Twitch credentials again

glacials:
Looks like the `--skip-worktree` has a downside where if you change the file, `--skip-worktree` it, then try to checkout another branch, Git may still complain at you that you have unstashed changes and prevent you from checking out. (And `git stash`ing doesn't even include the Dockerfile if it's been `--skip-worktree`d.)

I really like the idea of not needing to do extra stuff to get splits-io running the first time, so I'm going to try out this method where the extra work is still in the getting-login-working step, but is otherwise more like what we used to do.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,16 +1,12 @@
 # bundler config.
 /.bundle
 
-# the default SQLite database.
-/db/*.sqlite3
-/db/*.sqlite3-journal
-
 # all logfiles and tempfiles.
 /log/*.log
 /tmp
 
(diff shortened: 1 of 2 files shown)


glacials:
fuck

BatedUrGonnaDie:
The only way I can think of doing this is to somehow reference an external file (or environment variable) if it exists, and use those values if the file (or variable) does, else just use the placeholder values like it has now.

PR #317, opened 2017-10-02
Use case insensitive search

### LINKED ISSUE #313: Game shortname search is case sensitive
Game search in general is correctly case insensitive, but typing a game's shortname exactly into the search box is not. e.g. Orcs Must Die! 2 comes up when typing "orcs2" but not "Orcs2". This is especially important for mobile, where most keyboards capitalize the first letter automatically.

BatedUrGonnaDie:
Closes #313

### DIFF
--- a/app/models/game.rb
+++ b/app/models/game.rb
@@ -17,9 +17,8 @@ class Game < ApplicationRecord
   def self.search(term)
     term = term.strip
     return nil if term.blank?
-    term.downcase!
-    ids = where(shortname: term).or(where('LOWER(name) ILIKE ?', "%#{term}%")).pluck(:id)
-    ids ||= GameAlias.where('LOWER(name) ILIKE ?', "%#{term}%").pluck(:game_id)
+    ids = where('LOWER(shortname) = LOWER(?)', term).or(where('name ILIKE ?', "%#{term}%")).pluck(:id)
+    ids = ids | GameAlias.where('name ILIKE ?', "%#{term}%").pluck(:game_id)
     Game.where(id: ids)
   end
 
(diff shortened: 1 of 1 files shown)


app/models/game.rb
```diff
@@ -17,7 +17,9 @@ class Game < ApplicationRecord
   def self.search(term)
     term = term.strip
     return nil if term.blank?
-    ids = where(shortname: term).or(where('name ILIKE ?', "%#{term}%")).pluck(:id) | GameAlias.where('name ILIKE ?', "%#{term}%").pluck(:game_id)
+    term.downcase!
+    ids = where(shortname: term).or(where('LOWER(name) ILIKE ?', "%#{term}%")).pluck(:id)
+    ids ||= GameAlias.where('LOWER(name) ILIKE ?', "%#{term}%").pluck(:game_id)
```
glacials: Cool! Few things:
- The previous code uses `|`, not `||`. When used on an array is a set union operator.
- `ILIKE` is a case-insensitive `LIKE`, so `'LOWER(name) ILIKE ?'` is redundant.
- With this `.downcase!` we're now mixing Ruby data manipulation with SQL data manipulation. For simplicity I'd prefer to keep it to just one of those two (SQL is usually more performant). This will probably mean replacing `where(shortname: term)` with the less idiomatic `where('LOWER(shortname) = LOWER(?), term)` which would be fine.

BatedUrGonnaDie: Ah, I did not know that `|` was a union operator, thought it was acting as some kind of bitwise or.

glacials: It's both 👍 

`Fixnum#|` is a bitwise or, `Array#|` is a set union:
https://ruby-doc.org/core-2.2.0/Fixnum.html#method-i-7C
https://ruby-doc.org/core-2.2.0/Array.html#method-i-7C

PR #319, opened 2017-10-08
Kill DynamoDB

glacials:
Kill the last use of DynamoDB by moving run histories into PostgreSQL. This will knock a few bucks a month off the bill!


app/models/concerns/unparsed_run.rb
```diff
@@ -62,7 +62,7 @@ def parse(fast: true, convert: false)
       parse_result
     end
 
-    def parse_into_activerecord
+    def parse_into_db
       with_lock do
         segments.delete_all
```
BatedUrGonnaDie: Should we also delete all histories here as well?

glacials:
Nice catches!

BatedUrGonnaDie:
Looks good to me 👍

Issue #263, opened 2016-12-10
Compare page should become a first-class citizen


jhobz:
This task originally had a smaller scope, but as I was writing it I realized what I'd actually like is a comparison page that is more of a first-class citizen in general.

Right now the comparison page is adaptable enough to compare any two splits, even those of different games! What would be great is a UI flow for comparing any two splits of a given game, or at least any two of your own splits. This would be similar to the "compare to best on record" or the rivals system.

Proposed flow:
1. Go to a splits page (probably PB)
2. Click "compare" link (would replace "compare to best on record" or "compare to my PB")
3. Be presented a list of splits associated with the same game [alias] and category [alias]. Best on record, your PB, and your last PB are all be stuck to the top in their own UI elements (or at least highlighted in some way). There is also a box available for entering any splits ID.
4. Select a set of splits against which to compare.
5. Be presented the current compare page, plus a multitude of stats and graphs below the splits timeline.

Background:
I recently had someone ask me if they could compare one PB of theirs to another on Splits I/O. I figured it was possible, but I had to go through the cumbersome process of opening both PBs to grab the IDs, and then adding "/<id1>/compare/<id2>" to the URL bar. I had also been bragging about the awesome stats and graphs provided on PBs and was a bit disappointed to find none on the comparison page. It would be awesome if there were some! :)

glacials:
100% agree. For a feature I use so much it should feel way much more natural to use than it does today. And actually I'm upset that they're even as hard to find as they are. imo the first thing you see after uploading a run should be a comparison with your rival's run or your friends' runs or anything. Seeing a run on its own is ok but the default should be a comparison.

I really think we should basically merge the current run page with the comparison page, such that when you visit someone else's PB you're already immediately seeing it compared to yours, or their previous run, or WR or whatever, and you can swap it out for other runs without even a page load. That is the ultimate sick futureworld I want to get to.

Charts is something I hadn't even thought about for comparisons but now that you've brought it up I can't stop thinking up cool ones to display. Like 1D tug-of-war charts for each segment best, or "race to XX:YY" charts that develop over months, even containing any number of runners. Even just taking the charts on the current stats page and putting both runners in one chart, as dual-bar charts and stuff.

The stats page is also the slowest page right now because it has parse and sift through the entire run history, and with the recent new DynamoDB changes I've been making I want to improve from that end as well -- especially if this page is going to be the new primary page on the site.

jhobz:
[...]

PR #321, opened 2017-10-12
Move segment history graphs onto the run page

glacials:
Explodes the multi-line segment history graph on the stats page into one graph per segment, and puts them right inside the segment rows on the run show page.

### Current (stats page)
screenshot 2017-10-11 21 49 44

### New (run page)
screenshot 2017-10-11 21 50 09

This is kinda the first step of rolling the stats and compare pages into the standard run page (related: #263). Historically we've kept a lot of this stuff on the stats page because we need to look at history to generate it, and looking at history is expensive. But that's where all the useful stuff is, so I don't think we should keep it out of the limelight. We can figure out how to scale it as we go.

As a side note I'm trying out JS here over CoffeeScript because ES5 and 6 adoption seem to be making JS a lot more usable for my simple mind.


app/assets/javascripts/graphs/segment_history.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  color_styles = [
+    "rgba(14, 144, 210, 1)",
+    "rgba(128, 87, 165, 1)",
+    "rgba(94, 185, 94, 1)",
+    "rgba(250, 210, 50, 1)",
+    "rgba(221, 81, 76, 1)",
+    "rgba(243, 123, 29, 1)"
+  ]
```
BatedUrGonnaDie: Since every segment has its own graph can all the colors except the one be removed?

app/views/runs/_split_table.slim
```diff
@@ -14,6 +14,7 @@
           - else
             th.align-right Duration
             th.align-right Finished At
+            th.align-right History
```
BatedUrGonnaDie: This doesn't seem to be showing up in the table for me, the spot where it goes is just blank.

app/assets/javascripts/turbolinks.js
```diff
@@ -0,0 +1,11 @@
+$(function() {
```
BatedUrGonnaDie: Looks like something here is causing an error on the gametime switch toggle causing it to not activate.  The faq page also throws a js error but nothing appears broken on it.

app/assets/stylesheets/application.sass
```diff
@@ -19,9 +19,11 @@
 @import code
 @import faq
 @import footer
+@import graphs
```
BatedUrGonnaDie: Any reason to not move all of the graphs into this folder now?

app/assets/javascripts/graphs/segment_history.js
glacials: Ooo yeah, they can!

app/views/runs/_split_table.slim
glacials: Huh. That's weird, it shows for me. Do the "View" links in the `td`s show up? What browser / OS?

app/assets/stylesheets/application.sass
glacials: Nope just laziness, thanks for calling me out on it. Moving!

app/assets/javascripts/graphs/segment_history.js
glacials: oooo yes it can!

app/views/runs/_split_table.slim
BatedUrGonnaDie: chrome_2017-10-12_21-49-22

Both Windows 7 Chrome and OSX Chrome.

app/assets/javascripts/turbolinks.js
glacials: Looks like the FAQ page thing also happens on the current live site so it unrelated to this change. I'll make an issue for it separately.

Looking into the game switch thing.

app/views/runs/_split_table.slim
BatedUrGonnaDie: Amazingly I managed to click gametime runs on 2 different computers to see it happen the 2nd time tonight.

app/assets/javascripts/turbolinks.js
BatedUrGonnaDie: Oh, looks like its a naming conflict.  Change the name of this file and the error goes away.  I assume the include in the application.js file is loading this as it is checked first in the pipeline.

glacials: Yep, I figured it out at the same time you did lol. It tries to load Turbolinks the library and ends up loading my turbolinks.js instead. Oh Rails.

BatedUrGonnaDie:
Looks good to me 👍

glacials:
Hmm the deploy isn't having it, will have to look later when I have time.

BatedUrGonnaDie:
@glacials Do you think this will pass the deploy now?

glacials:
Hopefully! At the very least the current beta/prod setup will make it safe to try and try again.

PR #322, opened 2017-10-12
Make search in navbar expand to fill area

jhobz:
The shorter search bar kind of bugged me, especially since it expanded
to full-width in the mobile overflow menu. This uses flexbox to let the
search bar expand to fill the entire area at any window width greater
than or equal to 768px.

Full disclosure: this was NOT tested very well. I could not get Docker
to start locally (on Windows) and thus was only able to test this via
the ol' browser tools. I don't even know if this was the correct
stylesheet to edit.

Here's what it looks like on my local machine, for reference:
image

### DIFF
--- a/app/assets/stylesheets/application.sass
+++ b/app/assets/stylesheets/application.sass
@@ -61,7 +61,7 @@ h2 small
   display: block
 
 #download-menu
-  visibility: hidden  
+  visibility: hidden
   float: left
   margin: 0
   white-space: nowrap
@@ -113,3 +113,14 @@ a
 
 .kill-run-button
   cursor: pointer
+
+@media (min-width: 768px) {
+  #global-nav
+    display: flex !important // !important needed to override bootstrap use of !important
+    & > .col-sm-3
+      flex-grow: 1
+      & > .navbar-form > .input-group
+        display: flex;
+  #game_q
+    flex-grow: 1
+}


glacials:
Nice! Just tested locally, syntactically we've got the `{`, `}`, and `;` that need deleting because SASS. With that done it works!

However I think we can do this in a more Bootstrappy way than custom CSS rules. If we change the classes for [form][1]'s parent and child divs like
```slim
.col-lg-7.col-md-6.col-sm-5
  = form_for ...
    .input-group.col-md-12
```
it looks like we can get similar behavior using built-in Bootstrap classes!

[1]: https://github.com/glacials/splits-io/blob/601b4489307dd158e1733661e7b2e159ac7813fe/app/views/layouts/application.slim#L52

jhobz:
Bah, SASS. Gotta get with the times and use SCSS convention! ;)

I'll fix up the class stuff, but I want to try to get Docker working first so I don't put the burden of initial testing all on you.

BatedUrGonnaDie:
What windows are you running @jhobz ?

jhobz:
Windows 10, Creator Update (though the first time I tried was pre- Creator Update).

Here's a screenshot of the error I get when trying to just start Docker. This is Docker's Edge build, which was my attempt at solving the problem after the stable build failed with the same error.
image

BatedUrGonnaDie:
Are you on professional or enterprise?

jhobz:
Windows 10 Home.

BatedUrGonnaDie:
Looks like Docker CE for windows requires professional or enterprise.
https://www.docker.com/docker-windows

jhobz:
Wow, that's pretty asinine. I guess I didn't notice that originally because it doesn't actually put the asterisk in the download box anywhere, so I probably just scrolled down to the button and clicked it.

Guess I won't be able to contribute unless I ever get around to throwing linux back on my laptop. :(

BatedUrGonnaDie:
You can always do what I do and have vitualbox installed and just use a ubuntu vm.  It honestly works quite well and doesn't take that much cpu or ram.

jhobz:
I've gone down the virtualbox path before and just had horrible experiences, so I tend to just avoid it altogether at this point.

BatedUrGonnaDie:
To each their own.

glacials:
Closing for inactivity, reopen if needed.

PR #328, opened 2017-10-18
Fix GA error

### LINKED ISSUE #324: Console error on FAQ
The FAQ page has a console error, which prevents drag-and-drop upload from working on that page.

BatedUrGonnaDie:
Fixes #324

### DIFF
--- a/app/assets/javascripts/analytics.js
+++ b/app/assets/javascripts/analytics.js
@@ -69,7 +69,7 @@ $(function() {
       }
       break;
     default:
-      ga('set', 'page', event.path[0].location.pathname);
+      ga('set', 'page', window.location.pathname);
   };
 
   ga('send', 'pageview');
--- a/app/assets/javascripts/analytics.js
+++ b/app/assets/javascripts/analytics.js
@@ -69,7 +69,7 @@ $(function() {
       }
       break;
     default:
-      ga('set', 'page', event.data.path);
+      ga('set', 'page', event.path[0].location.pathname);
   };
 
   ga('send', 'pageview');


app/assets/javascripts/analytics.js
```diff
@@ -69,7 +69,7 @@ $(function() {
       }
       break;
     default:
-      ga('set', 'page', event.data.path);
+      ga('set', 'page', event.path[0].location.pathname);
```
glacials: What even is `event` here? What sets it?

BatedUrGonnaDie: Not gonna lie I thought you knew.  I used it because it was what the old code referenced.  I have no idea what sets it though.

jhobz: I _believe_ `event` is a hidden local in most JavaScript functions. Alternatively, it may be injected by jQuery to any callback functions, which this code technically is.

jhobz: A bit more research suggests it is actually a global variable set by the browser whenever an event is fired, so the contents may not be reliable.
https://stackoverflow.com/questions/6426497/is-event-a-global-variable-that-is-accessible-everywhere-inside-the-callback-cha

glacials: lol ok I have no idea why I wrote it like this. To be honest I think maybe I pulled this out of a code snippet that was supposed to be inside a callback like `function(event) { ... }` and never realized it. Let's use the proper way of getting path which looks like `window.location.pathname`.

PR #329, opened 2017-10-19
Add ability to export to llanfair-gered

### LINKED ISSUE #258: Support downloading in Gered's Llanfair fork's format
We now support uploads from [Gered's Llanfair fork][1], so the next thing is to add downloads into that format.

[1]: https://github.com/gered/llanfair

BatedUrGonnaDie:
Fixes #258

### DIFF
--- a/app/models/concerns/completed_run.rb
+++ b/app/models/concerns/completed_run.rb
@@ -74,9 +74,5 @@ module CompletedRun
     def completed?(timing)
       duration_ms(timing).present? && duration_ms(timing) > 0
     end
-
-    def completed_runs_count(timing)
-      histories.to_a.count { |history_element| !history_element.duration_ms(timing).zero? }
-    end
   end
 end
--- a/app/models/concerns/completed_run.rb
+++ b/app/models/concerns/completed_run.rb
@@ -74,5 +74,9 @@ module CompletedRun
     def completed?(timing)
       duration_ms(timing).present? && duration_ms(timing) > 0
     end
+
+    def completed_runs_count(timing)
+      histories.to_a.count { |history_element| !history_element.duration_ms(timing).zero? }
+    end
   end
 end
--- a/app/models/run_history.rb
+++ b/app/models/run_history.rb
@@ -1,14 +1,14 @@
 class RunHistory < ApplicationRecord
   belongs_to :run
 
-  scope :completed, (lambda do |timing|
+  def self.completed(timing)
     case timing
     when Run::REAL
-      where('realtime_duration_ms > 0').count
+      RunHistory.where('realtime_duration_ms > 0')
     when Run::GAME
-      where('gametime_duration_ms > 0').count
+      RunHistory.where('gametime_duration_ms > 0')
     end
-  end)
+  end
 
   def duration_ms(time_type)
     case time_type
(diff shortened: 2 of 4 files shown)


app/models/concerns/completed_run.rb
```diff
@@ -74,5 +74,9 @@ def total_playtime_ms(timing)
     def completed?(timing)
       duration_ms(timing).present? && duration_ms(timing) > 0
     end
+
+    def completed_runs_count(timing)
+      histories.to_a.count { |history_element| !history_element.duration_ms(timing).zero? }
```
glacials: This code fetches all history entries into memory then counts them. We should be able to have the database do the counting for us, which will be much faster and reduce the risk of running out of memory, with something like
```sql
case timing
when Run::REAL
  histories.where('realtime_duration_ms > 0').count
when Run::GAME
  histories.where('gametime_duration_ms > 0').count
```
This may be also be better laid out as a scope on RunHistory ([see examples on Run][1]), which will give us a more Rubyesque syntax from the caller like
```ruby
@run.histories.completed.count
```
and still provide the same benefit.

[1]: https://github.com/glacials/splits-io/blob/master/app/models/run.rb#L31-L36

BatedUrGonnaDie: `scope :by_category, ->(category) { where(category: category) }`
Is the `(category)` part a variable that you can pass is?  Never worked with scopes before so I don't understand the syntax that well.

glacials: Yeah, `scope` is a method that takes two arguments, a name for the scope and a lambda. Lambdas are a Ruby (and other languages) construct and are sort of like code blocks. The stuff in the parens is the argument list to the lambda, just like the stuff in `|pipes|` is the argument list to a code block.

It could be alternatively written as
```ruby
scope(:by_category, lambda { |category| where(category: category) })
```

app/models/run_history.rb
```diff
@@ -1,6 +1,15 @@
...
   belongs_to :run
 
+  scope :completed, lambda { |timing|
+    case timing
+    when Run::REAL
+      where('realtime_duration_ms > 0').count
+    when Run::GAME
+      where('gametime_duration_ms > 0').count
+    end
+  }
```
glacials: For multiline use `do ... end` rather than `{ ... }` to match convention.

BatedUrGonnaDie: One of these years I'll get all the different conventions down XD

BatedUrGonnaDie: Apparently there is some sort of precedence error when using a lambda in the scope function, so there has to be () around the lambda.  While looking to see what was the problem I saw that most people recommend that scopes be one or 2 lines, and everything else should be a function.  Should I leave it like this or change it around?

Nightfirecat: I find that `rubocop` is a lifesaver for sticking to convention.

BatedUrGonnaDie: The thing is rubocop is fine with either, but when I was google'ing around for the issue I saw other people saying that scopes should be short and sweet.

glacials: Rubocop can't catch everything, we should stick with popular convention even if Rubocop has no opinion on it.

BatedUrGonnaDie:
I moved the `.count` call out of the `self.completed` call because in the future maybe we will want to do operations on the relation of just completed run histories.  Good idea or nah?

PR #332, opened 2017-10-20
Use separate databases for each environment

BatedUrGonnaDie:
This makes it much easier to examine stuff in the rails console without having hundreds of test users, games, and categories filling up the queries instead of the data you want to test stuff with.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,6 @@ ENV ENABLE_ADS 0
 # Nothing below should ever need to be changed
 ENV RAILS_ENV development
 
-ENV DATABASE_URL postgres://postgres@db/postgres
 ENV S3_BUCKET splits-io
 ENV AWS_REGION local
 ENV AWS_ACCESS_KEY_ID beep
--- a/config/database.yml
+++ b/config/database.yml
@@ -1,0 +1,19 @@
+default: &default
+  adapter: postgresql
+  encoding: unicode
+  pool: 5
+  timout: 5000
+  username: postgres
+  host: db
+  port: 5432
+
+development:
+  <<: *default
+  database: app_development
+
+test:
+  <<: *default
+  database: app_test
+
+production:
+  url: <%= ENV['DATABASE_URL'] %>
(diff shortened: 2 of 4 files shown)


.dockerignore
```diff
@@ -0,0 +1,4 @@
+.git*
+log/*
+tmp/*
+docs/*
```
glacials: What's this file for?

BatedUrGonnaDie: Oh, this was a leftover of some old experiments on speeding up build times.  It stops things from being loaded into the docker container that is not needed for building.  I'll remove it next commit.

README.md
```diff
@@ -25,8 +25,18 @@ dependencies. The one downside is that you must first install Docker!
 The first time you run Splits I/O with Docker is the best, because you'll have time to get a coffee! Yum! After the
 first run, it will be much quicker (on Linux, docker commands may need to be prefixed with sudo).
 ```sh
+docker-compose build
+```
+Once the build finishes, you need to create the database.
+```sh
+docker-compose run web rake db:crete
```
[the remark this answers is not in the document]
BatedUrGonnaDie: It can't be done in `Dockerfile` because the images are not linked at that point.  The only place it could go would be in `docker-compose.yml` before the migrate statement.  Could do something like `command: bash -c "rm -f /app/tmp/pids/server.pid && bundle exec rails db:create db:migrate && bundle exec rails assets:precompile && bundle exec rails s -p 3000 -b '0.0.0.0'"`

.dockerignore
Nightfirecat: Future PR material, I hope? ;)

BatedUrGonnaDie:
I may have messed up my git history so it's a single new commit, but I think I found a way to create the tables fairly well.  By placing init files in the db container the image will autorun them in ascii order before allowing connections, which means it will be created before a connection can be made.

BatedUrGonnaDie:
What's nice is that it will run any scripts in that directory in ascii order, so if you need to do any manipulations in the future it can be added and run at startup .

PR #345, opened 2018-02-02
Try and fix aws build errors by removing es6 syntax

BatedUrGonnaDie:


BatedUrGonnaDie:
AFAIK, nothing in the pipeline compiles down to browser level ES.  Sprockets 4 is supposed to add support for it, or you can use webpack which is also starting to be bundled with rails as of 5.1 like yarn.  The logs say ExecJS can't find the token `reset_counter`, which means either its an es6 problem or that the scoping of `let` means that it can't be used in the inner loop, but I'm not a js expert so I don't know if it's a scoping problem.  The hope is that this fixes it and then we can add in either the sprockets beta release or webpacker to allow full es6 and beyond syntax since it allows you to use babel and its plugins which allow for all the new cool stuff.

glacials:
Cool, sounds like a solid investigation and good next step. 👍

PR #346, opened 2018-02-03
Add webpack support

BatedUrGonnaDie:
Currently things are in a bit of an in between state at the moment because I'm not sure how much you want to put into webpacker.

The gist of this is that is removes all JS gems from the gemfile, and moves them to yarn+webpacker.  The exception is jquery, jquery-ujs, tipsy, and spinjs.  Tipsy isn't on npm so we would need to move off of it but I believe you had another library you wanted to use for the tooltips, and both jquery parts can be flipped with 2 lines of code.  SpinJS only has its file in ES6 syntax which currently breaks sprockets.  If you want we can put all the JS completely through webpack and basically forsake sprockets for the JS.  Webpacker can also apparently do CSS and SASS so if we wanted to completely scrap sprockets that's also an option.  If you want to keep sprockets for most things I just need to remove some of the unneeded pieces of code that set up some plugins in webpacker.

Other smaller changes are changing the permissions on the binstubs in /bin since I was already adding the 3 binstubs for webpacker, add a new docker container to run the hot reload for webpacker (anytime anything in /app/javascript changes it will recompile the changes and then automatically reload the browser), and remove bower since as far as I could tell it was only used for bootstrap-toggle which is now in yarn.


BatedUrGonnaDie:
> Can you explain what you mean by this part? You say above that we can't move some libraries into webpack for various reasons, so how would we accomplish this?

Sorry, my explanation wasn't very good since I start interchanging yarn and webpack.  Let me try again.

Everything is in yarn, not webpack except for those libraries I mentioned in the first post.  Currently webpack does not actually compile or provide any JS or CSS at all.  Apparently tipsy is on npm and I'm blind and missed it when I looked last time so that should be able to go into yarn.  When I was talking about jquery's I meant putting those in webpack, Both of those could potentially also be installed off npm into yarn as well.  And finally, spin.js which has es6 syntax in the file where `Spinner` is defined, which prevents that from being installed through npm because sprockets breaks when it hits the export statement.  That could be moved into webpack but then we would have some JS code living in webpack and some living in sprockets and I wasn't sure if that was the best plan.  Definitely do-able though since only the file that defines the functions for spinner would have to go there.

BatedUrGonnaDie:
One thing to clarify though is that you don't need webpacker support to be able to use yarn.  #343 added support for adding npm packages, it just didn't move everything over.  So we would still be able to import everything except spin.js through yarn and remove all of the gems just like this PR does without adding webpack to the project at all.

glacials:
Ok cool, thanks for explaining. So it sounds basically everything except one spin.js file can go into yarn if we want that. But which do you feel is the better setup? Definitely sounds like you now know way more about this than I do now, so I'm down with whichever you believe will be the best option.

BatedUrGonnaDie:
I think having everything in webpacker is going to be the future to be honest.  Since the webpacker gem is just a compat layer on top of webpack, which can be extended by plugins it allows us to write es6 and even es7 right now.  At the very least getting everything into yarn seems like the better than the current way since a lot of the time the random rails gems for the js libraries end abandoned or having extremely slow updates.  I've personally PR'd the c3-rails gem 2 or 3 times to try and get it to the newer versions that fix bugs and improve performance.

glacials:
Ok nice, I'm down with webpack/webpacker then. It would be great to be able to write ES6 and beyond.

BatedUrGonnaDie:
Did you want to have all of the application code transferred into webpack with this PR, or did you just want to expose the libraries to code in sprockets and then slowly move things over to webpack over some time?

I looked through some articles and with exports we could export things into a global variable which would allow one file at a time to be moved.  Or I could just go through and try and put everything into webpack right away and this PR could rip all JS out of sprockets.  Your call on how large you want this PR to be.

glacials:
I vote for exposing the libraries and doing it slowly over time. Just leave instructions somewhere so I don't screw it up when I try to do it :P

BatedUrGonnaDie:
Alright, I believe I got everything to expose itself properly.  Some key notes are that I removed numeraljs since I didn't find any references to it in the JS and that turbolinks is only on npm for v5 and above, which jquery.turbolinks doesn't support.  In order to move that to turbolinks we would need to complete the upgrade off of using `on page:load` and move to `on turbolinks:load`.  You will also need to set the 2 new env variables on aws.  I'm not 100% sure it will allow it to build first try in production but I can troubleshoot it if it fails right away.

BatedUrGonnaDie:
Deploy seems to have gone without a hitch 🎉

PR #347, opened 2018-02-13
Speed up docker builds for machines with more than 2 cores

BatedUrGonnaDie:
Use best practices in dockerfile as well


BatedUrGonnaDie:
This will likely have no impact for the test environments.  Travis definitely only has 2 cores and I would assume docker and aws are the same.  But locally building it this sped up my build times by quite a bit.

PR #348, opened 2018-02-13
Optimize stats page

BatedUrGonnaDie:
27 queries -> 14, time switcher >200ms -> <10ms

### DIFF
--- a/app/controllers/runs/stats_controller.rb
+++ b/app/controllers/runs/stats_controller.rb
@@ -10,50 +10,35 @@ class Runs::StatsController < Runs::ApplicationController
       render 'runs/cant_parse', status: 500
     end
 
-    segments = @run.segments.includes(:histories).map do |segment|
-      segment.attributes.merge(history: segment.histories)
-    end
-
     gon.run = {
       id: @run.id36,
-      segments: @run.segments.map do |segment|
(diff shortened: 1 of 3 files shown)


app/controllers/runs/stats_controller.rb
```diff
@@ -10,50 +10,35 @@ def index
       render 'runs/cant_parse', status: 500
     end
 
-    segments = @run.segments.includes(:histories).map do |segment|
-      segment.attributes.merge(history: segment.histories)
-    end
```
glacials: deng was all this left over from DynamoDB? these queries were literally doing nothing lol

BatedUrGonnaDie: I took a search through all the JS to find if this was used but couldn't find anything.  I would assume at some point the graphs used just `history` but I think you updated them all at one point or another to use `histories`.

BatedUrGonnaDie: This was actually just one query, but because later on it didn't use `includes(:histories)` it ended up being N more queries and not even using the cached query that this one made.  By just adding the include further down it removed most of the queries, and then this one was being cached and used by the later one.  Although I saw no reason to keep this since it was doing nothing.

PR #349, opened 2018-02-16
Fix s3 error in production

BatedUrGonnaDie:
Seems like it behaves differently then fakes3.  This also cleans up some of the rubocop errors in run.rb.

### DIFF
--- a/app/models/run.rb
+++ b/app/models/run.rb
@@ -12,8 +12,8 @@ class Run < ApplicationRecord
   include ActionView::Helpers::DateHelper
 
   # Timing types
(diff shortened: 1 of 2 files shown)


BatedUrGonnaDie:
It looks like they return a different error than the official sdk.  In production if you try to access a file that doesn't exist in the bucket it throws `Aws::S3::Errors::Forbidden` for some reason.  I also don't think the docker image that we use for testing has been updated in a long time.  I believe that it is still on a pre-1.0 version since the licensing change they had.

PR #350, opened 2018-02-17
Move livesplit-core Repository URL

CryZe:
We just moved it over to the LiveSplit GitHub Organization, so splits.io should use these updated URLs too.


BatedUrGonnaDie:
Thanks for the update

PR #351, opened 2018-02-20
Reject splits with 0 segments

BatedUrGonnaDie:
Also fix some rubocop errors


BatedUrGonnaDie:
For the record there are still some rubocop errors in the files I touch like method too long and stuff that would require major changes to the code.  I try and get all the small ones when I can though.

PR #353, opened 2018-02-23
Use js for the convert form

### LINKED ISSUE #338: Convert convert page to api hit
Using html5 saveAs and blobs it's possible to have JS request the convert, and then prompt a save all without having to leave the same page.  This would reduce some code duplication that currently exists for converting runs (didn't know how to do those things when I made them).  Currently there is an api endpoint for 3rd party consumers, and regular controller for converting which is almost identical code.

BatedUrGonnaDie:
Closes #338

### DIFF
--- a/app/controllers/api/v4/converts_controller.rb
+++ b/app/controllers/api/v4/converts_controller.rb
@@ -18,10 +18,10 @@ class Api::V4::ConvertsController < Api::V4::ApplicationController
     )
 
     @run.parse_into_db
-    if @run.program == nil
+    if @run.program.nil?
       render status: 400, json: {
         status: 400,
-        message: "Unable to parse that run."
+        message: 'Unable to parse that run.'
       }
       return
     end
@@ -36,10 +36,14 @@ class Api::V4::ConvertsController < Api::V4::ApplicationController
       new_extension = params[:format]
     end
     filename_without_extension = params[:file].original_filename.split(old_extension)[0]
+    if filename_without_extension.present? && filename_without_extension[-1] == '.'
+      filename_without_extension = filename_without_extension.chop
+    end
 
     filename = "#{filename_without_extension}.#{new_extension}"
+    response.set_header('X-Filename', filename)
 
-    if params[:format] == "json"
+    if params[:format] == 'json'
       new_file = render_to_string(json: @run, serializer: Api::V4::Convert::RunSerializer)
     else
       new_file = render_to_string("runs/#{params[:format]}", layout: false)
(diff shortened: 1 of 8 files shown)


app/controllers/api/v4/converts_controller.rb
```diff
@@ -36,10 +36,14 @@ def create
       new_extension = params[:format]
     end
     filename_without_extension = params[:file].original_filename.split(old_extension)[0]
+    if filename_without_extension.present? && filename_without_extension[-1] == '.'
+      filename_without_extension = filename_without_extension.chop
+    end
```
glacials: are double periods a problem?

app/javascript/convert.js
```diff
@@ -0,0 +1,38 @@
+import {saveAs} from "file-saver";
+
+$(() => {
+  $("form#convert").submit((evt) => {
+    evt.preventDefault();
+    let formData = new FormData();
+    formData.set("file", $("input[type=file")[0].files[0]);
+    formData.set("format", $("select").val());
+    formData.set("historic", $("input[type=checkbox]").is("checked"));
```
glacials: Looks like the history checkbox is removed? But we're still referencing it here

app/javascript/convert.js
```diff
@@ -0,0 +1,38 @@
+import {saveAs} from "file-saver";
+
+$(() => {
+  $("form#convert").submit((evt) => {
+    evt.preventDefault();
+    let formData = new FormData();
+    formData.set("file", $("input[type=file")[0].files[0]);
+    formData.set("format", $("select").val());
```
glacials: Prefer to find elements by ID or unique class, so that if we end up adding another `<select>` to the page (e.g. in the navbar) we don't break this page without realizing it.

app/javascript/convert.js
```diff
@@ -0,0 +1,38 @@
...
+        saveAs(blob, filename);
+        $("#dropzone-overlay").fadeOut(125);
+      },
+      error: (xhr) => {
+        window.isUploading = false;
+        window.spinner.stop();
+        $("#droplabel").text(`Error converting: ${xhr.responseJSON.message}`);
+      }
+    });
+    return false;
```
glacials: Why does this function return false?

app/controllers/api/v4/converts_controller.rb
BatedUrGonnaDie: When I was testing yes.  It splits on the filetype leaving the period.  But sometimes the filename doesn't exist.

app/javascript/convert.js
BatedUrGonnaDie: I'll add some IDs to these form elements and refine the jQuery search.

app/javascript/convert.js
BatedUrGonnaDie: Thought I removed this, will remove the reference to it.

app/javascript/convert.js
BatedUrGonnaDie: I mistakenly thought it was for preventing jQuery propogation, but it is essentially just evt.preventDefault and evt.stopPropogation.  I'll remove the return since it's not needed.

app/controllers/api/v4/converts_controller.rb
glacials: Nah this is fine, was just curious what the need was

PR #354, opened 2018-02-23
Update regex for new format

### LINKED ISSUE #352: Fix speedrun.com URL validations
Speedrun.com URLs can look like `https://speedrun.com/<game>/run/<run>` now, but we only allow URLs that look like `https://speedrun.com/run/<run>`, so Speedrun.com URLs are not attachable to runs anymore without manually hacking the URL up.

BatedUrGonnaDie:
Closes #352

### DIFF
--- a/lib/speedrundotcom.rb
+++ b/lib/speedrundotcom.rb
@@ -15,11 +15,11 @@ module SpeedrunDotCom
       return nil if url.blank?
 
       uri = URI.parse(url)
-      unless uri.host =~ /^(www\.)?(speedrun.com)$/ && uri.path =~ %r{^/run/(.*)$}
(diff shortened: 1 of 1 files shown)


BatedUrGonnaDie:
For reference this adds an optional non-capturing group that eats up anything until the next `/` because AFAIK the shortname for games can be basically anything (IE Agar.io is actually `Agar.io`)

BatedUrGonnaDie:
The `?:` at the beginning means non-capturing which is useful for grouping stuff in my opinion.

glacials:
Oops a bit late, but can you update the error message as well, to use https and the new format? https://github.com/glacials/splits-io/blob/master/app/controllers/runs_controller.rb#L77

BatedUrGonnaDie:
Do you want the message to only have the new format or show both ways to do it?

glacials:
Just the new, since speedrun.com seems to use it / redirect to it in all places.

PR #356, opened 2018-02-27
Add gametime to the v4 api

### LINKED ISSUE #341: Gametime in API
The API only returns realtime, even for runs which have gametime. It should return both, as well as the default timing for that run.

BatedUrGonnaDie:
This also adds back splits to the regular run path. Closes #341


app/controllers/api/v4/application_controller.rb
```diff
@@ -36,9 +36,7 @@ def build_link_headers(links)
   end
 
   def force_ssl
-    if !request.ssl?
-      render status: 301, json: {error: "The Splits I/O API is only accessible over HTTPS."}
-    end
+    render status: 301, json: {error: 'The Splits I/O API is only accessible over HTTPS.'} unless request.ssl?
```
glacials: This was my original goof in v3 that I'm just now noticing, but if this is a 301 it should have a `Location: https://the-url-you-were-trying-to-hit` header set. Or if that's hard to do, just a 400 would do fine.

app/serializers/api/v4/run_histories_serializer.rb
```diff
@@ -0,0 +1,3 @@
+class Api::V4::RunHistoriesSerializer < Api::V4::ApplicationSerializer
```
glacials: Keeping with the pattern of other serializers, this seems like it should be `Api::V4::RunHistorySerializer` (a "history" being a single entry).

app/serializers/api/v4/segment_histories_serializer.rb
```diff
@@ -0,0 +1,11 @@
+class Api::V4::SegmentHistoriesSerializer < Api::V4::ApplicationSerializer
```
glacials: Same comment as above

spec/support/models/api/v4/run.json
```diff
@@ -97,6 +130,17 @@
...
+    "segments": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "required": [
+          "attempt_number",
+          "realtime_duration_ms",
+          "gametime_duration_ms"
+        ]
+      }
```
glacials: As an unwritten rule I like to be as strict as possible in the schema, which means always specifying `"additionalProperties": false` for objects and always fully describing all the field types by using the `properties` key, like you did above for the root run object.

spec/support/models/api/v4/run.json
```diff
@@ -77,11 +109,12 @@
         "$ref": "../../../models/api/v4/runner.json"
       }
     },
-    "history": {
+    "histories": {
       "type": "array",
       "items": {
         "type": "object",
         "required": [
+          "attempt_number",
```
glacials: Also describe this in the `properties` object, 5 lines down

app/controllers/api/v4/application_controller.rb
BatedUrGonnaDie: I'll take a look into it.  Do you also what to have a `location` key in the returned json response?

app/serializers/api/v4/run_histories_serializer.rb
BatedUrGonnaDie: Not really sure what I was thinking with the naming on these, will change to non-plural.

spec/support/models/api/v4/run.json
BatedUrGonnaDie: Got it, will update the schema.

app/controllers/api/v4/application_controller.rb
glacials: Nah just the header is fine.

spec/support/models/api/v4/run.json
```diff
@@ -97,6 +133,31 @@
...
+          "type": "number"
+        },
+          "realtime_duration_ms": {
+            "type": "number",
+            "minimum": 0
+          },
+          "gametime_duration_ms": {
+            "type": "number",
+            "minimum": 0
+          }
```
glacials: Indentation looks a bit off here but otherwise 👌

BatedUrGonnaDie: Whoops

BatedUrGonnaDie:
Still look good with that last change?

Issue #135, opened 2015-03-03
Look at autogenerating API documentation


glacials:
It's a lot of work to bump an API version, and it shouldn't need to be. About half of the work is in just rewriting/updating the documentation. Maybe Red Snow (https://github.com/apiaryio/redsnow) or similar can help.

Issue #129, opened 2015-02-19
Put run disowning in API v4


glacials:
If possible we should also take care of claim token authentication for this endpoint.

PR #358, opened 2018-03-06
Complete APIv4 according to readme specifications

### LINKED ISSUE #249: v4 api not working as documented
A few of the endpoints documented in the v4 api currently seem not to work (using parameters which should return data and do so in v3 and in the browser):
- `games/:game/runners` returns 404
- `runners/:runner` returns 404
- `categories/:category` and all of its routes return 204

BatedUrGonnaDie:
Closes #249

### DIFF
--- a/app/controllers/api/v4/categories/runners_controller.rb
+++ b/app/controllers/api/v4/categories/runners_controller.rb
@@ -1,4 +1,7 @@
 class Api::V4::Categories::RunnersController < Api::V4::ApplicationController
+  before_action :set_category, only: [:index]
+
   def index
+    render json: @category.users, each_serializer: Api::V4::RunnerSerializer, root: 'runners'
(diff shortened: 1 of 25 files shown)


BatedUrGonnaDie:
There's still #129 and #135  left in the issue tracker for v4.  Not sure if you want to block until those are wrapped up as well.

glacials:
Oh yeah. #135 shouldn't block, I'll leave it up to you if #129 should.

BatedUrGonnaDie:
I'll take a look into #129 and if it's easy enough bang that out before removing the notice.  If not it can always be added into v4 later since it most likely won't be a breaking change.  Definitely want to go through the readme again with a fine tooth comb when before that's removed though to make sure it's absolutely 100% accurate and nothing is missing.

PR #359, opened 2018-03-14
Redo all graphing to make stats page somewhat usable

BatedUrGonnaDie:
Currently this implements all graphs in the same panel as the c3 graphs to make it clear what the difference do.  Still loads all data from gon which will need to be switched out for an api call.


app/javascript/packs/application.js
```diff
@@ -7,30 +7,35 @@
...
+    $('input[type=checkbox][data-toggle^=toggle]').bootstrapToggle();
+});
 
+import "../convert";
+import "../highchart_theme.js";
+import "../graphs/pb_graph.js";
+import "../graphs/segment_history_mean_graph.js";
+import "../graphs/run_duration_graph.js";
+import "../graphs/segment_duration_graph.js";
+import "../graphs/reset_graph.js";
```
glacials: I forget if we're on ES6, but if we are semicolons are optional and generally not used

BatedUrGonnaDie: Anything in `app/javascript` is part of webpack so definitely es6 there, I just wasn't sure what the standard was so I opted to put them just in case.

BatedUrGonnaDie:
Generally speaking the new reset implementation should be more accurate, but I'll need to take a look at the raw file and make sure that the values shown are indeed accurate.  There may be some rounding differences now because I use raw millisecond values going in to the graphs and round to one decimal place in highcharts.

BatedUrGonnaDie:
I'll just address each issue in a different comment.

For the PB graph none of the bars are 0 for me, they are just so small because the last split has all the splits rolled into it which is skewing the graph so high.  I added an x zoom option to the graphs so that you can filter out those high values or if you want to just look at a small subset of splits.

BatedUrGonnaDie:
For the reset graph it looks like that file has some corrupted history because Ricco 2 and Ricco 4 both have substantially larger histories than Ricco 3.  From my understanding history objects are created if you get past that split, which means that anything past the split must have an equal number or less history objects and that's what the new graph calculates based off of.

BatedUrGonnaDie:
For split history forgot to filter out the 0's before taking the mean of the array, so just did a simple `Array.filter` to remove them.

BatedUrGonnaDie:
There is one more regression which is that on the segment duration over time graph the attempts are all continous, which will be fixed when loading the data from the api because currently the attempt number is not provided through the client, but is in the api so it seemed easier to just wait to fix that until we switch later in this branch.

BatedUrGonnaDie:
Alright, fixed the attempt counts on the segment history duration graph with the fetch data, while also adding local export options to all the graphs (can save it natively as a png, jpg, and a few others).  Also removed all the semicolons.

BatedUrGonnaDie:
I believe this should be ready to go after a re-review.  Indicators aren't the greatest but I'm not very good with UI and it's better than nothing.

A good example splits on the live site I've been using is https://splits.io/20ra .

PR #360, opened 2018-03-19
Correctly sum all segment's gameplay bests when parsing

### LINKED ISSUE #334: Sum of best and total potential timesave always use Real Time
When viewing the stats page under Game Time, the Sum of Best and Total Potential Time Save are wrong because they always just use Real Time.

Real Time:
chrome_2017-10-30_00-34-22

Game Time:
chrome_2017-10-30_00-34-06

The splits:
https://splits.io/1vcq/stats

julioolvr:
Hi! First of all I just wanted to say - great codebase, it was super easy getting into and the fact all the environment is setup by Docker makes everything extremely simple 💯

I'll leave some open questions directly on the related lines, but overall if the fix makes sense then I wonder if you want a migration in order to fix previous runs. I honestly learned about the existence of this site _yesterday_ so I have no idea how much data is in there and how long would such a migration take to complete, but let me know your thoughts.

Fixes #334

### DIFF
--- a/app/models/concerns/unparsed_run.rb
+++ b/app/models/concerns/unparsed_run.rb
@@ -85,6 +85,7 @@ module UnparsedRun
 
         write_segments(segments)
         write_segment_histories(segments)
+        reload
 
         all_segments_have_bests = segments.map.all? do |segment|
           segment.realtime_best.present?
@@ -94,8 +95,8 @@ module UnparsedRun
         gametime_best_times = nil
 
         if all_segments_have_bests
-          realtime_best_times = extract_best_times(segments, :realtime)
-          gametime_best_times = extract_best_times(segments, :gametime)
+          realtime_best_times = extract_best_times(Run::REAL)
+          gametime_best_times = extract_best_times(Run::GAME)
         end
 
         update(
@@ -200,8 +201,8 @@ module UnparsedRun
       SegmentHistory.import(histories)
     end
 
-    def extract_best_times(segments, timer)
-      segments.map(&:"#{timer}_best").sum * 1000
+    def extract_best_times(timing)
+      segments.map { |segment| segment.shortest_duration_ms(timing) }.sum
     end
   end
 end
--- a/app/models/concerns/unparsed_run.rb
+++ b/app/models/concerns/unparsed_run.rb
@@ -90,14 +90,12 @@ module UnparsedRun
           segment.realtime_best.present?
         end
 
-        sum_of_best_seconds = nil
-        sum_of_best_milliseconds = nil
-        if all_segments_have_bests
-          sum_of_best_seconds = segments.map(&:realtime_best).sum
+        realtime_best_times = nil
+        gametime_best_times = nil
 
-          if sum_of_best_seconds.present?
-            sum_of_best_milliseconds = sum_of_best_seconds * 1000
-          end
+        if all_segments_have_bests
+          realtime_best_times = extract_best_times(segments, :realtime)
+          gametime_best_times = extract_best_times(segments, :gametime)
         end
 
         update(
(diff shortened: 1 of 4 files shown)


spec/factories/run_factory.rb
```diff
@@ -16,6 +16,9 @@
     # a bad run and a better run from the same category, against which suggestions can be generated
     compare_subject:          {filename: 'compare_subject.lss'},
     compare_object:           {filename: 'compare_object.lss'},
+
+    # specific-content runs
+    with_segments_bests:       {filename: 'with_segments_bests.lss'}
```
julioolvr: I took the file from the actual bug report, not sure if that's an overkill. I think of this in the context of the factory as a file that represent a specific scenario to test, maybe in the future some more will show up, but let me know your thoughts.

spec/models/run_spec.rb
```diff
@@ -89,6 +89,24 @@
     end
   end
 
+  context 'with bests in all segments' do
+    let(:run) { FactoryBot.create(:with_segments_bests_run) }
+
+    it 'has a realtime sum of best in ms' do
+      run.parse_into_db
+      run.reload
+      expect(run.realtime_sum_of_best_ms).to be_within(run.segments.count)
```
julioolvr: I had to use `be_within` here because rounding is done differently when storing the sum_of_best counts and when storing each individual segment. Rounding is actually done by truncating in both cases, but for sum_of_best we first sum all times and then truncate, and for segments we truncate each time individually.

In the sample file, there are times like `132.740999`. This means that the corresponding segment gets saved as `132740`, but summing up a couple of those makes the sum_of_best result a couple of seconds longer. I wonder if it would make sense to round instead of truncate when saving the segments (`132.740999` would be `132741` instead of `132740`) but that would be a way bigger change than what I intended to do so I assume there can be at most a difference of milliseconds equal to the number of segments in the worst case.

app/models/concerns/unparsed_run.rb
```diff
@@ -201,5 +199,10 @@ def write_segment_histories(segs)
 
       SegmentHistory.import(histories)
     end
+
+    def extract_best_times(segments, timer)
+      seconds = segments.map(&:"#{timer}_best").sum
+      { seconds: seconds, milliseconds: seconds * 1000 }
+    end
```
glacials: This change is innocuous by itself, but combined with a second or third future innocuous change it can become dangerous! If a future caller behaves poorly like say
```ruby
extract_best_times(segments, params[:timing])
```
then it can open up a way for users to call unallowed methods on `Segment`s by adding `?timing=anything` to their URL -- and we might not realize it because each piece of code on its own looks relatively safe.

One possible solution is to use [`Segment#shortest_duration_ms`][1], which can take a timing (`Run::REAL` / `Run::GAME`) as an argument and behaves ok if it receives something else.

PS: In this repo "timer" refers to the program (LiveSplit etc.) and "timing" refers to the type of time-passing (realtime, gametime).

[1]: https://github.com/glacials/splits-io/blob/master/app/models/segment.rb#L34-L41

glacials:
As for migrations, the number of runs is big enough that I say skip it for now. I've been on-and-off fiddling with periodically reparsing every run, but until then I can just tweet after this is merged with basically "If you have a preexisting run you want to see gametime sum-of-bests for, click the reparse button". And eventually when I get periodic reparsing up, every run will get brought up to date :D

app/models/concerns/unparsed_run.rb
julioolvr: Goot thinking 👍 I'm still looking into this one though - I just learned that what I called the parameter `segments` here is actually an array of `Split`s, which as far as I can see are a thin wrapper around the parsed hash generated by `Parser`. So I see two alternatives:

* To add a method similar to `Segment#shortest_duration_ms` to `Split`
* To do a `reload` after calling `segments.import` [here](https://github.com/julioolvr/splits-io/blob/18cc412cc8456b43d1909c56ae6c383bfa47faca/app/models/concerns/unparsed_run.rb#L158). That will make the segments available from `Run#segments` and I can then use `Segment#shortest_duration_ms`.

I'm leaning towards the latter but I'm not sure yet, let me know what you think but in any case I'll think about it some more with my morning coffee ☕️

glacials: Aha, I forgot about `Split`! I say do the latter -- `Split` used to serve a bigger purpose before we switched to livesplit-core for parsing, and now I'm not sure what I want it to be: thin wrapper with no logic, or completely gone. Either way, the second sounds best!

You may even be able to do it without hard reloading if `write_segments` can be modified to return the `Segment`s it imported.

julioolvr: Alright, I ended up going with a `reload`, since `activerecord-import` doesn't seem to return instances of the records it saved (which makes sense since they don't seem to instantiate those in memory). I could've done something like finding all the segments by the ids returned by `segments.import` but I feel like it would be pretty much the same as doing `reload` and the latter sounds simpler. That being said I think it's objective so I'm open to changing it.

glacials: The import call itself takes a bunch of `Segment`s as arguments, are those reusable as return values?

julioolvr: 🤦‍♂️

Yup, that's right. I'll use those, rename things a bit and send a new commit 👍

julioolvr: There! I only renamed `segments` to `splits` since I know have the `segments` variable which holds the `Segment` instances in `#parse_into_db`. It might make sense to actually use those `Segment` instances once they are generated (instead of mixing both during the rest of the method) but I wanted to keep the changes to a minimum for now.

PR #361, opened 2018-03-21
Update all the things again

BatedUrGonnaDie:
Since I can't really test this myself, I'm pushing the upgrade for the patreon gem in a separate commit since it had issues thats time.  If it's not too much work could you test it out for me @glacials ?  Otherwise I can just remove that commit and it can be reviewed based solely on the first commit.


BatedUrGonnaDie:
I really wish any services had a "test integration" button so you could make sure what you're writing is actually working.  I'll just leave the patreon gem pinned at the working version since I don't really want to break payment processing stuff and I don't think there is a super convincing reason to test it right now.  Sometime tomorrow I'll rollback that commit and fix the merge conflicts.

PR #362, opened 2018-03-23
Optimize api stuff

BatedUrGonnaDie:
Authie will still trigger session stuff if you visit it logged in while in a browser, but this should fix it for all consumers of the api.

### DIFF
--- a/app/controllers/api/v3/application_controller.rb
+++ b/app/controllers/api/v3/application_controller.rb
@@ -1,6 +1,8 @@
 class Api::V3::ApplicationController < ActionController::Base
   include Rails::Pagination
 
+  skip_before_action :set_browser_id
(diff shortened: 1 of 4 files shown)


BatedUrGonnaDie:
Blocked for now by weird behavior in authie.  Waiting to see if a total kill switch can be configured per controller.

BatedUrGonnaDie:
Looks like as soon as you call `current_user` authie will run, so it will be disabled on all endpoints except the ones that need it still.

BatedUrGonnaDie:
Did you want me to include the authie skip in health controller in this PR?

glacials:
Sure, might as well.

BatedUrGonnaDie:
Should be good to go with that now I believe.

PR #368, opened 2018-03-29
Optimize title partial rendering time

BatedUrGonnaDie:
I had noticed that it seemed to be taking the majority of the wall clock time spent for all pages that had the title partial, so I spoofed ~11k runs locally to get larger search times locally.  This change reduced render times ~50% (usually 550ms -> 250ms) since it didn't have to load all the runs and I believe it also skips converting the active record collection to a ruby array to perform `present?`.

I also cleaned up the run title code on the title partial since it was basically duplicated since `Run.to_s` was implemented.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -6,7 +6,8 @@ class Ability
 
     can :read, [Run, Category, Game, User]
 
-    can [:update, :create, :destroy], Run.owned, user_id: user.id
+    can [:update, :create, :destroy], Run, user_id: user.id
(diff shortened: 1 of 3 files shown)


BatedUrGonnaDie:
Looks like I was able to reduce the query that was ruining performance, only takes 5Xms now consistently down from ~600 with my 11k runs.

I also switch to `.any?` because to me that seems more logical and doesn't increase times at all.

BatedUrGonnaDie:
It's nice because this will also speed up the stats page as well, not just the run page.

Issue #331, opened 2017-10-19
Rubocop!


glacials:
Clean up the whole app by getting Rubocop runs to pass, then add it to the testing flow!

PR #370, opened 2018-04-04
Fixes a lot of rubocop warnings

BatedUrGonnaDie:
Progress towards #331, fixes about 700ish warnings overall.  Most of the warnings remaining are methods/blocks too long/complex.

### DIFF
--- a/app/controllers/api/v3/runs_controller.rb
+++ b/app/controllers/api/v3/runs_controller.rb
@@ -4,9 +4,7 @@ class Api::V3::RunsController < Api::V3::ApplicationController
 
   before_action only: [:create] do
     # If an OAuth token is supplied, use it (and fail if it's invalid). Otherwise, upload anonymously.
-    if request.headers['Authorization'].present?
-      doorkeeper_authorize! :upload_run
-    end
+    doorkeeper_authorize! :upload_run if request.headers['Authorization'].present?
   end
 
   def show
@@ -29,7 +27,7 @@ class Api::V3::RunsController < Api::V3::ApplicationController
 
     render status: 201, location: api_v3_run_url(@run), json: {
       status: 201,
-      message: "Run created.",
+      message: 'Run created.',
       id: @run.id36,
       claim_token: @run.claim_token,
       uris: {
@@ -48,19 +46,13 @@ class Api::V3::RunsController < Api::V3::ApplicationController
   # TODO: These methods are not documented and thus is not officially supported by the API. It uses cookie
   # authentication, but the official release of it into the API should use token authentication.
   def destroy
-    if !@run.destroy
-      raise "Can't destroy run"
-      return
-    end
+    raise "Can't destroy run" unless @run.destroy
 
     head 200
   end
 
   def disown
-    if !@run.update(user: nil)
-      raise "Can't disown run"
-      return
-    end
+    raise "Can't disown run" unless @run.update(user: nil)
 
     head 200
   end
@@ -82,7 +74,6 @@ class Api::V3::RunsController < Api::V3::ApplicationController
         status: 401,
         message: "Run with id '#{params[:id]}' is not owned by you. You must supply a cookie proving your are the owner of this run in order to disown or delete it."
       }
-      return false
     end
   end
 end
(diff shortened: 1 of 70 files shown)


app/controllers/api/v3/runs_controller.rb
```diff
@@ -4,9 +4,7 @@ class Api::V3::RunsController < Api::V3::ApplicationController
 
   before_action only: [:create] do
     # If an OAuth token is supplied, use it (and fail if it's invalid). Otherwise, upload anonymously.
-    if request.headers['Authorization'].present?
-      doorkeeper_authorize! :upload_run
-    end
+    doorkeeper_authorize! :upload_run if request.headers['Authorization'].present?
```
glacials: I have no idea of the precedence Ruby puts on stuff like this, is this equivalent to
```ruby
doorkeeper_authorize!(:upload_run) if request.headers['Authorization'].present?
```
or
```ruby
doorkeeper_authorize!(:upload_run if request.headers['Authorization'].present?)
```
? Either way I don't like the ambiguity to humans, so let's be explicit when we have situations like this.

app/controllers/api/v3/users/games/categories/pbs_controller.rb
```diff
@@ -32,7 +32,7 @@ def set_category
   end
 
   def set_pb
-    unless @pb = @user.pb_for(@category)
+    unless (@pb = @user.pb_for(@category))
```
glacials: Let's actually ditch this whole `[if/unless] thing = other_thing` idea, this is too easy to misread as `[if/unless] thing == other thing`, or to read it correctly and think it was a typo.

app/controllers/api/v4/application_controller.rb
```diff
@@ -76,11 +75,11 @@ def set_runner
...
-    if params[:historic] == '1'
-      @run = Run.includes(:game, :category, :user, :histories, segments: [:histories]).find36(params[:run])
-    else
-      @run = Run.includes(:game, :category, :user, :segments).find36(params[:run])
-    end
+    @run = if params[:historic] == '1'
+             Run.includes(:game, :category, :user, :histories, segments: [:histories]).find36(params[:run])
+           else
+             @run = Run.includes(:game, :category, :user, :segments).find36(params[:run])
+           end
```
glacials: Looks like there's an extra `@run = ` in here (although it incidentally probably still works as intended)

app/controllers/api/v4/runs_controller.rb
```diff
@@ -9,9 +9,7 @@ class Api::V4::RunsController < Api::V4::ApplicationController
   before_action only: [:create] do
     if current_user.nil? # If a cookie is supplied, use it because we're probably on the website.
       # If an OAuth token is supplied, use it (and fail if it's invalid). Otherwise, upload anonymously.
-      if request.headers['Authorization'].present?
-        doorkeeper_authorize! :upload_run
-      end
+      doorkeeper_authorize! :upload_run if request.headers['Authorization'].present?
```
glacials: Same as above

app/models/run.rb
```diff
@@ -80,13 +80,13 @@ def program_from_attribute(func_name, value)
...
 
-    alias_method :find10, :find
+    alias find10 find
     def find36(id36)
       find10(id36.to_i(36))
     end
   end
 
-  alias_method :id10, :id
+  alias id10 id
```
glacials: Whoa, didn't know this was a thing

spec/controllers/api/v4/categories/runs_controller_spec.rb
```diff
@@ -6,7 +6,7 @@
       let(:category) { create(:category, :with_runs) }
       subject { get :index, params: {category: category.id} }
 
-      it 'retruns a 200' do
```
glacials: Wow did you catch this, or did Rubocop? Either way, amazing

BatedUrGonnaDie: I actually saw it while clicking through the different files and noticed that something was changing even though the wording looked identical XD

app/controllers/api/v3/runs_controller.rb
BatedUrGonnaDie: Yeah, I agree that it looks cleaner when its more laid out, and the precedence might change.

app/controllers/api/v3/users/games/categories/pbs_controller.rb
BatedUrGonnaDie: I personally like when things are on separate lines as well since it reduces confusion for myself because I assume that it's a boolean expression.

PR #373, opened 2018-04-05
Save gametime_duration_ms for each segment history entry

BatedUrGonnaDie:
Turns out we never actually saved these, no wonder they were always 0 in my testing before XD


BatedUrGonnaDie:
I think these tests should cover most cases where something could go wrong, but I can always check every single history if you want complete assurance.

BatedUrGonnaDie:
Is the failing docker cloud build something that I need to address?  If I click on the details link then it just takes me to a 404 page so I'm not sure if it's something that I broke.

glacials:
No don't worry about it, I think I disconnected the build from here, not sure why it's still showing. If travis passes you're fine.

PR #374, opened 2018-04-05
Move segment history graphs to run page

glacials:
Moves segment history graphs from the run stats page to the run page. This is a redo of #321 to work with Highcharts.

screenshot 2018-04-05 10 18 38


app/assets/stylesheets/application.sass
```diff
@@ -38,11 +38,13 @@ ul.goals
...
-  -moz-animation: spin 4s linear infinite
-  animation: spin 4s linear infinite
+  animation: rotation 4s infinite linear
 
-@-webkit-keyframes spin { 100% { -webkit-transform: rotate(-360deg); } }
+@-webkit-keyframes rotation
+  from
+    -webkit-transform: rotate(0deg)
+  to
+    -webkit-transform: rotate(-359deg)
```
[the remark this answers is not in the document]
BatedUrGonnaDie: It spins like that for me in chrome, not sure if it is browser specific issue though.

app/javascript/segment_graph_builder.js
```diff
@@ -0,0 +1,73 @@
...
+  }).catch(function(error) {
+    spinners.forEach(function(spinner) {
+      spinner.stop()
+    })
+    for(const graphHolder of document.getElementsByClassName('graph-holder')) {
+      graphHolder.find('.panel-title').find('h1').text('Error retrieving data for graphs')
+    }
+  })
+
+  if (document.getElementById('run-timer').dataset.timer === 'livesplit') {
```
BatedUrGonnaDie: Is there a reason to not use gon for this as it is used above to get the run id?

app/javascript/segment_graph_builder.js
```diff
@@ -0,0 +1,73 @@
...
+    headers: {
+      accept: 'application/json'
+    }
+  }).then(function(response) {
+    if (response.ok) {
+      return response.json()
+    }
+    throw new Error('Request for run from api failed')
+  })
+
```
BatedUrGonnaDie: Extra whitespace on blank line.

app/javascript/segment_graph_builder.js
```diff
@@ -0,0 +1,73 @@
+import {build_segment_graph} from "graphs/segment_duration_graph.js"
+import {Spinner} from 'spin.js'
+
+$(function() {
+  const graphHolders = document.getElementsByClassName('segment-graph-holder')
+  if(graphHolders.length === 0) {
+    return
+  }
+
+  let spinners = []
```
BatedUrGonnaDie: This should probably be `const` as well since we don't want to change the actual value at all.

app/views/runs/show.slim
```diff
@@ -12,6 +12,8 @@
 
 = render partial: 'runs/full_timeline', locals: {run: @run, timing: timing}
 
+#run-timer data-timer=@run.timer
```
BatedUrGonnaDie: See above gon question.

app/assets/javascripts/events.js
```diff
@@ -0,0 +1,12 @@
+$(function() {
+  // Set up our own class of events we can listen for to load some JS only on specific pages.
```
BatedUrGonnaDie: This can be achieved by creating another pack file and only including it on the pages that we want that js served on. Do you think that's a better solution long term then using events to manage it?

app/assets/stylesheets/application.sass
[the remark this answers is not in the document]
glacials: I'm in Chrome too and it doesn't 🤔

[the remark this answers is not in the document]
BatedUrGonnaDie: Oh, did you mean before this change?  Then yeah it didn't spin 360 and this change fixes that.  My bad thought you meant even with this change it didn't work.

app/javascript/segment_graph_builder.js
glacials: Hmm, I was originally thinking to migrate off gon to keep us sane about passing variables through like this (so we don't do the accidentally-load-every-segment-history thing), but now that I think about it, it's just as easy to accidentally pass too much through this way also, so there's really no benefit. We will just have to be careful in PRs about catching that kind of stuff.

app/assets/javascripts/events.js
glacials: Will that work for Turbolinks transitions also?

app/javascript/segment_graph_builder.js
glacials: Wow I honestly did not understand `const` until this prompted me to go look it up, changed!

app/assets/javascripts/events.js
BatedUrGonnaDie: On second thought I think that using multiple pack files for each page is worse.  Since Rails doesn't support HTTP/2 it would just increase load times by needing to download multiple JS files.  Technically I think it does work with current turbolinks version as it should check the header section and merge any changes that it sees, but I'm not sure what the behavior is of old versions so I wouldn't quote me on that.

Unrelated to that though, is there a reason to keep this in sprockets as opposed to putting it in webpack?  Also I don't see anywhere in the code that is using this new event system, is this for future proof or am I missing something?

BatedUrGonnaDie: Make sure to remember that idea.  Might come in handy for some other pages :+1:

PR #376, opened 2018-04-10
Reject negative attempt counts

### LINKED ISSUE #372: Attempt number can show negative on Segment Duration Over Time graph
image

https://splits.io/2696/stats?timing=real

BatedUrGonnaDie:
Closes #372 .

I checked the test files manually to make sure that they did have negative attempts, everything checked out.

### DIFF
--- a/spec/models/run_spec.rb
+++ b/spec/models/run_spec.rb
@@ -177,7 +177,7 @@ describe Run, type: :model do
     end
 
     it 'reports its total playtime' do
-      expect(run.total_playtime_ms(Run::REAL)).to eq 77_743_500
+      expect(run.total_playtime_ms(Run::REAL)).to eq 77_066_934
     end
 
     it 'reports no history using fast parsing' do
@@ -370,7 +370,7 @@ describe Run, type: :model do
 
       it 'has the correct realtime segment histories for the last segment' do
         expect(run.segments.last.histories.map { |s| s.duration_ms(Run::REAL) }).to match_array [
-          0, 126_778, 126_655, 124_846, 155_191, 140_486, 122_007, 143_338, 132_288, 123_207, 177_090, 164_684, 145_047,
+          126_655, 124_846, 155_191, 140_486, 122_007, 143_338, 132_288, 123_207, 177_090, 164_684, 145_047,
           123_260, 125_012, 147_050, 126_685, 128_863, 141_873, 141_239, 157_672, 136_347, 145_621, 139_113, 148_362,
           146_256, 125_921, 123_539, 129_806, 138_700, 141_579, 149_764
         ]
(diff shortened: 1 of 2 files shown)


spec/models/run_spec.rb
```diff
@@ -177,7 +177,7 @@
     end
 
     it 'reports its total playtime' do
-      expect(run.total_playtime_ms(Run::REAL)).to eq 77_743_500
+      expect(run.total_playtime_ms(Run::REAL)).to eq 77_066_934
```
glacials: If I understand route changes and negative attempts correctly, this number could more or less reset to 0 if a game's entire route changes, is that correct? If that's the case then we should do a small-to-medium amount of work to try to prevent that. Maybe calculating playtime at parse-time and storing it in the `runs` table, or maybe including negative attempts in parsing but using a `scope` on `SegmentHistory` to exclude them?

BatedUrGonnaDie: According to Woofer the negative attempts are purely synthetic, and never a time that was actually recorded by the runner.  Do you still want to count those as part of the playtime?

glacials: Interesting, in that case then nevermind. Do you know what the actual attempt is for? And what its duration means?

BatedUrGonnaDie: From @wooferzfg directly on discord:

those are not "real" attempts and only exist in the history to maintain the accuracy of the sum of best segments

sum of best uses pb splits, best segment times, and the segment history. the problem is that there are certain cases where you can modify your splits and lose data. in these cases, the negative indexed elements get added

case 1: if you pb, but your pb segments are not in the segment history. those used to be in the sum of best, but after you pb, those segments are gone. so they get added to the segment history

case 2: you insert a new segment C between segments A and B and the best segment for segment B isnt in the history. so a negative indexed element gets added that goes from A -> B
these are the only cases where we do it
on deletion we don't do it because the user actually wants the data to be gone

glacials: Hmm, I think I understand case 2, but I don't understand what the negative attempts are eventually used for by LiveSplit (if anything?).

I don't understand case 1. That completely went over my head, I think need an ELI5.

wooferzfg: The negative attempts are mostly just used to calculate the sum of best. They're also used if you remove a segment: if you have segments A, B, and C, and you remove B, the new best segment from A to C will be calculated by finding the best combined A to C segment that any attempt had in the segment history, which includes the negative attempts.

wooferzfg: Also, both of the cases that I described on Discord are only relevant if you manually enter some of your PB split times or best segment times since this results in some information that is used in the Sum of Best calculation but is not in the segment history. The goal of the negative attempts is to ensure that this information is not lost.

glacials: I see, thanks for the explanations! So as a summary it sounds like attempts < 0 are legitimate, but just don't fit cleanly into a chronological ordering because they were manually entered or imported from a timer that doesn't have the right metadata, and they need to be kept around because they may be the best attempt on a segment?

If that's correct, @BatedUrGonnaDie you can go ahead and merge this to fix the original graph bug, but let's also make an issue to use livesplit-core's total playtime feature so we don't have to reinvent their wheel. And maybe let's also document somewhere the stuff that took us this much talking to figure out :P

wooferzfg: Yes, that's correct.

PR #380, opened 2018-04-20
Use sort_by instead of order to reduce queries

BatedUrGonnaDie:
Using `order` creates a new sql query for every segment to allow the db to do the sorting, whereas `sort_by` will convert the results from preloading into a standard ruby array and sort.


BatedUrGonnaDie:
One would think rails would be smarter about it when preloading stuff, but it seems not.

PR #381, opened 2018-05-02
Add search to users

BatedUrGonnaDie:
Removes `games#index` and moves all that functionality into the search controller as well.

`pg_search` was chosen over other gems (mainly `textacular`) because it is more powerful in what it and can do, and returns everything as named scopes which allows nice chaining in ActiveRecord, which textacular would throw errors and doesn't seem like it will get a patch anytime soon for that.

A side effect of this is that game searching "should" be fast because it is now using trigram search for anything that doesn't fully match the shortname, which uses a basic search.  Since this is a bit more strict in finding results compared to the previous method (which would match anything that had the string contained), this should result in less games being matched unless they contain the same long substring in their name.

Also I checked before adding trigram and AWS does support quite a few pg extensions, the trigram module being one of them.  However I do not know if it requires any extra setup outside of just running the migration on RDS.


config/routes.rb
```diff
@@ -36,9 +36,9 @@
 
   get  '/convert',    to: 'converts#new'
 
-  get '/search',        to: redirect('/games') # all deprecated, use GET /games
-  get '/search/:q',     to: redirect('/games?q=%{q}')
-  get '/search(?q=:q)', to: redirect('/games?q=%{q}')
+  get '/search',        to: 'search#index'
+  get '/search/:q',     to: 'search#index'
+  get '/search(?q=:q)', to: 'search#index'
```
glacials: It would be nice if one of these
```ruby
get '/search/:q'
get '/search(?q=:q)'
```
redirected to the other, so we can have one de facto URL style for the search results page. I have no strong preference for which one is the standard.

config/routes.rb
```diff
@@ -68,7 +68,6 @@
 
   get '/users/:user/pbs/:game/:category(/*trailing_path)', to: 'users/pbs#show', as: :user_pb
 
-  get   '/games',            to: 'games#index', as: :games
```
glacials: I know it's temporary until we add the game directory, but can we support this as a redirect to `/search` in the meantime? Just in case bookmarks / tweets / etc. still use it.

app/views/games/categories/show.slim
```diff
@@ -1,7 +1,7 @@
 - title @game.name
 - content_for(:header) do
   ol.breadcrumb
-    li.breadcrumb-item = link_to "Games", games_path
+    li.breadcrumb-item = link_to "Games", search_path
```
glacials: Additional comment below on `config/routes.rb` but if we keep the `/games` route around as a redirect to `/search` until we add the games directory, we can keep using `games_path` in all these places and we won't forget to change it back when the games directory at `/games` is added.

app/views/layouts/application.slim
```diff
@@ -49,8 +49,13 @@ html
...
               li.nav-item class=('active' if on_page[:faq])
                 a.nav-link.px-3 href=faq_path FAQ
-              = form_for(:game, method: :get, url: games_path, html: {class: 'form-inline ml-2 flex-grow-1', role: 'search'}) do |f|
+              = form_for(:search, method: :get, url: search_path, html: {class: 'form-inline ml-2 flex-grow-1', role: 'search'}) do |f|
                 .input-group
+                  select.input-group-addon name='search_type'
+                    option value='game' selected=('true' if @search_type == 'game')
+                      | Game
+                    option value='user' selected=('true' if @search_type == 'user')
+                      | User
```
glacials: Is it possible to ditch the select boxes and always search both users and games, displaying both types of results together on the search results page? imo the fewer clicks/keypresses/decisions users have to make, the better.

config/routes.rb
BatedUrGonnaDie: I personally like `get '/search(?q=:q)'` if that's cool with you.

config/routes.rb
BatedUrGonnaDie: Yeah, this should also fix the error on the home page then since it's related to `games_path` not existing anymore.

app/views/games/categories/show.slim
BatedUrGonnaDie: Do you want this to just be a simple redirect in the route file like search used to redirect to games?

app/views/layouts/application.slim
BatedUrGonnaDie: Are you ok with this being in different tabs like the categories on the games page?  I feel like having everything in one long list would get a bit cluttered especially if you plan on adding search to anything else in the future.

app/views/games/categories/show.slim
glacials: Yeah that's perfect.

app/views/layouts/application.slim
glacials: hmmmm. imo putting them in different tabs is better than the selectors (one less decision to make before performing the search) but still not as good as all-in-one-page (0-1 more clicks to get to the result you want). Can the clutter issue be solved with pagination?

The eventual future I want is like how the new twitch.tv search works, with a few results from each type immediately visible, and you can click "more" on any result type to see a fuller list (see below).

This is obviously way out of scope, but small steps toward it will help.

screenshot 2018-05-02 09 41 58

screenshot 2018-05-02 09 54 18

BatedUrGonnaDie:
I'm not sure how to cleanly add the run table in when there are different things to search on.  Should it show runs from just games, just users, mix them up in some way, show neither, maybe have tabs/pills?

Everything else though should be addressed though.

BatedUrGonnaDie:
I didn't think about it that way.  Makes total sense when looking at an overview of the landscape.

PR #382, opened 2018-05-02
Add .glyphicon-new-window after external links

glacials:
Adds the new-window glyphicon after links that open in new windows, like in Wikipedia. Also makes some external links that weren't opening in new windows, open in new windows.

Looks like: 
screenshot 2018-05-02 11 46 31


app/assets/stylesheets/links.sass
```diff
@@ -14,6 +14,8 @@
   &:hover
     color: $green
 
-.external-link
-  &:hover
-    color: #000
+a[target="_blank"]:not(.carbon-img):not(.discord):after, a[href^="mailto:"]:after
+  font-family: 'Glyphicons Halflings'
```
BatedUrGonnaDie: As far as I'm aware we don't have glyphicons in the app anymore since bootstrap 4 dropped it due to licensing issues.  On my machine these icons also don't load for me.  When I did the move to BS4 I opted to use font-awesome, which it looks like you can do the same thing with them.  [Example Code for FA](https://fontawesome.com/how-to-use/web-fonts-with-css)

Looks like a similar icon is [this guy](https://fontawesome.com/icons/external-link-alt?style=solid).  If you wanna stick with glyphicons though instead of FA then you'll probably have to readd it through a gem and you could also swap out the FA icons back to glyphicons.

glacials: Oh wow I had no idea. So does it only work for me because I have the font installed or something?

Either way yeah I'll switch this to font-awesome, thanks for the explanation!

BatedUrGonnaDie: You might have the glyphicon font installed on your system which would explain it, that should let it load properly for you but not me.

BatedUrGonnaDie:
Looks all good now 👍

PR #383, opened 2018-05-04
Minor bootstrap 4 update formatting mishaps

BatedUrGonnaDie:
Currently it looks super pushed together, this just adds a bit more space to match live site spacing.


BatedUrGonnaDie:
I also saw some errors with the row formatting in the rivals page so I cleaned that up.

BatedUrGonnaDie:
Yeah, that was a problem for me as well and only caught these in the beta deploy since it had everything setup already from my really account.

PR #386, opened 2018-05-17
Fix AWS worker error

BatedUrGonnaDie:
When AWS clones in the repo, the `bin/` directory loses its executable status, which causes bundle to error on the binstub.  By passing it to ruby like you would on windows it fixes the error.


BatedUrGonnaDie:
The binstubs should load everything.  On Windows I know to start the binstubs you just pass it to Ruby directly.

PR #393, opened 2018-06-02
fix name error and no method error

BatedUrGonnaDie:


BatedUrGonnaDie:
I just saw it when I was looking at the twitch follow error on rollbar ¯\_(ツ)_/¯

PR #395, opened 2018-06-04
Theme change & some run page stats

glacials:
Changes the site to the "Dashboard" (sometimes called "toolkit") Bootstrap 4 theme. Adds some stats to the run page, continuing the effort of rolling the stats & compare pages into the run page.

## Current site:
screenshot 2018-06-03 18 24 29

## This branch:
screenshot 2018-06-03 18 23 20

Several additional smaller things:
- Add a Makefile for easier in-container commands (`make test`, `make rubocop`, etc.)
- Add a Rubocop section to the README under "Running locally"
- Replace the custom download sliding menu with a Bootstrap dropdown
- Replace tipsy (hasn't been updated in forever) with tippyjs
- Align nav menu's left/right sides with the rest of the page
- Replace the Realtime/Gametime switch with a simpler button
- Replace the gold bloop next to Patrons' names with a gold background around the name (not used everywhere yet)
- Use Dashboard's `headerSortUp`/`headerSortDown` classes for `th` styling
- Add `format_ms_casual`, a helper like `format_milliseconds` but outputs "1m 3s" rather than "00:01:03"
- Fix a bug where non-URL video URLs like "beep" would 500 on save
- Add breadcrumbs to many pages
- Add a new "Splits I/O" root breadcrumb that goes to `/`
- List game aliases on the game page ("also known as: ...")
- Add some game stats to the game page
- Remove dual realtime/gametime columns on the splits table (just show based on current timing)
- Add some box shadows to various tables
- Remove "best on site" label for runs, as this was often misassigned to blank runs
- Change run page sub-nav (stats, edit, etc.) styling to buttons
- Clarify some button actions on the run edit page
- Add font-awesome glyphs to lots of buttons
- Fix a bug where search pagination links didn't work
- Fix a bug where creating game aliases would 500 (superuser)

TODOs:
- [x] Drag-and-drop uploading is broken (must fix before merging)
- [x] Use the new username badges in other places there are usernames
- [ ] Squishing the browser horizontally sometimes has weird layouts
- [ ] In general replace naked links (that blue color that doesn't seem like it belongs) with buttons / badges / navs / tabs / prettier ways to have links


Makefile
```diff
@@ -0,0 +1,32 @@
...
+
+ifeq ($(OS),Windows_NT)
+	detectedOS := Windows
+else
+	detectedOS := $(shell sh -c 'uname -s 2>/dev/null || echo not')
+endif
+
+docker-compose := docker-compose
+ifeq ($(detectedOS),Linux)
+	docker-compose := sudo docker-compose
```
BatedUrGonnaDie: This doesn't seem to actually work in my VM.  I confirmed that `uname -s 2>/dev/null || echo not` does indeed return `Linux`, but it doesn't preface any docker commands with sudo.

BatedUrGonnaDie:
A `make clean` command would be pretty cool as well.  This could probably be as simple and running `$(docker-compose) down`.

Makefile
```diff
@@ -0,0 +1,32 @@
...
+endif
+
+build:
+	$(docker-compose) build
+
+rubocop:
```
BatedUrGonnaDie: Getting this error with the rubocop make command.
```
Makefile:20: recipe for target 'rubocop' failed
make: *** [rubocop] Error 123
```

app/views/layouts/application.slim
```diff
@@ -9,7 +9,10 @@ html
     = render partial: 'shared/rollbarjs'
     = javascript_pack_tag 'application'
     = javascript_include_tag 'application'
+    = javascript_include_tag '/assets/toolkit.min.js'
```
BatedUrGonnaDie: Getting a 404 routing error when attempting to load this file.  From what I can tell this isn't checked into the repo with this commit.

app/views/layouts/application.slim
```diff
@@ -9,7 +9,10 @@ html
...
+    = javascript_include_tag 'https://unpkg.com/tippy.js/dist/tippy.all.min.js'
     = stylesheet_link_tag :application, media: :all
+    = stylesheet_link_tag '/assets/toolkit-inverse.min.css'
```
BatedUrGonnaDie: Same as the non-inverse min.css file above.

app/views/layouts/application.slim
```diff
@@ -9,7 +9,10 @@ html
     = render partial: 'shared/rollbarjs'
     = javascript_pack_tag 'application'
     = javascript_include_tag 'application'
+    = javascript_include_tag '/assets/toolkit.min.js'
+    = javascript_include_tag 'https://unpkg.com/tippy.js/dist/tippy.all.min.js'
```
BatedUrGonnaDie: Is there a reason as to not have this be managed by yarn and to load it from the URL specifically?

config/routes.rb
```diff
@@ -34,11 +34,10 @@
   get  '/cant-parse', to: 'runs#cant_parse', as: :cant_parse
   get  '/random',     to: 'runs#random',     as: :random
 
-  get  '/convert',    to: 'converts#new'
+  get  '/convert', to: 'converts#new'
 
-  get '/search',        to: 'search#index'
-  get '/search(?q=:q)', to: 'search#index'
```
BatedUrGonnaDie: I'm not super great with routing, is this route not needed when the base route without the query is defined?

Makefile
glacials: lol, turns out Makefiles need spaces for indenting up here, and tabs for recipes down below. Tabs up here don't work. Should be fixed now.

Makefile
glacials: Hm, couldn't reproduce this one. Do you have `xargs`?

app/views/layouts/application.slim
glacials: Should be fixed, moved this to be included in webpack rather than in `public/` (and not committed)

app/views/layouts/application.slim
glacials: Didn't know how, think I got it now though. Should be fixed

app/views/runs/new.erb
```diff
@@ -1,5 +1,9 @@
 <% title "upload" %>
```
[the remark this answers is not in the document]
glacials: The meta tag below is the reason, when I tried to replicate it in slim LiveSplit would behave like it wasn't there. I'm guessing it was a spacing issue, but I didn't dig deeper and just switched to erb where I knew it worked.

config/routes.rb
glacials: Nope it is not. A request to `GET /search?q=mario` will still match the `get '/search'` route and `params[:q]` will be set to `"mario"`.

BatedUrGonnaDie: Good to know, thanks for the info.

Makefile
BatedUrGonnaDie: Confirming that this is working correctly now :+1:

Makefile
BatedUrGonnaDie: Yep, the command runs just fine when I paste it into the terminal.  Looking at the make version and the regular command they output the same stuff, maybe rubocop outputs a non-zero exit code when it finds offenses and that causes the error for me.  Shouldn't matter though since it has the same output.

README.md
```diff
@@ -53,21 +55,29 @@ out everything (even if the only thing preventing you from setting up is better
...
+To run tests from inside the Docker container, use
 ```sh
-docker-compose run -e RAILS_ENV=test web rspec
+make test
 ```
 
+### Running Rubocop
+We use [Rubocop][rubocop] for code cleanliness and styling. To run it against changed files, commit your changes and run
+```sh
+make rubocop
```
BatedUrGonnaDie: This should be `make lint` now since the recipe changed.

app/javascript/packs/application.js
```diff
@@ -16,7 +16,6 @@ import "underscore"
 import {Spinner} from "spin.js"
 import "clipboard"
 import * as Cookies from "js-cookie"
-import "tipsy"
```
BatedUrGonnaDie: This should probably import tippy.  Currently I'm getting a `Uncaught ReferenceError: tippy is not defined`.  This might also fix the drag and drop upload error.

app/models/concerns/twitch_user.rb
```diff
@@ -4,50 +4,41 @@
...
-        success = true
-      end
-      update(twitch_user_follows_checked_at: nil) unless success
-    end
-
-    def twitch_sync!
-      body = Twitch::User.get(twitch_id)
+  def sync_twitch_follows!
+    ActiveRecord::Base.transaction do
+      TwitchUserFollow.where(from_user: self, to_user: (twitch_followed_users - current_followed_users)).delete_all!
```
BatedUrGonnaDie: Should be `delete_all`, otherwise this throws an error.

app/validators/run_validator.rb
```diff
@@ -7,21 +7,25 @@ def validate(record)
...
-      unless uri.host.match?(/^(www\.)?(twitch\.tv|youtube\.com|youtu\.be)$/)
-        record.errors[:base] << 'Your video URL must be a link to a Twitch or YouTube video.'
-      end
+    record.video_url.strip!
+    if record.video_url.blank?
+      record.video_url = nil
+      return
     end
+
+    record.errors[:base] << 'Your video URL must be a link to a Twitch or YouTube video.' unless valid_domain?(url)
```
BatedUrGonnaDie: This should probably be `record.video_url` instead of just `url`.

app/javascript/packs/application.js
glacials: Yep it does, nice catch!

app/views/games/categories/show.slim
```diff
@@ -1,20 +1,54 @@
 - title @game.name
 - content_for(:header) do
-  ol.breadcrumb
+  ol.breadcrumb.shadow
+    li.breadcrumb-item = link_to "Splits I/O", root_path
```
BatedUrGonnaDie: A small thing with all breadcrumbs, there is a larger gap on the right side of each breadcrumb when there is more than one.  This is caused by a left padding on each subsequent breadcrumb after the first one.  It's uniform on the live site and not sure if you want this to stay the same.

app/views/games/categories/show.slim
```diff
@@ -1,20 +1,54 @@
...
+            i.fas.fa-edit>
+            span Edit
 
 article data-turbolinks-temporary=true
-  = render partial: 'shared/category_tabs', locals: {game: @game, current_category: @category, link_type: :normal}
-  = render partial: 'shared/run_table', locals: table_locals(:category, category: @category)
+  .row
+    .col-sm-3.mb-3
+      .statcard.p-3
+        h3.statcard-number = @game.users.count
```
BatedUrGonnaDie: This isn't a concern right now, but if this page becomes a performance problem, it might be worth looking into cache counters for the model.  Essentially this tells rails to track how many of a relation exist all the time and store it in the db.

app/views/rivalries/index.slim
```diff
@@ -1,5 +1,8 @@
 - title 'Rivalries'
```
BatedUrGonnaDie: Same MS issue as above.

chrome_2018-06-06_19-04-35

app/views/runs/_split_table.slim
```diff
@@ -1,62 +1,33 @@
...
-              - else
-                span = format_milliseconds(segment.duration_ms(timing))
-            - if segment.skipped?(timing)
-              td.align-right.time -
-            - else
-              td.align-right = format_milliseconds(segment.end_ms(timing))
+            td.align-right.align-middle = format_ms(segment.end_ms(timing))
           td.align-right
-            a.segment-graph-toggler href='#' data-segment=segment.id data-collapsed='1' View History
+            a.segment-graph-toggler.btn.btn-secondary href='#' data-segment=segment.id data-collapsed='1'
```
BatedUrGonnaDie: This shows up for all runs, even those that don't have a history.  This should probably stay hidden if they don't have a history to show.

app/views/runs/_timing_badge.slim
```diff
@@ -0,0 +1,16 @@
+- if run.completed?(Run::REAL) && run.completed?(Run::GAME)
```
BatedUrGonnaDie: It's a bit hard to find the timing switch if you don't know where to look for it.  Not really sure a great fix, but Stef also agreed that it's not as obviously as to where it is.

app/views/runs/_title.slim
```diff
@@ -1,50 +1,68 @@
...
+    button#download-link.btn.btn-outline-success.dropdown-toggle href='#' role='button' data-toggle='dropdown' aria-haspopup='true' aria-expanded='false'
+      i.fas.fa-download>
+      span Download
+    .dropdown-menu.bg-dark aria-labelledby='download-link'
+      - timer = Run.program(run.timer)
+      a.dropdown-item.text-secondary href=download_path(run, timer.to_sym) = timer
+      a.dropdown-item.text-secondary href="https://one.livesplit.org/#/splits-io/#{run.id36}" target='_blank' LiveSplit One
+      - (Run.exportable_programs - [Run.program(run.timer)]).each do |timer|
+        a.dropdown-item.text-secondary href=download_path(run, timer.to_sym) = timer
+    - if run.category
```
BatedUrGonnaDie: This will show up if its the only run on the site, which then gives you a link to compare it to itself.  I feel like this should just not be there in this case.

This also shows up if it's the best run on site, leading to the same behavior.

app/views/runs/_split_table.slim
```diff
@@ -1,62 +1,33 @@
...
+            - if segment.skipped?(timing) || segment.reduced?(timing)
+              span.time -
             - else
-              td.align-right = format_milliseconds(segment.end_ms(Run::REAL))
-            - if segment.skipped?(Run::GAME)
-              td.align-right.time -
-            - else
-              td.align-right = format_milliseconds(segment.end_ms(Run::GAME))
+              span = format_ms(segment.duration_ms(timing))
+          - if segment.skipped?(timing)
```
BatedUrGonnaDie: If the split is skipped, then the finished at column will just be blank instead of having the `-` through the column like on the live site.

app/views/runs/_quick_stats.slim
```diff
@@ -0,0 +1,53 @@
...
+            small.delta-indicator.delta-positive.tip-top title='Since last PB'
+              = run.attempts - previous_upload.attempts
+        span.statcard-desc = 'Attempt'.pluralize(run.attempts)
+    .col-sm-3
+      .statcard.p-3
+        h3.statcard-number
+          span = format_ms(run.total_playtime_ms(timing))
+          - if previous_upload.present?
+            small.delta-indicator.delta-positive.text-success.tip-top title='Since last PB'
+              = format_ms_casual(run.total_playtime_ms(timing) - previous_upload.total_playtime_ms(timing))
```
BatedUrGonnaDie: I think the formatting is a bit messed up for this.

chrome_2018-06-06_19-22-11

app/views/runs/_quick_stats.slim
```diff
@@ -0,0 +1,53 @@
+- previous_upload = @run.user.try(:previous_upload_for, @run.category)
```
BatedUrGonnaDie: Looks like the delta arrows appear when the run you're looking at is the second best you have on record.

chrome_2018-06-06_19-26-09

app/views/runs/stats/index.slim
```diff
@@ -54,7 +56,7 @@ article#stats
             td
               span Total potential timesave
               small.tip-html< title="<code>PB - sum of best</code>" = "(?)"
```
BatedUrGonnaDie: This tip seems to be broken now.

app/views/search/index.slim
```diff
@@ -33,4 +34,7 @@ article#search-results
 article#runs
   - if @results[:games].present?
     h4 Runs
-    = render partial: 'shared/run_table', locals: table_locals(:games, games: @results[:games])
+    .card.shadow = render partial: 'shared/run_table', locals: { \
+      runs: Run.by_game(@results[:games]).nonempty.unarchived, \
+      cols: %i[runner timer name uploaded] \
```
BatedUrGonnaDie: `timer` should probably be `time`, unless you are explicitly trying to get the timer to show.  If it is the case, timer is not showing, and time would be a nice addition to this output so you know what kind of run you are about to look at.

app/views/shared/_run_table.slim
```diff
@@ -5,38 +5,34 @@
   center
     i No runs matched!
 - else
-  .card
-    - if description.present?
-      .card-header.text-white = description
+  - if description.present?
+    .card-header.text-white = description
+  .table-responsive
```
BatedUrGonnaDie: Is there a reason to wrap everything below in the table-responsive instead of adding it to the table html element below?

app/views/users/show.slim
```diff
@@ -12,4 +15,8 @@
     - @user.games.named.each do |game|
       => link_to game.shortname || game.name.downcase, game, class: 'game-link stealth-link'
 article
-  = render partial: 'shared/run_table', locals: table_locals(:pbs, user: @user)
+  = render partial: 'shared/run_table', locals: { \
```
BatedUrGonnaDie: This should be wrapped in a card like the rest of the tables on the site.

app/views/shared/_run_table.slim
```diff
@@ -65,5 +61,5 @@
                     = link_to '+', new_rivalry_path(category: run.category)
             - if cols.include?(:uploaded)
               td.align-right.nowrap class=('short' if run.short?) title=run.created_at = "#{time_ago_in_words(run.created_at)} ago"
-    .mx-auto
-      = paginate(runs, theme: 'twitter-bootstrap-4', pagination_class: 'mt-3')
+  .mx-auto
+    = paginate(runs, theme: 'twitter-bootstrap-4', pagination_class: 'mt-3')
```
BatedUrGonnaDie: Not sure if this is fixable, but the spacer that is put in the pagination between the last number and then `Next` is not styled liked the rest of the paginator.


screen shot 2018-06-07 at 2 29 31 am

app/views/shared/_follows.slim
```diff
@@ -22,7 +22,7 @@
                     .gold-split title='Patron'
                 div = "#{run.game} #{run.category}"
                 div
-                  = format_milliseconds(run.duration_ms(Run::REAL))
+                  = format_ms(run.duration_ms(Run::REAL))
```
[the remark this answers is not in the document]
BatedUrGonnaDie: I should mention that I don't really see this problem when it's showing green numbers.  This seems to apply to all the places that I mentioned it being super precise.

app/views/runs/_quick_stats.slim
```diff
@@ -0,0 +1,53 @@
...
+        h3.statcard-number = Run.program(run.timer)
+        span.statcard-desc = link_to('Timer', Run.program(run.timer).website, target: '_blank', class: 'text-muted')
+  .row
+    .col-sm-3
+      .statcard.p-3
+        h3.statcard-number
+          span> = run.attempts
+          - if previous_upload.present?
+            small.delta-indicator.delta-positive.tip-top title='Since last PB'
+              = run.attempts - previous_upload.attempts
```
BatedUrGonnaDie: This number can be negative but still show a green arrow.

app/views/shared/_run_table.slim
glacials: Looks intentional by Dashboard, it's not ideal but similar to before I think it's less valuable to change it than the amount of time we'd pay later by trying to restyle the theme.

app/views/runs/_quick_stats.slim
glacials: Fixed

app/models/concerns/twitch_user.rb
```diff
@@ -4,50 +4,41 @@
...
-      update(
-        twitch_id: twitch_id,
-        name: name.downcase,
-        avatar: uri.to_s,
-        twitch_display_name: display_name
-      )
-    rescue RestClient::ResourceNotFound
-      nil
+  def current_followed_users
+    Rails.cache.fetch("#{id}/follows") do
```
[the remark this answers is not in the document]
BatedUrGonnaDie: If we don't pass the ABC score in a lot of places, it might be worth just doing inline ignores on specific methods that way it still keeps other code to a higher standard.

I would agree that caching this is not great since AR caches queries already, and depending on how the cache is set up in production this might be slower than what AR does already.

app/views/shared/_run_table.slim
BatedUrGonnaDie: Good to know.  For some reason I thought in the past it was just attached to a table element.  Maybe in some last version it was like that or I'm just crazy.

app/views/shared/_follows.slim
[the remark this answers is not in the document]
BatedUrGonnaDie: I think it would depend on the length of the run to be honest.  Maybe if the run is less than an hour then show the millisecond value.

Gemfile.lock
```diff
@@ -76,16 +76,14 @@ GEM
       momentjs-rails (~> 2.8)
       sass-rails (~> 5.0)
       selectize-rails (~> 0.6)
-    api-pagination (4.7.1)
+    api-pagination (4.7.0)
```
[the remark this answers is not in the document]
BatedUrGonnaDie: Running `docker-compose run web bundle update` on the current commit of this branch results in the Gemfile.lock being nearly identical to the one on master.  There are a few gems that have received minor updates since the last round of updates.

README.md
```diff
@@ -76,7 +86,7 @@ to rebuild the Docker image for your changes to apply.
 Splits I/O uses [livesplit-core][livesplit-core] for parsing. The parser is located in `lib/parser/*`. To upgrade it,
 run
 ```sh
-docker-compose run web bundle exec rake update_lsc
+make update_lsc
 ```
 and commit the changes.
```
BatedUrGonnaDie: Can we get a small blurb about the `make clean` recipe, and maybe include that it will destroy the db and local s3?

app/controllers/patreon_users_controller.rb
```diff
@@ -1,65 +1,87 @@
...
 
-  def redirect_uri
-    "#{request.protocol}#{request.host_with_port}/auth/patreon/callback"
+    user_response['included'].find do |obj|
+      obj['type'] == 'pledge' && obj['relationships']['creator']['data']['id'] == creator_id
+    end['attributes']['created_at']
+  end
+
+  def link_success_msg
+    'You\'re now linked with Patreon! If you\'re a patron, your benefits have been activated :)'
```
BatedUrGonnaDie: Instead of using `\'` for the quotes, it should probably just be wrapped with regular quotes.  Rubocop won't throw a cop warning if there is a apostrophe in the string in the quotes.

app/javascript/packs/application.js
```diff
@@ -25,11 +23,8 @@ global.Spinner = Spinner
 global.Clipboard = clipboard
 global.Cookies = Cookies
```
BatedUrGonnaDie: Not any current line related, but I see that you put the toolkit.js in the webpack managed stuff, but it doesn't seem to be imported anywhere, which means it probably won't ever be ran.  Not sure if they do anything special on top of what bootstrap is doing.

app/models/concerns/twitch_user.rb
```diff
@@ -4,50 +4,37 @@
...
-    end
-
-    def twitch_sync!
-      body = Twitch::User.get(twitch_id)
+  def sync_twitch_follows!
+    ActiveRecord::Base.transaction do
+      current_followed_users = User.where(twitch_id: Twitch::Follows.followed_ids(twitch_id))
 
-      basic_info = JSON.parse(body)
+      TwitchUserFollow.where(from_user: self, to_user: (twitch_followed_users - current_followed_users)).delete_all
```
BatedUrGonnaDie: Not a problem, just a side note that `delete_all` doesn't run any callbacks associated with the model, if we ever end up adding any to this particular model.  Same as the note about `delete_all` that I put on the segment history and run history models.

app/views/games/categories/leaderboards/sum_of_bests/index.slim
```diff
@@ -3,36 +3,38 @@
...
+                small
+                  | Potential save
+                  a.tip< title='(PB - sum of best)'
+          tbody
+            - i = 0
+            - runners.each do |runner|
+              - pb = runner.pb_for(@category)
+              tr id=[:run, pb.id]
+                td.align-right = i += 1
+                td.align-left = link_to runner, runner, class: 'user-link'
```
BatedUrGonnaDie: This should probably be a badge as well, it is on the regular category show page.

app/views/games/categories/show.slim
```diff
@@ -1,20 +1,54 @@
...
+            = link_to('last run', @game.runs.order(created_at: :desc).first, class: 'text-muted')
+  .card
+    .card-header
+      = render partial: 'shared/category_tabs', locals: {game: @game, current_category: @category, link_type: :normal}
+    = render partial: 'shared/run_table', locals: { \
+      runs: @category.runs.nonempty.unarchived, \
+      cols: %i[runner time name uploaded], \
+    }.merge(sorting_info)
   center
     small = link_to 'sum of best leaderboard', game_category_sum_of_bests_path(@game, @category), class: 'leaderboard-link'
```
BatedUrGonnaDie: On the SoB leaderboard, this is part of the card, whereas on this page it is outside of it.  One should probably change to maintain consistency.

app/views/games/categories/show.slim
```diff
@@ -1,20 +1,54 @@
...
+  h1 = @category.game
+  - if @game.aliases.where.not(name: @game.name).present?
+    h6 Also known as:
+    ul
+      - @game.aliases.where.not(name: @game.name).each do |game_alias|
+        li = game_alias
   h6
     ul.run-options
+      - if @game.srl_uri.present?
+        li = link_to image_tag(asset_path("srl.png")), @game.srl_uri
```
BatedUrGonnaDie: With all other buttons having outlines, it might be nice to have this outlined as well to maintain a consistent look, but it's completely subjective on what you like best.

app/views/rivalries/new.slim
```diff
@@ -1,6 +1,10 @@
 - title 'Set a rival'
```
BatedUrGonnaDie: This page has some formatting issues.  Most notably is no padding on the avatar, and the same follow issue down below.

chrome_2018-06-08_19-37-36

app/views/runs/_split_table.slim
```diff
@@ -1,62 +1,36 @@
...
-              td.align-right.time -
-            - else
-              td.align-right = format_milliseconds(segment.end_ms(Run::REAL))
-            - if segment.skipped?(Run::GAME)
-              td.align-right.time -
-            - else
-              td.align-right = format_milliseconds(segment.end_ms(Run::GAME))
-          - else
-            td.align-right
+          - if run.segments.any? { |segment| segment.histories.any? }
```
BatedUrGonnaDie: If no splits have a history, this becomes N queries.  Using `includes(:histories)` solves this, while also caching the result for the iteration just below.

app/views/runs/_timeline.slim
```diff
@@ -4,8 +4,8 @@
     .alert.alert-warning This run doesn't have any game time segments.
 - else
   .timeline-background.mt-3
-    .timeline
+    .timeline.shadow
       - run.collapsed_segments(timing).each.with_index do |segment, index|
         .pure-u.split id="#{run.id36}-split-#{index}" class=next_timeline_color(run.id36) style="width: #{segment.duration_ms(timing).to_f / scale_to * 100}%; z-index: #{index}"
           .title = segment.name.presence || '-'
-          .time = format_milliseconds(segment.duration_ms(timing))
+          .time = format_ms(segment.duration_ms(timing))
```
BatedUrGonnaDie: There should probably be a newline at the end of this file.

app/views/runs/_title.slim
```diff
@@ -1,50 +1,67 @@
...
+ul.run-options
+  li
+    button#download-link.btn.btn-outline-success.dropdown-toggle href='#' role='button' data-toggle='dropdown' aria-haspopup='true' aria-expanded='false'
+      i.fas.fa-download>
+      span Download
+    .dropdown-menu.bg-dark aria-labelledby='download-link'
+      - timer = Run.program(run.timer)
+      a.dropdown-item.text-secondary href=download_path(run, timer.to_sym) = timer
+      a.dropdown-item.text-secondary href="https://one.livesplit.org/#/splits-io/#{run.id36}" target='_blank' LiveSplit One
+      - (Run.exportable_programs - [Run.program(run.timer)]).each do |timer|
```
BatedUrGonnaDie: This was probably me in the past, but here we call `Run.program(run.timer)` exactly again as we did above, so this could probably just be `[timer]` now.

app/views/runs/_title.slim
```diff
@@ -1,50 +1,67 @@
...
+        span Video
+  - if can?(:edit, run)
+    li
+      a.btn.btn-outline-success href=edit_run_path(run)
+        i.fas.fa-edit>
+        span Edit
+  - if run.user.nil?
+    li#claim-nav-link-container style='display: none;'
+      - if current_user.present?
+        = link_to "claim", '#', id: 'claim-nav-link', data: {confirm: "Claim this run as #{current_user.twitch_display_name}?"}
```
BatedUrGonnaDie: This confirm should probably use `current_user.to_s` in the event somehow that a display_name isn't present.

app/views/runs/compare.slim
```diff
@@ -10,14 +10,9 @@
   h5
```
BatedUrGonnaDie: The `short_run` variable at the top doesn't seem to be used anymore, so it can be deleted.

app/javascript/highchart_theme.js
```diff
@@ -11,7 +11,7 @@ Highcharts.theme = {
     "rgba(243, 123, 29, 1)"
   ],
   chart: {
-    backgroundColor: '#32383e',
+    backgroundColor: '#262830',
```
BatedUrGonnaDie: The background for the graphs isn't the same as the card background anymore.  Not sure if this was intended or not.

chrome_2018-06-08_19-53-58

app/views/runs/stats/index.slim
```diff
@@ -19,101 +21,102 @@ article#stats
...
+          - else
+            tr
+              td
+                span Sum of best
+              td = format_ms(@run.sum_of_best_ms(timing))
+          - if !has_sob || !has_attempts
+            tr.tip title="This run has no gold splits recorded"
+              td
+                span Total potential timesave
+                small.tip-right< title="<code>PB - sum of best</code>" = "(?)"
```
BatedUrGonnaDie: Since all the rest of the tips are FA icons, it might be worth considering turning this into one as well to make it consistent, but up to you on that one.

app/views/search/index.slim
```diff
@@ -16,21 +17,24 @@ article#search-results
...
-                  - case search_type
-                  - when :games
-                    = link_to obj, game_path(obj), class: 'game-link'
-                    small.float-right = obj.categories.pluck(:name).join(', ')
-                  - when :users
-                    = link_to obj, user_path(obj), class: 'game-link'
-                    small.float-right = obj.games.pluck(:shortname).compact.uniq.join(', ')
-
+      .card.mb-3.shadow
+        .table-responsive
```
BatedUrGonnaDie: Is there a reason to not have this in a `.card-body` and to leave it on the top level card?

app/views/search/index.slim
```diff
@@ -16,21 +17,24 @@ article#search-results
...
+                      = link_to obj, game_path(obj), class: 'game-link'
+                      small.float-right = obj.categories.pluck(:name).join(', ')
+                    - when :users
+                      = link_to obj, user_path(obj), class: 'game-link'
+                      small.float-right = obj.games.pluck(:shortname).compact.uniq.join(', ')
 article#runs
   - if @results[:games].present?
     h4 Runs
-    = render partial: 'shared/run_table', locals: table_locals(:games, games: @results[:games])
+    .card.shadow = render partial: 'shared/run_table', locals: { \
```
BatedUrGonnaDie: Same as above.

Gemfile.lock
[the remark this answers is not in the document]
glacials: Huh. I can replicate that too. I have no idea how I got into this situation then, I never explicitly downgraded any gems.

Whatever, I'll commit the new lockfile and leave this to the bundle gods.

app/controllers/patreon_users_controller.rb
glacials: Ohh cool I didn't know that about Rubocop!

Gemfile.lock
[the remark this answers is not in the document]
BatedUrGonnaDie: :pray: Bundle Gods :pray:

app/javascript/packs/application.js
glacials: Nice thanks, probably nothing vital but I'm sure this callout saved a future headache

app/views/games/categories/show.slim
glacials: Hmm, good point but in trying to outline it the green outline actually clashes pretty bad with the SRL logo colors. I'll use a `.btn.btn-dark` to hit halfway between consistent and decent-looking.

app/views/rivalries/new.slim
glacials: Good catch, redid it all to look decent. Probably still want to revisit at some point but it's back to looking good enough for now.

app/javascript/highchart_theme.js
glacials: I think I must have fixed this in a recent commit -- this is what I see:

screenshot 2018-06-09 16 29 01

glacials: Oh woops, it's not fixed on the stats page, only the individual segment history graphs. Will fix.

BatedUrGonnaDie: Looks like the backgrounds are different colors, might be from the striped table?

app/views/search/index.slim
glacials: Hmm, it seems like in the [`.card` docs][1] they generally use `.card-body` only when they're putting naked text right into the card. For the `.card > .list-group`, `.card > .card-text`, and `.card > .card-img` use cases they put those elements right inside the `.card` with no `.card-body`.

There's no example for tables, but it seems to match that convention of not being naked text, so I'd argue it doesn't belong inside a `.card-body`.

Practically, using a `.card-body` gives this:
screenshot 2018-06-09 16 55 12
whereas not using one gives:
screenshot 2018-06-09 16 55 15

and I like the second one more. You could also accomplish this with a `.card-body.p-0`, but at that point I'm not sure what the difference is between that and not using one, so I'd defer back to my original point above.

[1]: https://getbootstrap.com/docs/4.1/components/card/

BatedUrGonnaDie: Ah, when I went through the docs it made it seem like you should put stuff in the body portion, but I was probably misinterpretting.  I agree it looks better without it.  Thanks for the info!

app/views/layouts/application.slim
```diff
@@ -26,58 +26,59 @@ html
...
+                    = link_to 'Settings', settings_path, class: 'dropdown-item text-secondary'
+                    = link_to 'Tools', tools_path, class: 'dropdown-item text-secondary'
+                    = link_to 'Sign Out', session_path(auth_session), method: :delete, class: 'dropdown-item text-secondary'
+            - else
+              li.nav-item
+                a.nav-link.px-3 href='/auth/twitch' Sign Up
+              li.nav-item
+                a.nav-link.px-3 href='/auth/twitch' Sign In
       = render partial: 'shared/ad'
       header#header = yield(:header)
```
BatedUrGonnaDie: Alerts seem to be in a bit of a weird spot now.

firefox_2018-06-09_23-37-17

app/views/runs/_title.slim
```diff
@@ -1,50 +1,67 @@
...
-          = link_to "sign in to claim", '/auth/twitch'
+          i.fas.fa-video>
+        span Video
+  - if can?(:edit, run)
+    li
+      a.btn.btn-outline-success href=edit_run_path(run)
+        i.fas.fa-edit>
+        span Edit
+  - if run.user.nil?
+    li#claim-nav-link-container style='display: none;'
```
BatedUrGonnaDie: The claim link is unstyled, but seems a bit redundant.  Could remove it if you wanted, but it shouldn't be left as is.

app/views/games/categories/leaderboards/sum_of_bests/index.slim
```diff
@@ -3,36 +3,38 @@
   h2 Fastest sum of best
```
BatedUrGonnaDie: Breadcrumbs seem to be missing on this page.

Also, should it have the srl logo button and edit button on this page or nah?

app/views/games/categories/leaderboards/sum_of_bests/index.slim
```diff
@@ -3,36 +3,38 @@
...
+            tr
+              th
+              th.align-left Runner
+              th.align-left Sum of best
+              th.align-left
+                small PB
+              th.align-left
+                small
+                  | Potential save
+                  a.tip< title='(PB - sum of best)'
```
BatedUrGonnaDie: This should probably be a FA icon as well if you want to keep the theme of using them like on the stats page.

app/views/rivalries/new.slim
```diff
@@ -1,23 +1,30 @@
...
+  ol.breadcrumb.shadow
+    li.breadcrumb-item = link_to "Splits I/O", root_path
+    li.breadcrumb-item = link_to 'Rivals', rivalries_path
+    li.breadcrumb-item = link_to 'Add', new_rivalry_path
+    - if @category.present?
+      li.breadcrumb-item = link_to(@category.game, game_path(@category.game))
+      li.breadcrumb-item = link_to(@category, game_category_path(@category.game, @category))
   - if @category.present?
+    h2 Choose a Rival
     h5
```
BatedUrGonnaDie: This should probably be removed since we have the breadcrumbs right above it.

app/views/runs/_title.slim
```diff
@@ -1,50 +1,67 @@
...
-      span by #{link_to run.user, user_path(run.user), class: 'stealth-link user-link'}
-      - if run.user.silver_patron?
-        .gold-split title='Patron'
-h6
+    span> by
+    span = user_badge(run.user)
+    .badge
+      i.fas.fa-check>
+      span Personal Best
+  = render partial: 'runs/timing_badge', locals: {run: run, timing: timing}
```
BatedUrGonnaDie: On some pages this is on its own line above everything else, others it's inline with the username and stuff.

app/views/runs/_title.slim
BatedUrGonnaDie: I should note that the giant claim box is right below the claim link, it's probably in another one of the screenshots.

app/views/runs/compare.slim
```diff
@@ -1,6 +1,5 @@
 - title @run
 - timing = params[:timing] || @run.default_timing
-- short_run = @run.short? || @comparison_run.short?
```
BatedUrGonnaDie: This page is also missing breadcrumbs.

app/views/runs/stats/_segments.slim
```diff
@@ -4,44 +4,50 @@
     .float-right
       - if run.attempts.nil? || run.attempts.zero?
         strike ↓ download segment history csv
-        small.tip-html< title="Segment history is not available for this run" = "(?)"
+        span.tip-top< title="Segment history is not available for this run"
+          = icon('fas', 'question-circle')
```
BatedUrGonnaDie: The rest of this PR uses `.fas.fa-question-circle`, is there a reason to change it for this page only?

app/views/settings/index.slim
```diff
@@ -86,62 +90,63 @@
...
-                          small = button_to application_path(application), class: 'btn btn-primary', method: :delete, \
-                            data: { \
-                              confirm: "Really delete #{application.name}? This will also revoke all \
-                                        #{application.access_tokens.count} authorizations and invalidate the associated \
-                                        tokens." \
-                            }
-                            span = icon("fas", "times")
+                          td
+                            - if (Time.now - 24.hours) > application.created_at
+                              i Hidden forever
```
BatedUrGonnaDie: This should probably be wrapped in a `btn.btn-primary` to maintain consistency with the client id.  Looks a bit out of place currently.

app/views/runs/_title.slim
glacials: I put both in on purpose when I made this feature: the giant box is an attention-getter, but if you don't want to claim you can dismiss the box with the X. If you do, the link in nav serves as a way to still claim it later.

But I agree even though each has their purpose that it feels redundant. I won't try to handle this now though since it's current behavior in production, but I'll fix the styling on the link to be a button like the others.

BatedUrGonnaDie: Ah, I didn't realize that dismissing the claim box would make it stay gone.  It makes perfect sense to keep it then.  Just needs to be styled on.

app/views/runs/_title.slim
glacials: In the first image there's no user or Personal Best label, so it's the same line but it's just otherwise empty. This seems correct.

BatedUrGonnaDie: Question then: should this page also say something like `by: ???` like it does on the game page or should it just be blank?

app/views/runs/stats/_segments.slim
glacials: Uhh because I just now discovered this helper lol

Is the helper the preferred way to do it?

app/views/runs/_title.slim
glacials: Yeah perhaps, I can change this.

app/views/settings/index.slim
glacials: The colors of a `.btn-primary` didn't look right for informational non-interactable text, but I went for a `.btn.btn-secondary.btn-block` with `disabled`:

screenshot 2018-06-10 22 36 27

app/views/runs/stats/_segments.slim
BatedUrGonnaDie: The helper is included with the FA gem.  No idea what the preferred way to do it is if I'm being honest, just seemed weird that nothing else used it and then it was being used here ¯\_(ツ)_/¯

app/views/runs/_title.slim
```diff
@@ -1,72 +1,64 @@
...
-      - if run.user.silver_patron?
-        .gold-split title='Patron'
-h6
+h1 = run.to_s
+h5
+  span> by
+  span = user_badge(run.user)
+  .badge
+    i.fas.fa-check>
+    span Personal Best
```
BatedUrGonnaDie: Every single run is now a personal best, including unowned runs.

app/views/settings/index.slim
```diff
@@ -121,7 +121,7 @@
                                   span#copied style='display: none;' = icon("fas", "check")
                           td
                             - if (Time.now - 24.hours) > application.created_at
-                              i Hidden forever
+                              button.btn.btn-secondary.btn-block disabled=true Hidden forever
```
BatedUrGonnaDie: I lied! 2 more things!  Their should probably not be any space between the disabled block and the X button.

app/views/runs/_title.slim
glacials: lol

app/views/settings/index.slim
glacials: The X button deletes the application so it belongs to the whole row, not to that column, so I think it's correct that they're separate.

screenshot 2018-06-11 10 06 51

Though I admit that's not obvious as a user. I will try to visually differentiate it a bit more.

BatedUrGonnaDie: Ah, that makes much more sense then.   I thought it would just reset the client secret so that you could get a new one if you llwt the old one.  Is there not a way to do that?

glacials: No, you have to make a new client. I'll make a ticket to allow resetting since most providers let you do that.

BatedUrGonnaDie:
I don't know how to do another like review thing but approved.

PR #398, opened 2018-06-04
Fix sql depreciations introduced in 5.2

BatedUrGonnaDie:
Misattributed to pg_search, when previously looked at it was checking the wrong line.

In docker-compose this fixes running the webpack server for my VM.  For some reason with `bundle exec` it throws cannot find ruby.  It effectively does the same thing since the binstub loads everything that it needs

### DIFF
--- a/app/models/game.rb
+++ b/app/models/game.rb
@@ -61,13 +61,13 @@ class Game < ApplicationRecord
   def popular_categories
     categories.joins(:runs).group('categories.id')
               .having('count(runs.id) >= ' + (Run.where(category: categories).count * 0.05).to_s)
-              .order('count(runs.id) desc')
+              .order(Arel.sql('count(runs.id) desc'))
   end
 
   def unpopular_categories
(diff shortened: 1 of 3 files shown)


app/models/game.rb
```diff
@@ -61,13 +61,13 @@ def to_s
   def popular_categories
     categories.joins(:runs).group('categories.id')
               .having('count(runs.id) >= ' + (Run.where(category: categories).count * 0.05).to_s)
-              .order('count(runs.id) desc')
+              .order(Arel.sql('count(runs.id) desc'))
```
glacials: Is this the new way to do all raw SQL?

BatedUrGonnaDie: What do you mean with all raw SQL?  I'm not really sure why Rails has an issue with this statement, but the depreciation notice said that if you can guarantee that there are no user provided values to wrap it in `Arel.sql` to solve to depreciation.

PR #404, opened 2018-06-14
Prevent search engines from scraping download urls

BatedUrGonnaDie:


glacials:
What's the reason for the change?

BatedUrGonnaDie:
Every couple of days (or maybe once a week) I notice that in NewRelic it says that the download action has up to 80rpm over the course of an hour or 2.  I'm assuming that it's being crawled since the show action also sees a large jump (around 1/5 the traffic) in the same time frame.  Also if it was a bot looking to download runs that can be achieved through the API so I thought it would be scraping.  I don't see any reason that the downloads should be scraped, but I could be missing something there.

BatedUrGonnaDie:
I'll try to remember to put the reasoning in the description box in the future.  Don't know why I didn't include it to begin with to be honest.

PR #405, opened 2018-06-14
Fix dropdowns eating first click

glacials:
Fixes a bug where dropdown menus eat your first click and do nothing, requiring a second click to actually drop down.

The Dashboard theme apparently bundles base Bootstrap CSS and JS inside itself, so this was caused by us including both base JS and the theme JS, and I don't know specifically what was happening but I'm guessing attaching the same handler twice to a click is generally bad. Removing the base Bootstrap JS fixes the issue.

I also removed some leftover code from bootstrap-toggle, which is the iOS-like realtime/gametime switch currently in production (but no longer in master).

With this fix, master will be in a good state to go to production.

### DIFF
--- a/config/webpack/environment.js
+++ b/config/webpack/environment.js
@@ -17,7 +17,6 @@ environment.plugins.prepend(
     clipboard: 'clipboard',
     tipsy: 'tipsy',
     "jquery.turbolink": "jquery.turbolinks",
-    "bootstrap-toggle": "bootstrap-toggle"
   })
 )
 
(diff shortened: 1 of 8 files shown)


config/webpack/environment.js
```diff
@@ -17,7 +17,6 @@ environment.plugins.prepend(
     clipboard: 'clipboard',
```
BatedUrGonnaDie: Can you also remove the c3 and d3 lines from this file?  The move to highcharts removed those 2 dependencies, but I forgot to remove them at that time.

BatedUrGonnaDie:
Looks good, no longer takes 2 clicks.

PR #416, opened 2018-06-18
Replace hamburger with dots icon

### LINKED ISSUE #414: [Suggestion] Overflow menu for PB table on home page should be on right side and icon should change
image

The position and style of the overflow menu icon (hamburger icon) gives off the feeling of a drag & drop handlebar.

Example:
image

It also just feels like a weird UX to have the Disown & Delete actions show up at the *beginning* of a line rather than the end of it (at least for LTR orientation). My suggestion is to replace the icon with the popular "3 dots" icon (shown below) and move it to the right-most end of the row. The hamburger icon is more often used to represent a larger sidebar menu rather than a small dropdown one, and the 3 dots icon has kind of taken its place for the latter purpose. A small downwards facing arrow would be another alternative. This change could theoretically even be implemented in SASS only, although an HTML structure change would probably be best anyways.

image

glacials:
Changes the run table hamburger icon for disowning / deleting to three vertical dots, and moves it to the right side of the table. Fixes #414.


glacials:
Hmm, I can't reproduce this. What browser / OS?

glacials:
Also, does this happen for you with the hamburgers on production right now, or just the dots?

BatedUrGonnaDie:
Windows 7 Chrome, only happens on dev.

jhobz:
If you don't mind pushing it to beta.splits.io, I could try taking a whack at debugging it. It sucks that Docker still doesn't work on Windows 10 Home.

glacials:
Sure, just pushed and it's on beta now. The normal deploy process for production takes over beta though, so let me know if it falls off and I'll put it back.

jhobz:
Ok so I don't know how to fix it, but I've at least learned a few more things about it:

1. It only happens on a table with no pagination or on the last page of a table with pagination. Could this be why you couldn't reproduce it perhaps Glacials?
2. ***Anything*** that causes a layout recalculation on the `div.dropdown.text-center` container seems to fix the issue, as far as I can tell. The reason it fixes on the second opening of the dropdown is likely because the order of the styles on child `div.dropdown-menu.bg-dark.show` change, which would force a layout recalc in the parent.

Given these factors, my assumption is that there's just some kind of layout bug in Chrome maybe? My suggestions for things to try would be:

1. Convert the inline styles on `div.dropdown-menu.bg-dark.show` to a class.
2. If not 1, force the order of styles on `div.dropdown-menu.bg-dark.show` to remain consistent. My guess is that this would make the bug *always* happen, rather than fix it, but at least it's info.
3. Try removing the `will-change: transform` rule. The [MDN page on this property](https://developer.mozilla.org/en-US/docs/Web/CSS/will-change) is pretty vocal about it not being very intuitive and even recommends against using it if you're not noticing performance issues.

Below are some screenshots of the relevant elements during a few different states, in case it helps.
After page load/refresh:
chrome_2018-06-19_19-30-13

After clicking overflow icon (introduces whitespace):
chrome_2018-06-19_19-30-36

After dismissing dropdown menu and then clicking overflow icon again (removes whitespace):
chrome_2018-06-19_19-30-46

After dismissing and opening dropdown for the 3rd time (no change from previous):
chrome_2018-06-19_19-30-55

glacials:
Nice investigation!

> It only happens on a table with no pagination or on the last page of a table with pagination. Could this be why you couldn't reproduce it perhaps Glacials?

Apparently not because I can't reproduce on macOS in a non-paginated table, and I can reproduce on Windows 10 in a paginated table on the first page, so I'm not sure what's going on.

The style attribute on the `.dropdown-menu` is controlled by stock [Bootstrap dropdowns][1], so we don't have control over it. This includes the fact that it's a style rather than a class, the order of the styles, and the `will-change` rule.

[1]: https://getbootstrap.com/docs/4.1/components/dropdowns/

What I've started looking at is **what changed**? This isn't broken on production. This branch only has four changes:

- The column was moved from the left edge of the table to the right edge
- The icon was changed from a hamburger to dots
- We added `.text-center` to `div.dropdown`
- We removed `.align-right` from `tr`

One of the above changes caused this breakage. Which one? Why?

By messing around in the inspector on Windows I discovered can stop the breakage per-row by moving the `td` back to the beginning of its row. I can also _cause_ the breakage on production by moving the `td` to the end!

animation of causing the breakage on production

So it's the position in the table! In some trials I saw a horizontal scroll bar appear at the bottom of the table briefly so I suspect the dropdown menu is appearing inside the table and clipping off the right side. This spawns a horizontal scrollbar for the table which increases the height of the table, but in the next frame the dropdown menu is rubber-banded back in so it hugs the right edge of the table.

With nothing more to scroll to, the horizontal scrollbar kills itself but the space where it lived doesn't get eliminated until the next render event, like you said. I tested this with other render events like shrinking the window size, and they kill it as well.

On top of all this, I suspect the dropdown menu "remembers" its position once it's rendered, which is why the issue never happens again for that specific menu.

One easy fix to this is to change our `.dropdown` to `.dropleft`, forcing the menu to appear to the left of the dots:

picture of dropdown appearing to the left of the dots

Another is to remove the wrapping `.table-responsive`, which is what causes the table to be horizontally scrollable independently of the page. By doing this, the dropdown menu can render outside of the table:

picture of the dropdown menu appearing on top of the table border, partially out and partially in

However, the ability to scroll the table horizontally is really useful for phones in portrait mode, and works really well with touch gestures. For this reason I think we should dropleft this guy and be done with it.

The next question is: is this a bug in something? Splits I/O? Bootstrap? Popper.js, the third party library Bootstrap uses for these menus? The internet? The universe? Probably one of the last two.

BatedUrGonnaDie:
Looks good to me with the `.dropdown-left` change.

PR #417, opened 2018-06-18
Fix follows alignment

### LINKED ISSUE #415: "Recent PBs by People You Follow" section on home page does not line up with title
image

All 3 of these elements should probably align together, but at the very least the title and box below should match each other.

glacials:
Prevents the follows table from shifting a bit to the right vs. other cards on this page by using `.p-0`. I also refactored the `.list-group` a bit to get the colors more consistent (wrapped it in a `.card`) and started using `.media` which is the supported way of putting a picture to the left of some content like this in Bootstrap. Fixes #415.


glacials:
before: 
screenshot 2018-06-17 23 04 55


after:
screenshot 2018-06-17 23 05 14

PR #422, opened 2018-06-22
Hook up "time since upload" to Action Cable

glacials:
Auto-updates the "time since upload" stat on the run page by publishing periodic updates over Action Cable. So after initial upload, "less than a minute" becomes "1 minute" becomes "2 minutes" etc. without refreshing the page.

I'm using Action Cable for this instead of doing it just locally because I'm looking to set up Action Cable infrastructure for other things and this is a trial run of putting something with Action Cable into production. I also don't want to duplicate the timestamp-to-words logic and I remember moving all time-parsing and timeline-building logic to the frontend and regretting it because I had to fight Rails over small things until I undid it all.

Also looking at getting this working in webpacker with [actioncable-modules](https://www.npmjs.com/package/actioncable-modules) but it didn't work right away so I'll have to look at it some more.


BatedUrGonnaDie:
The rails team maintains an official version of action cable on npm, not sure if they ever fixed the module thing the one you linked talked about though.  Might be worth looking into.

app/assets/javascripts/cable/subscriptions/run.js
```diff
@@ -0,0 +1,7 @@
+$(function() {
+  if (gon.run) {
```
BatedUrGonnaDie: Be careful of multiple subscriptions on multiple page loads since turbolinks will keep the connection active across page loads.

app/channels/run_channel.rb
```diff
@@ -0,0 +1,16 @@
+class RunChannel < ApplicationCable::Channel
+  def subscribed
+    run = Run.find36(params[:run_id])
+    stream_for(run)
+  end
+
+  class << self
```
BatedUrGonnaDie: Not sure about class methods, but all public methods are callable from the JS side of things, which means that anyone can all them through the dev console and pass any data that they want to the functions defined here.

glacials: Tested this and it does not seem that I can call class methods.

PR #431, opened 2018-08-10
Allow editing applications & resetting secrets

### LINKED ISSUE #430: No way to edit existing applications
This means if you make your redirect_uri `debug`, you can never change it and would have to remake the app and hope you never need to change it.

glacials:
Adds the ability to edit existing developer applications (changing name, changing redirect URI, resetting secrets). Fixes #430.

### DIFF
--- a/app/controllers/applications_controller.rb
+++ b/app/controllers/applications_controller.rb
@@ -1,5 +1,6 @@
 class ApplicationsController < ApplicationController
-  before_action :set_application, only: [:destroy]
+  before_action :set_application, only: [:edit, :update, :destroy]
+
   def new
   end
 
@@ -11,6 +12,40 @@ class ApplicationsController < ApplicationController
     else
       error_text = @application.errors.full_messages.to_sentence
       redirect_to new_application_path, alert: ":( Failed to create application: #{error_text.chomp('.')}."
+    end
+  end
+
+  def edit
+    if cannot?(:edit, @application)
+      render :forbidden, status: 403
+      return
+    end
+  end
+
+  def update
+    if cannot?(:edit, @application)
+      render :forbidden, status: 403
+      return
+    end
+
+    if params[:reset_secret] == '1'
+      if @application.update(
+        secret: Doorkeeper::OAuth::Helpers::UniqueToken.generate,
+        secret_generated_at: Time.now
+      )
+        redirect_to(application_path(@application), notice: "Secret reset! :O Don't lose it this time!")
+      else
+        error_text = @application.errors.full_messages.to_sentence
+        redirect_to application_path(@application), alert: ":( Failed to reset secret: #{error_text.chomp('.')}."
+      end
+      return
+    end
+
+    if @application.update(application_params[:doorkeeper_application])
+      redirect_to(application_path(@application), notice: "Application saved! :D")
+    else
+      error_text = @application.errors.full_messages.to_sentence
+      redirect_to application_path(@application), alert: ":( Failed to update application: #{error_text.chomp('.')}."
     end
   end
 
--- a/app/controllers/applications_controller.rb
+++ b/app/controllers/applications_controller.rb
@@ -1,57 +1,59 @@
 class ApplicationsController < ApplicationController
-  before_action :set_application, only: [:edit, :update, :destroy]
+  before_action :set_application, only: %i[edit update destroy]
 
   def new
   end
 
   def create
+    if cannot?(:create, Doorkeeper::Application)
+      render(:forbidden, status: :forbidden)
+      return
+    end
+
     @application = Doorkeeper::Application.new(application_params[:doorkeeper_application])
     @application.owner = current_user if Doorkeeper.configuration.confirm_application_owner?
     if @application.save
-      redirect_to settings_path, notice: 'Application created! :)'
+      redirect_to(settings_path, notice: 'Application created! :)')
     else
       error_text = @application.errors.full_messages.to_sentence
-      redirect_to new_application_path, alert: ":( Failed to create application: #{error_text.chomp('.')}."
+      redirect_to(new_application_path, alert: ":( Failed to create application: #{error_text.chomp('.')}.")
     end
   end
 
   def edit
-    if cannot?(:edit, @application)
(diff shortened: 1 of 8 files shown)


app/controllers/applications_controller.rb
```diff
@@ -14,6 +15,40 @@ def create
...
+  def update
+    if cannot?(:edit, @application)
+      render :forbidden, status: 403
+      return
+    end
+
+    if params[:reset_secret] == '1'
+      if @application.update(
+        secret: Doorkeeper::OAuth::Helpers::UniqueToken.generate,
+        secret_generated_at: Time.now
```
BatedUrGonnaDie: This should be `Time.now.utc` just so no timezone shennanigans can happen.

app/controllers/applications_controller.rb
```diff
@@ -14,6 +15,40 @@ def create
...
+        redirect_to(application_path(@application), notice: "Secret reset! :O Don't lose it this time!")
+      else
+        error_text = @application.errors.full_messages.to_sentence
+        redirect_to application_path(@application), alert: ":( Failed to reset secret: #{error_text.chomp('.')}."
+      end
+      return
+    end
+
+    if @application.update(application_params[:doorkeeper_application])
+      redirect_to(application_path(@application), notice: "Application saved! :D")
```
BatedUrGonnaDie: Single quotes rubocop warning.

app/views/settings/index.slim
```diff
@@ -1,8 +1,8 @@
-- title "settings"
+- title "Settings"
```
BatedUrGonnaDie: This should probably be single quotes as well since you changed a bunch of `"` later in the file.

app/views/settings/index.slim
```diff
@@ -115,20 +115,20 @@
...
-                              .input-group-btn
+                              .input-group-append
                                 .btn.btn-primary.clipboard-btn data-clipboard-target="#client-id-#{application.uid}"
-                                  span#copy = icon("fas", "copy")
-                                  span#copied style='display: none;' = icon("fas", "check")
+                                  span#copy = icon('fas', 'copy')
+                                  span#copied style='display: none;' = icon('fas', 'check')
                           td
-                            - if (Time.now - 24.hours) > application.created_at
+                            - if (Time.now - 24.hours) > application.secret_generated_at
```
BatedUrGonnaDie: This should probably be `Time.now.utc` as well.

app/views/applications/edit.slim
```diff
@@ -0,0 +1,75 @@
...
+              .col-lg-8
+                input.form-control type='text' value=@application.uid disabled=true
+              .col-lg-4
+                b Client ID
+                i< = "(not changeable)"
+          p
+            .row
+              .col-lg-8
+                .input-group
+                  - if (Time.now - 24.hours) > @application.secret_generated_at
```
BatedUrGonnaDie: `Time.now.utc` again.

app/views/applications/edit.slim
```diff
@@ -0,0 +1,75 @@
...
+                    input.form-control type='text' value="<Hidden forever>" disabled=true
+                  - else
+                    input.form-control.border-warning type='text' value=@application.secret disabled=true
+                  .input-group-append
+                    = button_to(application_path(@application), class: 'btn btn-warning', method: :patch, \
+                      params: {reset_secret: '1'}, \
+                      data: { \
+                        confirm: "Really reset the client secret for #{@application.name}? This will invalidate \
+                        the current secret, preventing it from being used. Existing user authorizations will not \
+                        be affected." \
```
BatedUrGonnaDie: Not problem with this PR, but there should probably be a way to revoke all tokens and auths if resetting the secret doesn't do that.  If you do leak your secret there should be a way to make sure someone else cannot be requesting token with your previously set up apps.

app/controllers/applications_controller.rb
```diff
@@ -14,6 +15,40 @@ def create
...
 
+  def edit
+    if cannot?(:edit, @application)
+      render :forbidden, status: 403
+      return
+    end
+  end
+
+  def update
+    if cannot?(:edit, @application)
```
BatedUrGonnaDie: Should this be the `:update` permission?  I see other abilities using `:update` but I'm not sure if it's different for those or not.

app/controllers/applications_controller.rb
```diff
@@ -14,6 +15,40 @@ def create
     end
   end
 
+  def edit
+    if cannot?(:edit, @application)
+      render :forbidden, status: 403
+      return
```
BatedUrGonnaDie: Redundant `return` rubocop warning.

db/migrate/20180810055852_add_secret_generated_at_to_applications.rb
```diff
@@ -0,0 +1,5 @@
+class AddSecretGeneratedAtToApplications < ActiveRecord::Migration[5.2]
+  def change
+    add_column :oauth_applications, :secret_generated_at, :datetime, default: -> { 'CURRENT_TIMESTAMP' }, null: false
```
BatedUrGonnaDie: This is going to have the default value be the current time when the migration is run, showing all old client secrets again for another 24 hours.  Just want to make sure you are aware of this.

app/models/ability.rb
```diff
@@ -3,17 +3,30 @@ class Ability
 
   def initialize(user)
     user ||= User.new
+    grant_anon_perms
+    grant_normal_perms(user)
+    grant_admin_perms(user) if user.id == 1
```
BatedUrGonnaDie: `grant_admin_perms` is defined without a parameter, this causes an error when you go to create an oauth application.

app/views/applications/edit.slim
BatedUrGonnaDie: Ah, if the functionality exists that's probably fine to be honest.  I just know that accidents happen and I've accidentally committed keys before, so having a way to wipe all access tokens is good.

app/models/ability.rb
glacials: oops!

BatedUrGonnaDie:
Looks good to me now.

PR #435, opened 2018-08-18
Fix Rubocop errors in controllers dir

davygora:
Related to: https://github.com/glacials/splits-io/issues/331
Fixed rubocop errors in `app/controllers/.`

### DIFF
--- a/app/controllers/admin/application_controller.rb
+++ b/app/controllers/admin/application_controller.rb
@@ -9,9 +9,7 @@ module Admin
     before_action :authenticate_admin
 
     def authenticate_admin
-      if current_user.nil? || current_user.id != 1
-        redirect_to root_path
-      end
+      redirect_to root_path if current_user.nil? || current_user.id != 1
     end
 
     # Override this value to specify the number of elements to display at a time
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -27,7 +27,7 @@ class ApplicationController < ActionController::Base
   end
 
   def after_sign_out_path_for(_resource_or_scope)
-    request.referrer
+    request.referer
   end
 
   def not_found
--- a/app/controllers/runs_controller.rb
+++ b/app/controllers/runs_controller.rb
@@ -3,10 +3,10 @@ require 'uri'
 require 'speedrundotcom'
 
 class RunsController < ApplicationController
-  before_action :set_run,        only: [:show, :download, :destroy, :compare, :edit, :update]
+  before_action :set_run,        only: %i[show download destroy compare edit update]
   before_action :set_comparison, only: [:compare]
 
-  before_action :first_parse, only: [:show, :edit, :update, :download], if: -> { @run.parsed_at.nil? }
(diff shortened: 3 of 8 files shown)


app/controllers/admin/application_controller.rb
```diff
@@ -9,9 +9,7 @@ class ApplicationController < Administrate::ApplicationController
     before_action :authenticate_admin
 
     def authenticate_admin
-      if current_user.nil? || current_user.id != 1
-        redirect_to root_path
-      end
+      redirect_to root_path if current_user.nil? || current_user.id != 1
```
glacials: I've been trying to include parens on most function calls recently (is that a Rubocop option?) because I have nightmares around not knowing Ruby's order of operations. Can we put them on `redirect_to` here?

app/controllers/application_controller.rb
```diff
@@ -27,7 +27,7 @@ def remove_www
   end
 
   def after_sign_out_path_for(_resource_or_scope)
-    request.referrer
+    request.referer
```
glacials: lol, is this really caught by Rubocop? That's pretty funny actually

app/controllers/runs_controller.rb
```diff
@@ -3,10 +3,10 @@
 require 'speedrundotcom'
 
 class RunsController < ApplicationController
-  before_action :set_run,        only: [:show, :download, :destroy, :compare, :edit, :update]
+  before_action :set_run,        only: %i[show download destroy compare edit update]
   before_action :set_comparison, only: [:compare]
```
glacials: Does Rubocop not care about `%`-syntax for single-element arrays? & if so do we know why?

app/controllers/application_controller.rb
davygora: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Rails/RequestReferer

app/controllers/runs_controller.rb
davygora: btw, we can change config file and than we'll write `%i[compare] intead `[:compare]`

app/controllers/admin/application_controller.rb
davygora: yeap it's ruby-style-guide:
https://github.com/rubocop-hq/ruby-style-guide/blob/master/README.md#if-as-a-modifier

glacials: I was talking about this one actually: https://github.com/rubocop-hq/ruby-style-guide/blob/master/README.md#method-invocation-parens

But I'm not sure if `redirect_to` counts as part of a Rails DSL.

app/controllers/runs_controller.rb
glacials: Unless there are any good arguments against that, let's go ahead and do it for consistency.

app/controllers/admin/application_controller.rb
davygora: Oh sorry, yes we can use parens with `redirect_to`, but in this case parens - useless.
Let's leave everything as is.

glacials:
Closing for inactivity, reopen if needed.

PR #436, opened 2018-08-18
Fix rubocop errors in spec, helper, rakefile, config.ru

davygora:
Related to: https://github.com/glacials/splits-io/issues/331
Fixed rubocop errors in:
`Rakefile`
`app/helpers/rivalries_helper.rb` - deleted
`config.ru`
`spec/controllers/api/v3/games_controller_spec.rb`


app/helpers/rivalries_helper.rb
```diff
@@ -11,7 +11,7 @@ module RivalriesHelper
     'connect your foot with their butt'
     'connect their butt with your foot'
     'kick kick kick kick kick kick'
-  ].freeze
+  ".freeze
```
glacials: Does this work? `BUTT_KICK_SENTENCES` should be an array of strings, but this looks like it's one big string now.

davygora: `%q[]` - it's not an array of strings
we need to use `%w[]` for array of strings :)

image

glacials: Oh wow, I guess this never worked then haha. It actually looks like the only method we use this in (`next_butt_kick_sentence` below) is never called, so let's actually just kill this whole file then.

PR #437, opened 2018-08-20
Add account stats to frontpage

glacials:
Adds some account stats to the frontpage in the same style as the run page quick stats:

- Run storage space being used
- Number of runs
- Account age

Also moves the Patreon link out of the user dropdown menu, next to Discord.


app/models/concerns/runner_user.rb
```diff
@@ -0,0 +1,18 @@
+require 'active_support/concern'
+
+module RunnerUser
+  extend ActiveSupport::Concern
+
+  included do
+    def s3_bytes_used(since: 100.years.ago)
+      runs.pluck(:s3_filename).sum(0) do |s3_filename|
+        o = $s3_bucket_internal.object("splits/#{s3_filename}")
```
davygora: `o` - let's think up a better name for this variable :)

BatedUrGonnaDie: Could always just go with `run_object`.

app/views/layouts/application.slim
```diff
@@ -58,19 +58,20 @@ html
                   = button_tag(type: 'submit', class: 'btn btn-primary', name: nil) do
                     = icon('fas', 'search')
           ul.nav.navbar-nav
+            li.nav-item
```
BatedUrGonnaDie: Previously the patreon link vanishes if you are a patreon, and this will change it to always visible.  Is this correct behavior?

app/views/runs/index.slim
```diff
@@ -21,6 +21,28 @@ article
...
+                  span> = number_to_human_size(bytes_used)
+                  small.delta-indicator.delta-positive.tip-top title='Since last year'
+                    = number_to_human_size(bytes_used_past_year)
+                span.statcard-desc Space used
+            .col-sm-3
+              .statcard.p-3
+                h2.statcard-number
+                  span> = current_user.runs.count
+                  small.delta-indicator.delta-positive.tip-top title='Since last year'
+                    = current_user.runs.where(created_at: 1.year.ago..Time.now).count
```
BatedUrGonnaDie: This should be `Time.now.utc`.

app/views/runs/index.slim
```diff
@@ -21,6 +21,28 @@ article
     .col-lg-12
       - if logged_in?
         - if current_user.runs.any?
```
BatedUrGonnaDie: Not a problem, but if you wanted to remove one level of indent for readability this `if` could be combined with the one above.

app/views/layouts/application.slim
glacials: Yeah this is intentional, but I could be convinced either way. On one hand it's a simpler user experience to have the same buttons always there and not vanish for an apparently unknown reason. On the other hand you could see the button as promoting Patreon and if you're a patron it's a useless distraction. It's a toss-up.

app/views/runs/index.slim
glacials: Damn, I always forget this.

app/views/runs/index.slim
glacials: They have different `else`s however. Might still clean this up in another branch later though, the indentation levels are getting out of hand.

BatedUrGonnaDie: Ah, I didn't think to look for `else`s, I just noticed that the indentation was getting pretty nutty.

app/views/layouts/application.slim
BatedUrGonnaDie: I think with the button being extracted to the main header from the dropdown, having the button always there is better.  That way the header always has a consistent look.

Issue #396, opened 2018-06-04
Twitchless accounts


glacials:
Not set on implementing this yet, but something to think about / investigate.

Features that depend on your Splits I/O account being backed by a Twitch account:

- Usernames (without a Twitch account, we can't guarantee you don't share a name with someone else, so we'd probably have to separate the two)
- Therefore API paths
- Follows
- Therefore rivals
- ???

PR #440, opened 2018-08-24
Move Twitch info into its own table

glacials:
Moves Twitch info into its own DB table / Rails model. This serves two purposes:

1. Cleans things up a bit; gets some Twitch user code out of a concern and into its own model
2. Gets us ready for #396

There should be no user impact from this change.


db/migrate/0_enable_extensions.rb
```diff
@@ -0,0 +1,6 @@
+class EnableExtensions < ActiveRecord::Migration[5.0]
+  def change
+    enable_extension 'uuid-ossp'
+    enable_extension 'pgcrypto'
+  end
+end
```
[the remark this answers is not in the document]
BatedUrGonnaDie: I think I had this problem before when I added a new model, but I think I somehow used the old style for ID's and got around it at the time because I wasn't sure what the problem was.

db/migrate/20180823215913_move_twitch_data_out_of_users.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+
+    remove_column :users, :twitch_token
+    remove_column :users, :twitch_display_name
+    remove_column :users, :twitch_id
+    remove_column :users, :avatar
+    remove_column :users, :email
+    remove_column :users, :twitch_user_follows_checked_at
+  end
+end
```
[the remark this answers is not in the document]
BatedUrGonnaDie: It's probably best to separate these anyways since then you could sign up with a different email then what you have linked to your twitch account.

db/migrate/20180823215913_move_twitch_data_out_of_users.rb
```diff
@@ -0,0 +1,35 @@
+class MoveTwitchDataOutOfUsers < ActiveRecord::Migration[5.2]
+  def change
```
BatedUrGonnaDie: If it's not too much work it would be nice for this to have an up and down migration so that if for some reason it needs to be reverted it can be.

app/controllers/sessions_controller.rb
```diff
@@ -2,23 +2,28 @@ class SessionsController < ApplicationController
...
+    twitch_user.assign_attributes(
+      access_token: auth.credentials.token,
+      name:         auth.info.nickname,
+      display_name: auth.info.name,
+      email:        auth.info.email,
+      avatar:       auth.info.image
     )
 
-    if user.errors.present?
+    if !twitch_user.user.save
```
davygora: better to write `unless twitch_user.user.save` instead `if !twitch_user.user.save`

glacials:
Should be all fixed.

glacials:
Nice catch! Moved the avatar logic to TwitchUser with User#avatar passing through to it. In the future with other non-Twitch auth types, this could be a user decision on which avatar to use.

I moved the default avatar logic to SessionsController#create, TwitchUser#sync!, and the migration (i.e. any place that populates the db field). The column is `NOT NULL` so we'll just guarantee from now on there will be a valid URL in there.

app/models/twitch_user.rb
```diff
@@ -1,41 +1,44 @@
...
 require 'twitch'
 
-module TwitchUser
-  extend ActiveSupport::Concern
+class TwitchUser < ApplicationRecord
+  belongs_to :user
 
-  def sync_twitch_follows!
+  def avatar
+    URI.parse(read_attribute(:avatar)).tap do |uri|
```
BatedUrGonnaDie: Rubocop yells about this, should use `self[:avatar]` instead.

app/models/twitch_user.rb
```diff
@@ -1,41 +1,44 @@
...
+  end
+
+  def sync!
+    body = JSON.parse(Twitch::User.get(twitch_id))
+
+    update(
+      twitch_id:    body['_id'],
+      name:         body['name'].downcase,
+      display_name: body['display_name'],
+      avatar:       URI.parse(body['logo'] || default_avatar).tap { |uri| uri.scheme = 'https' }.to_s
```
BatedUrGonnaDie: I believe calling `default_avatar` like this will cause an exception to be thrown as it's now defined as a class method instead of an instance method.  I tried calling it through a debug terminal above the update call and it threw no method exception.

glacials:
Good catches. Also refactored some code to satisfy Rubocop.

app/models/twitch_user.rb
```diff
@@ -3,8 +3,23 @@
...
+    twitch_user.user = User.new if twitch_user.user.blank?
+    twitch_user.user.name = auth.info.nickname
+    twitch_user.assign_attributes(
+      access_token: auth.credentials.token,
+      name:         auth.info.nickname,
+      display_name: auth.info.name,
+      email:        auth.info.email,
+      avatar:       auth.info.image || TwitchUser.default_avatar
+    )
+    self.errors.add(twitch_user.user.errors) unless twitch_user.user.save
```
BatedUrGonnaDie: Rubocop complains about redundant `self` here.  Also it doesn't appear that the new `TwitchUser` instance is ever getting saved, just the associated `User` object.

app/models/twitch_user.rb
```diff
@@ -0,0 +1,59 @@
...
+  rescue RestClient::ResourceNotFound
+    nil
+  end
+
+  def sync_follows!
+    ActiveRecord::Base.transaction do
+      current_followed_users = User.where(twitch_id: Twitch::Follows.followed_ids(twitch_id))
+
+      # If TwitchUserFollow is changed to have child records or destroy callbacks, change this to destroy_all
+      TwitchUserFollow.where(from_user: user, to_user: (twitch_followed_users - current_followed_users)).delete_all
```
BatedUrGonnaDie: This job now throws an error on `twitch_followed_users` since this is no longer in the `User` model, and it is a method defined over there.

app/models/twitch_user.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      TwitchUserFollow.import!((current_followed_users - twitch_followed_users).map do |u|
```
BatedUrGonnaDie: This line will have the same problem.

PR #441, opened 2018-08-31
Redesign upload page

glacials:
Redesigns the upload page to use Bootstrap cards, and merges the convert page into it.

screenshot 2018-08-30 23 00 33
screenshot 2018-08-30 23 00 39

Not in love with the colors or accordion, open to feedback there.

**Also**: Introduces `site_title` method in ApplicationHelper, and replaces all breadcrumbs on the whole site to use it instead of manually writing "Splits I/O". This has the benefit of making it more obvious when you're on Beta or local dev because it will say "Splits I/O (Local)" etc.

Made breadcrumbs consistently formatted while I was at it, and ditched the `title` method in ApplicationHelper because it didn't save many keystrokes and adds another layer of non-Rails-default things to learn when developing on splits-io.

NOTE: By migrating the upload page from .erb to .slim, this kills the old-ass meta tags and specific HTML formatting that old versions of LiveSplit would parse to upload runs. Merging this branch will break those versions' ability to upload. It's been a few years since that functionality was replaced with API support, so we should be fine to finally kill this. cc @CryZe @wooferzfg


wooferzfg:
I think it's been like 3.5 years since those old versions of LiveSplit, so yeah, feel free to get rid of it.

glacials:
Fixed by removing the text under a xl width, and the Discord link under an lg width.

screenshot 2018-09-01 10 09 27

PR #445, opened 2018-09-12
Add check on typeahead to see if it is already added

### LINKED ISSUE #439: Search field sometimes breaks
untitled

Sometimes when pressing the Back button, the Search field's autocomplete breaks the field like above.

BatedUrGonnaDie:
Closes #439

### DIFF
--- a/app/javascript/search.js
+++ b/app/javascript/search.js
@@ -1,6 +1,10 @@
 import Bloodhound from 'typeahead.js'
 import Handlebars from 'handlebars'
 
+document.addEventListener("turbolinks:before-cache", function() {
+  $('.search').typeahead('destroy')
+})
+
 document.addEventListener('turbolinks:load', function() {
   $('.search').typeahead({
     minLength: 3,


BatedUrGonnaDie:
Looks like sometimes search won't trigger after hitting back a few times, needs more work.

BatedUrGonnaDie:
Alright, turns out VBox is still trash and doesn't actually update JS files until you restart the docker container, so that's why it looked like it worked before.  This time it should actually be fixed!

BatedUrGonnaDie:
Some background on this: since the typeahead actually modifies the DOM with a new wrapper and input and stuff, when you leave a page Turbolinks caches the page to show it on back/forward without making a new request (also as a preview if you go back to it but that doesn't apply here).  So basically it would cache the page with the DOM modifications, and then when it restores from the cache on back/forward it doesn't actually fetch a new DOM from the backend which causes it to load the version that already has the modifications.  Then the `turbolinks:load` event fires and it inserts another typeahead, which then nests itself indefinitely as long as you keep hitting back/forward.  This also effected the regular search page if that was the page you hit back/forward on.

BatedUrGonnaDie:
The first solution was to try and prevent typeahead from trying twice, but unfortunately Turbolinks does a `cloneNode(true)` to cache the DOM, which strips all event handlers from the page.  The docs around typeahead are a bit spotty, and I couldn't find anyone posting about manually reconnecting the even listeners to do it that way, which is why I went with removing it in the pre-cache stage so that it could just fire normally on load.  In the caching portion they actually say that if you need to do DOM cleanup do it in pre-cache event.  [Link](https://github.com/turbolinks/turbolinks#understanding-caching] to the caching portion which talks about this stuff in case I'm not being clear or messing something up.

PR #446, opened 2018-09-13
Remove old parse code that was no longer used

BatedUrGonnaDie:
Also removes unused api endpoint and rake task.


app/models/concerns/unparsed_run.rb
```diff
@@ -4,60 +4,10 @@ module UnparsedRun
...
-          split.realtime_best.present?
-        end && parse_result[:splits].map(&:realtime_best).sum.to_f
-      )
-
-      @convert_cache = parse_result
-      @segments_cache = parse_result[:splits]
-
-      parse_result
-    end
-
```
glacials: oh man that feels so good. this method has been around basically since day 1

spec/models/run_spec.rb
```diff
@@ -180,14 +180,10 @@
       expect(run.total_playtime_ms(Run::REAL)).to eq 77_066_934
     end
 
-    it 'reports no history using fast parsing' do
-      expect(run.parse(fast: true)[:realtime_history]).to be_nil
-    end
-
     it 'reports correct history when using slow parsing' do
```
glacials: "when using slow parsing" can be chopped off of this sentence it looks like

spec/models/run_spec.rb
```diff
@@ -236,13 +232,9 @@
       expect(run.total_playtime_ms(Run::REAL)).to eq 2_870_185
     end
 
-    it 'reports no history using fast parsing' do
-      expect(run.parse(fast: true)[:realtime_history]).to be_nil
-    end
-
     it 'reports correct history when using slow parsing' do
```
glacials: same

app/models/concerns/unparsed_run.rb
BatedUrGonnaDie: It's always nice to go back and clean up all the old code :)

spec/models/run_spec.rb
BatedUrGonnaDie: How does it look now?  Tried to make it sound a bit better.

PR #447, opened 2018-09-17
Show pause time on the stats page

### LINKED ISSUE #291: Add pause duration to splits stats page
Pause duration is a new field available in LiveSplit 1.7+ formats.

BatedUrGonnaDie:
Closes #291

### DIFF
--- a/app/views/runs/stats/index.slim
+++ b/app/views/runs/stats/index.slim
@@ -112,10 +112,18 @@ article#stats
                 span.time -
           tr
             td
-              span Total Playtime
+              span Total playtime
             td
               - if @run.total_playtime_ms(timing) != 0
                 span = format_ms(@run.total_playtime_ms(timing))
+              - else
+                span.time -
+          tr
+            td
+              span Total pause time
+            td
+              - if @run.pausetime_duration_ms != 0
+                span = format_ms(@run.pausetime_duration_ms)
               - else
                 span.time -
       .card-footer.clearfix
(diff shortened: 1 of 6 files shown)


BatedUrGonnaDie:
There isn't currently a LS 1.7 file for testing, so I didn't add any correct parsing tests for it.  Woofer gave me a file but I wasn't sure if you wanted me to add that to the testing flow for just one test right now.  Is that something you want me to add with this as well?

app/views/runs/stats/index.slim
```diff
@@ -112,12 +112,20 @@ article#stats
...
-              span Total Playtime
+              span Total playtime
             td
               - if @run.total_playtime_ms(timing) != 0
                 span = format_ms(@run.total_playtime_ms(timing))
               - else
                 span.time -
+          tr
+            td
+              span Total pause time
```
glacials: "Total playtime" means sum of all time put into the game across all runs, so having "total pause time" right next to it might make people think it's the sum of all time put into pauses across all runs.

BatedUrGonnaDie: Oh, this was total pause time across all runs.  I sum all of them from all the run attempts.  Did you want just for this run, the sum, or both?

glacials: Ohh, didn't notice that, my bad. What's the use case for seeing total pausetime?

BatedUrGonnaDie: Uh, my lack of reading comprehension when looking at the issue tracker...

PR #337, opened 2017-11-13
Splits I/O JSON Schema

glacials:
Adds a canonical Splits I/O JSON schema specified by [json-schema][1].

[1]: http://json-schema.org/

PR #449, opened 2018-09-22
Exchange format

glacials:
Adds support for the Splits I/O Exchange Format, first detailed in #337. Importing is supported by livesplit-core 0.10, exporting is using an ERB file (with plans to move that to an ActiveModel Serializer once things settle).

livesplit-core 0.10 also added support for worstrun and Source Live Timer, so I added modules for those as well.

cc @rmadlal @CryZe @wooferzfg


docs/api.md
```diff
@@ -87,21 +89,22 @@ render, so please only request it if you are actually using it. Be prepared for
...
+| None                             | JSON                       | `application/json`                    |
+| `application/json`               | JSON                       | `application/json`                    |
+| `application/splitsio`           | Splits I/O Exchange Format | `application/splitsio`                |
+| `application/wsplit`             | WSplit                     | `application/wsplit`                  |
+| `application/time-split-tracker` | Time Split Tracker         | `application/time-split-tracker`      |
+| `application/splitterz`          | SplitterZ                  | `application/splitterz`               |
+| `application/livesplit`          | LiveSplit                  | `application/livesplit`               |
+| `application/urn`                | Urn                        | `application/urn`                     |
+| `application/llanfair-gered`     | Llanfair-Gered             | `application/llanfair-gered`          |
+| `application/original-timer`     | Original Run File          | One of the following `Content-Type`'s |
```
CryZe: worstrun and source-live-timer are missing here, not sure if intended.

app/models/run.rb
```diff
@@ -45,8 +45,8 @@ class Run < ApplicationRecord
   class << self
     def programs
       [
-        LiveSplit, Llanfair, LlanfairGered, Llanfair2, WSplit, TimeSplitTracker, Portal2LiveTimer, Urn, SplitterZ,
-        Splitty, FaceSplit, ShitSplit
+        SplitsIO, LiveSplit, Llanfair, LlanfairGered, Llanfair2, WSplit, TimeSplitTracker, Portal2LiveTimer, Urn,
+        SplitterZ, Splitty, FaceSplit, ShitSplit
```
CryZe: worstrun and SourceLiveTimer are missing here?

docs/api.md
glacials: Yeah this is intended, this table is a list of exportable file formats, and worstrun and source-live-timer are not exportable.

app/models/run.rb
glacials: Thanks, forgot to add them here.

PR #454, opened 2018-10-01
Fixes sum depreciation notice and behavior

### LINKED ISSUE #452: Rails 6 Sum Depreciation Warning
Sum will ignore arguments when using a block, and will throw an error in 6.  `s3_bytes_used` will throw an error when upgrading.  Full error below:

`DEPRECATION WARNING: When 'sum' is called with a block, it ignores other arguments. This behavior is now deprecated and will result in an ArgumentError in Rails 6.0. (called from s3_bytes_used at /app/app/models/concerns/runner_user.rb:8)`

BatedUrGonnaDie:
`sum(0) do` was actually ignoring the 0 already and returning nil, so this fixes the behavior as well as correctly returning 0 from the function.  This also removes the begin statement which is redundant according to rubocop.

I also committed the removing of `google_users` table from schema file since that isn't currently on master, but if you want me to remove it let me know.

Closes #452

### DIFF
--- a/app/models/concerns/runner_user.rb
+++ b/app/models/concerns/runner_user.rb
@@ -5,15 +5,13 @@ module RunnerUser
 
   included do
     def s3_bytes_used(since: 100.years.ago)
-      runs.where('created_at > ?', since).sum(0) do |run|
-        begin
-          s3_run_object = $s3_bucket_internal.object("splits/#{run.s3_filename}")
-          s3_run_object.exists? ? s3_run_object.content_length : 0
-        rescue Aws::S3::Errors::Forbidden => e
-          Rollbar.warn(e, 'Missing/forbidden S3 file while counting runs', run_id: run.id, s3_filename: run.s3_filename)
-          0
-        end
-      end
+      runs.where('created_at > ?', since).sum do |run|
+        s3_run_object = $s3_bucket_internal.object("splits/#{run.s3_filename}")
+        s3_run_object.exists? ? s3_run_object.content_length : 0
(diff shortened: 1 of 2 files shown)


app/models/concerns/runner_user.rb
```diff
@@ -5,15 +5,13 @@ module RunnerUser
...
-          0
-        end
-      end
+      runs.where('created_at > ?', since).sum do |run|
+        s3_run_object = $s3_bucket_internal.object("splits/#{run.s3_filename}")
+        s3_run_object.exists? ? s3_run_object.content_length : 0
+      rescue Aws::S3::Errors::Forbidden => e
+        Rollbar.warn(e, 'Missing/forbidden S3 file while counting runs', run_id: run.id, s3_filename: run.s3_filename)
+        0
+      end || 0
```
glacials: Wow, I knew you could `rescue` without a `begin` at the method-level, but I had no idea you could do it for arbitrary blocks. Sweet!

db/schema.rb
```diff
@@ -290,7 +275,6 @@
   end
 
   add_foreign_key "game_aliases", "games", on_delete: :cascade
-  add_foreign_key "google_users", "users"
```
glacials: Thanks, this was from a branch-in-progress and it slipped into master accidentally.

app/models/concerns/runner_user.rb
BatedUrGonnaDie: I wanna say you can do it for any kind of block (method, do, if/else) but don't quote me on that.

PR #460, opened 2018-10-11
Expose stdin for web and worker containers

BatedUrGonnaDie:
This allows calls to `binding.pry` or `byebug` to be attached to instead of just ignoring them.

To attach to the instance, just run `docker attach splits-io_web_1` for the web instance, or `docker attach splits-io_worker_1`.  You can also attach to the container id found through `docker ps` if those don't work.

To detach from the container you need to exit the debug session so that execution will resume, and then do the attach escape code which is `ctrl + p` followed by `ctrl + q`.  If you do a `ctrl + c` from this you will shut down whatever the running process is (ie the web server or the worker process).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,10 @@ ifeq ($(detectedOS),Linux)
   docker-compose := sudo -E docker-compose
 endif
 
+ifeq ($(container),)
+  container := web
+endif
+
 build:
 	$(docker-compose) build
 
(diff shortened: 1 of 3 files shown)


BatedUrGonnaDie:
I added a make command to do it with a way to specific containers but defaulting to web if you don't supply one (since most of the time you would just need web).  Can you let me know how it looks/runs?

Issue #292, opened 2017-06-10
Export to Google Sheets


glacials:
With CSV exporting already in place, there's not much more work to export directly to Google Sheets. We should allow linking Google accounts then exporting to a Google Sheets sheet. We can probably make the spreadsheet look a lot prettier if we're not using CSV as a middleman.

PR #465, opened 2018-10-26
Google linking and sign in; username changes

### LINKED ISSUE #396: Twitchless accounts

glacials:
Adds Google sign in and account linking. Fixes #396.

This branch lets you disconnect your Twitch account if you have a Google account linked, and vice versa. You cannot disconnect both.

This branch is a stepping stone to completely linkless accounts (username+password, and/or email-based passwordless authentication). Adding a second type of link brought all the issues with Splits I/O accounts being strictly backed by Twitch accounts to the surface without the complexity of doing linkless at the same time. I chose Google over something else because of #292.

Because accounts aren't necessarily backed by Twitch accounts anymore, we can't rely on someone else to guarantee username uniqueness, formatting, or a reasonable initial value. So this branch also adds case-insensitive uniqueness in the database, a regex check for usernames, and a "change username" section to `/settings`.

There are two liberties I took in this branch that I think are correct, but are worth bringing up: to automatically remove old links in the face of new conflicting ones, and to automatically create links based on proof of email ownership. For example:

Initial state:
- Alice has a Google account with email mynameisalice@gmail.com
- Alice has a Twitch account with email alice@yahoo.com
1. Alice signs in with Google
2. Alice signs out
3. Alice signs in with Twitch (which creates a new Splits I/O account)
4. Alice links her new account to Google
Splits I/O will remove her previous link and create this new one,
orphaning her old account.

Initial state:
- Bob has a Google account with email bob@gmail.com
- Bob has a Twitch account with email bob@gmail.com
1. Bob signs in with Google
2. Bob signs out
3. Bob signs in with Twitch
Splits I/O will associate the Twitch link with his old account
implicitly, signing him into the same account instead of making a new
one.

There's still some todos below I want to solve, but initial feedback would be helpful.

TODO:
- [ ] Fix tests
- [ ] Solve username conflicts on sign up and username change (currently 500s)
- [ ] Add messaging like "link to Twitch first" for rivals pages
- [ ] Add messaging like "link to Twitch first" for follows (or not)
- [ ] Make sure avatars work sanely (Gravatar? Maybe later)
- [ ] Fix detailed comparison page, which I believe always tries to link to Twitch in the tips section
- [ ] Add tests


Makefile
```diff
@@ -19,6 +19,7 @@ endif
 
 build:
 	$(docker-compose) build
+	docker-compose run web bash -c "bundle exec rails db:migrate && bundle exec rails db:seed"
```
BatedUrGonnaDie: This should probably use `$(docker-compose)` and not just `docker-compose`.  Running seed from the build command though can lead down a tricky slope with duplicate entries if we're not careful.

app/models/google_user.rb
```diff
@@ -0,0 +1,53 @@
...
+    google_user
+  end
+
+  def avatar
+    URI.parse(self[:avatar]).tap do |uri|
+      uri.scheme = 'https'
+    end.to_s
+  end
+
+  def sync!
```
BatedUrGonnaDie: This sync method seems to check from twitch not from google.  If this is intended then it could be moved to a concern to DRY up some code.

app/models/google_user.rb
```diff
@@ -0,0 +1,53 @@
...
+      email:                   auth.info.email,
+      avatar:                  auth.info.image,
+      url:                     auth.info.urls.google
+    )
+
+    google_user.save
+    google_user
+  end
+
+  def avatar
```
BatedUrGonnaDie: Since this is the same between login models, this could also be in a concern.

app/models/user.rb
```diff
@@ -41,15 +45,23 @@ def self.search(term)
   end
 
   def avatar
-    twitch.avatar
+    if twitch.present?
```
BatedUrGonnaDie: With this and URI, as more login methods are added these checks seem like they would get out of hand a bit quickly.

db/migrate/20181001192951_make_user_names_case_insensitive.rb
```diff
@@ -0,0 +1,12 @@
+class MakeUserNamesCaseInsensitive < ActiveRecord::Migration[5.2]
+  def up
+    enable_extension :citext
+    change_column :users, :name, :citext
+    remove_index :users, :name
+    add_index :users, :name, unique: true
+  end
+
+  def down
+    change_column :users, :name, :text
```
BatedUrGonnaDie: Is there a reason to not put the state of the db back exactly the way it was before?

app/views/layouts/application.slim
```diff
@@ -64,7 +64,7 @@ html
             li.nav-item.d-none.d-lg-inline
               a.discord.nav-link.px-3.border-0 href='https://discord.gg/yGaxX39' target='_blank'
                 img src=asset_path('discord.svg') height='20'
-            - if logged_in?
+            - if current_user.present?
```
BatedUrGonnaDie: Out of curiosity, is there a reason to use this over `logged_in?`?

glacials:
Don't review again yet, have some more stuff I want to get to soon.

glacials:
^ There's the stuff, ready for review!

config/routes.rb
```diff
@@ -41,19 +41,22 @@
...
+  get '/auth/google/callback', to: 'google_users#in'
+  get '/auth/google/unlink',   to: 'google_users#unlink'
+
+  get '/auth/twitch/callback', to: 'twitch_users#in'
+  get '/auth/twitch/unlink',   to: 'twitch_users#unlink'
+
   get '/auth/failure', to: 'sessions#failure'
-  get '/auth/:provider/callback', to: 'sessions#create'
 
-  resources :sessions, only: [:destroy]
```
BatedUrGonnaDie: This removes the named path `session_path` which causes the `application.slim` to 500 on line 87.  Adding back `as: :session` seems to fix it

app/controllers/users_controller.rb
```diff
@@ -1,24 +1,19 @@
...
 
   def set_user
     @user = User.find_by(name: params[:user]) || not_found
   end
 
-  def verify_ownership
-    render :unauthorized, status: 401 unless @user == current_user
+  def downcase
+    if params[:user] != params[:user].downcase
+      redirect_to user_path(params[:user].downcase)
```
BatedUrGonnaDie: Is there a reason for the hard redirect here?

spec/controllers/google_users_controller.rb
```diff
@@ -0,0 +1,54 @@
+require 'rails_helper'
```
BatedUrGonnaDie: This file should have `_spec` after `controller`.

spec/controllers/twitch_users_controller.rb
```diff
@@ -0,0 +1,53 @@
+require 'rails_helper'
```
BatedUrGonnaDie: This file should have `_spec` after `controller`.

spec/controllers/twitch_users_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+        end
+      end
+
+      context 'and no redirect path' do
+        it 'redirects to root' do
+          expect(response).to redirect_to('/')
+        end
+      end
+    end
+
```
BatedUrGonnaDie: Missing `end` statement.

spec/controllers/google_users_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+        end
+      end
+    end
+  end
+
+  describe '#unlink' do
+    subject(:response) { get :unlink, params: {provider: 'google'} }
+
+    context 'when linked' do
+      expect(response).to redirect_to(settings_path)
```
BatedUrGonnaDie: `response` only available inside of `it` blocks.  Fails test once you change the name of the file.

spec/controllers/google_users_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+
+    context 'when not linked' do
+      expect(response).to redirect_to(settings_path)
```
BatedUrGonnaDie: Same `it` problem

spec/controllers/twitch_users_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+  describe '#unlink' do
+    subject(:response) { get :unlink, params: {provider: 'google'} }
+
+    context 'when linked' do
+      expect(response).to redirect_to(settings_path)
```
BatedUrGonnaDie: Same `it` problem.

spec/controllers/twitch_users_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+
+    context 'when not linked' do
+      expect(response).to redirect_to(settings_path)
```
BatedUrGonnaDie: Same `it` problem.

app/controllers/users_controller.rb
BatedUrGonnaDie: Yeah, I meant leaving it as is and just doing the lookup with the downcased username.  That was you don't add a full round trip to any request that doesn't lowercase the username (and I figured there would be a lot of that since you explicitly state you can link it in any casing on one of the pages).

glacials: I like the behavior of forcing lowercase mostly for philosophical reasons about URLs. I believe in one-URL-per-page for the internet in general.

There are some minor benefits such as helping the belief that changing your username case doesn't matter (because you already have seen URLs in full-lower that access users whose names aren't).

I've also experienced hidden caching issues in another project after not realizing we were double- or triple- (or more) caching pages because multiple URLs rendered the same user profile page. I don't think we have anything like that in Splits I/O right now, but it's just one low-chance possible future bug bite.

For supporting arguments, see the [W3 spec on HTML and URLs][1], which states "users should always consider that URLs are case-sensitive" combined with the [robustness principle][2]'s "be conservative in what you send, be liberal in what you accept" (the at-rest URL being something we "send" since people copy-paste it).

[1]: https://www.w3.org/TR/WD-html40-970708/htmlweb.html
[2]: https://en.wikipedia.org/wiki/Robustness_principle

BatedUrGonnaDie: Very good reasons, glad I asked because I didn't know the spec said to treat them as case sensitive.  Learn something new everyday.

Issue #272, opened 2017-01-27
Graphs could be smarter


jhobz:
See #271 for info on existing bugs with this graph, which bred the idea for some of these improvements.

# Idea 1: Skipped splits (or, more accurately, the split after a skipped split)
If a segment is skipped in LiveSplit, the next segment will take on the entire length of both parts of the run, obviously. This poses a problem when trying to make conclusions about progress since it can cause wild outliers in the graph.
![image](https://cloud.githubusercontent.com/assets/1765799/22354339/de96789e-e3f2-11e6-8832-41068f0b0a8f.png)
_The final split becomes several minutes longer due to 4 skipped splits before it._

Once problem 2 in #271 is resolved, it should be known if a split was skipped on any given attempt. Because of this, the SHOT graph could make assumptions when it sees a skipped split. In the image above, there may not be much that can be assumed since it's the first attempt, but if a split were to be skipped on a non-first/non-last attempt, the attempts before and after it could be used to average out the lines.
Alternatively, segments that follow a skipped segment (or series of skipped segments) could simply not render a data point on the graph. The graph could either draw a line to connect the previous and next attempts, or the line could remain broken to clearly indicate skipped splits.

# Idea 2: Improvement over real time
This may make more sense as a standalone graph (in fact, it probably would), but the SHOT graph gave me the idea, so I'll include it here. Seeing how your segments have improved over the course of attempts is very interesting and useful to see how progress is going, however, it would also be interesting to see how _quickly_ or _slowly_ that progressed happened over the course of real time. For example, you may improve by _x%_ over 100 attempts, but did those attempts take place in the same day? Same week? Same _year_? Is your first attempt of the day generally worse (or better) than the end of the day? Which month did you see the most improvement? LiveSplit stores the date and time of each attempt's start and finish (or reset); it would be cool to leverage this information. Ideally this info would be leveraged at both the segment and full run levels, which is why it may make more sense to integrate into the existing RDOT and SHOT graphs via some sort of toggle. Personally, my guess is that my graph would look like the waves you draw in grade school with my first attempt of each day being significantly worse than my last of the previous day.

# Idea 3: Regression lines
This one's kinda iffy, but regression curves could be applied to the RDOT & SHOT graphs. Would probably only make sense if there are more than some arbitrary high number of data points.

# Idea 4: Resets per Split should make assumptions about which split _caused_ the reset
[...]

PR #471, opened 2018-11-15
Patreon Reward: Amount of Practice until PB chart

### LINKED ISSUE #453: New Graphs: "Attempts Before PB" and/or "Playtime Before PB"
We have a run duration over time graph to show how you're improving, but it would be cool to have a graph that shows you how _long_ each of those improvements is taking you, in hours played or # attempts. This could also lead up to a prediction like

> Based on your history, you should PB within _n_ more attempts.

or

> Based on your history, you need about _n_ more hours of practice before you'll PB.

I like the latter more, but for many runners most practice takes place outside of runs so it might not be as objectively useful.

glacials:
Adds a new feature for $4/month and up patrons, since all they had was the gold username thingy, and it's the most popular tier :D

This is an "effort to PB" chart, including polynomial regression to predict future performance. The chart activates if the _owner_ of the splits passes the patron check.

screen shot 2018-11-14 at 5 04 05 pm

Closes #453. Addresses some parts of #272. Moves all stats page graphs to the run page.

cc @jhobz


app/javascript/graphs/pb_graph.js
```diff
@@ -3,7 +3,7 @@ import Exporting from 'highcharts/modules/exporting'
 Exporting(Highcharts)
 
 const build_pb_graph = function(run) {
-  if ($('#pb-graph-highchart').length === 0) {
+  if ($('#pb-graph').length === 0) {
```
BatedUrGonnaDie: Other graphs removed jQuery selectors, this should as well.

app/javascript/packs/application.js
```diff
@@ -9,7 +9,9 @@
 
 import "jquery"
 import "jquery-ujs"
+import PolynomialRegression from "../../../node_modules/js-polynomial-regression/dist/PolynomialRegression"
```
BatedUrGonnaDie: Instead of importing this as global, it should be instead imported into the files that need it.  The reason I imported all these to global was so that non-webpack js that was still served through the asset pipeline would continue working while the js was slowly transitioned.  It seemed to be a common approach to moving off sprockets for the js.

package.json
```diff
@@ -15,7 +15,9 @@
     "jquery-ujs": "^1.2.2",
     "jquery.turbolinks": "^2.1.0",
     "js-cookie": "^2.2.0",
+    "js-polynomial-regression": "^0.9.0",
     "moment": "^2.20.1",
+    "moment-duration-format": "^2.2.2",
```
BatedUrGonnaDie: I'm so glad someone made this finally instead of having to hack around with moment objects.

app/models/run.rb
```diff
@@ -159,6 +159,22 @@ def publish_aging
     publish_age_every(:day, 30)
   end
 
+  def playtime_between_pbs
```
BatedUrGonnaDie: Is this method called anywhere?

app/javascript/packs/application.js
glacials: Actually I didn't end up using this dependency, so I'll remove it. But when I was trying to use it, I was running into an issue I think due to the location of the dist. webpacker was giving me errors that it couldn't be found when importing until I did this. I think I just don't understand webpack or webpacker at all.

app/models/run.rb
glacials: Oops nope, leftover from before I decided to do this in JS.

package.json
glacials: Yeah seriously, that was insane. This is so much better.

CryZe:
I feel like the axes should be swapped with the PB time being the Y axis (from low at the bottom to high on the top) and then the Total Playtime on the X axis.

glacials:
Very good point! That would probably look a lot less insane than this with the reversed axis. I'll make the change!

glacials:
Looks like polynomial regression doesn't quite work for this chart type when flipped: 🙃 

screen shot 2018-11-14 at 10 42 48 pm

Logarithmic regression works well:

screen shot 2018-11-14 at 10 44 55 pm

but the `highcharts-regression` library doesn't have support for extrapolation on logarithmic regressions. I've opened https://github.com/streamlinesocial/highcharts-regression/pull/84 which implements it, so I'll let this PR sit for a short time for that to be reviewed / merged if the author chooses. If not I'll copy the raw changes over here.

Here's what it will look like with the fix:

screen shot 2018-11-14 at 10 45 45 pm

jhobz:
Was reading up on this over on the Patreon, didn't notice the cc 😄 . My only suggestion was gonna be the axes swap, so glad that already got done! I'm worried that the regression line in those screenshots doesn't seem to plateau as much as the data suggests, though that may just be due to a lack of data points. Are segment histories factored in at all, or simply PB times? A more accurate representation of what you're going for may be something like Life Playtime vs. Average difference between segment length and PB segment (i.e. you're likely to PB when your average segment times become _y_ amount better than PB, which is likely to happen around _x_ hours)? That's probably overthinking it though haha.

Either way, neat! Always excited to see more patron rewards.

glacials:
This is just based on PB times, but that's a really good idea to look at specific segment improvements. I'll look at upgrading this (or making a second graph?) in that direction a bit later!

glacials:
My PR in the regression library hasn't received any comments yet, so I'll just hardcode the changes into our vendored copy and upgrade if it ever gets merged.

PR #474, opened 2018-12-13
Autofind VODs

glacials:
Add the ability to automatically find and associate VODs of runs based on start/end timestamps included in LiveSplit splits. This is done automatically when possible on upload, but also adds a button for it to the run edit page.


app/models/run.rb
```diff
@@ -163,6 +163,30 @@ def publish_aging
...
+    Twitch::Videos.recent(user.twitch.twitch_id).each do |vod|
+      match = /^((\d+)h)?((\d+)m)?((\d+)s)?$/.match(vod['duration'])
+      hours   = match[2].to_i.hours
+      minutes = match[4].to_i.minutes
+      seconds = match[6].to_i.seconds
+
+      vod_start = DateTime.parse(vod['created_at'])
+      vod_end   = vod_start + hours + minutes + seconds
+
+      if vod_start - 2.hours < pb.started_at && pb.ended_at < vod_end + 2.hours
```
glacials: In practice I found videos up to 90 minutes and 30 seconds "off" from the times recorded in the runs, and I don't know why.

lib/parser/livesplit_core_parser.rb
```diff
@@ -69,12 +69,15 @@ def self.parse(run, fast: false)
...
 
         run_object[:history] << {
-          attempt_number: attempt_id,
+          attempt_number:       attempt_id,
+          realtime_duration_ms: time.real_time.try(:total_seconds).try(:*, 1000) || 0,
           gametime_duration_ms: time.game_time.try(:total_seconds).try(:*, 1000) || 0,
-          realtime_duration_ms: time.real_time.try(:total_seconds).try(:*, 1000) || 0
+
+          started_at: attempt.started && DateTime.parse(attempt.started.to_rfc3339),
+          ended_at:   attempt.ended   && DateTime.parse(attempt.ended.to_rfc3339)
```
CryZe: Both .started and .ended unlike the other methods return owned objects that you need to manage yourself. Or at least the GC will have to, but I'm not sure if we ever confirmed the GC integration works. So you probably want to call .dispose on what is being returned from them. (The general rule is that you need to call dispose on anything that doesn't have a Ref or RefMut postfix: https://livesplit.org/livesplit-core/classes/attemptref.html#started )

jhobz:
Relevant (dead) project that might aide in some knowledge: https://github.com/Dalet/run-highlighter. Hasn't been updated since the revamp to the highlight system, so likely none of that stuff is useful, but it used to auto-search VODs based on splits and that part might be good to look at.

glacials:
cc @Dalet!

Dalet:
*(I have not read the source code)*

As long as the attempt's timestamps were synchronized via NTP (indicated by `isStartedSynced` and `isEndedSynced` attribute on the `<Attempt>` element in the LiveSplit splits file), it should be good.
The computer clock is unfortunately unreliable, even if Windows synchronizes periodically, it isn't enough.

I've never seen an error over 30 seconds with a synchronized timestamp, it's usually accurate within 10 seconds. I assume it varies depending on how well synchronized Twitch's servers clocks are.

Also, if you didn't know: if you link to a Twitch highlighter page, you can select a segment automatically with `start`, `end` and `title` parameters in the query string. The Run Highlighter add-on isn't required anymore.

glacials:
@Dalet Thank you for the tips! I'll look at adding links to highlight after this.

lib/twitch.rb
```diff
@@ -41,21 +41,34 @@ def get(id, cursor)
...
+      JSON.parse(get(id))['data']
+    end
+
+    class << self
+      def get(id)
+        route(id).get(Twitch.headers)
+      end
+
+      def route(id)
+        Twitch.route["/videos?user_id=#{id}"]
```
Dalet: Only archive videos should be searched. The timestamp on highlights indicates when the highlight was saved, not when it was recorded.

glacials: Ohhh dang, that would explain why I thought it was 90 minutes off. I guess that's how long it takes me to highlight my PBs after the stream :O 

Until now I was actually planning on changing this to search only highlights, because even if an archive is found it will expire eventually. So it sounds like this feature is mostly useless as-is, since archives aren't good enough and highlights can't be found accurately. So I think I should add the highlight link to this branch.

jhobz:
How does this feature handle it if someone clicks the button but the archive has already expired? The default is 14 days for most users, 60 days for affiliates, partners, and Twitch Prime users. I'm guessing they just get sent to an error page on Twitch? Should there maybe be a button next to "Save" that says something along the lines of "Didn't work?" and informs people about the archive expiration?

glacials:
@jhobz The button goes away after 60 days regardless of your user type. I'm hardcoding 60 days rather than looking up your user type and choosing between 14 and 60 because the code is much simpler and I don't have to deal with edge cases where you change user types after we've already found the stream.

So if you're not a Twitch Prime user, affiliate, or partner, after 14 days and before 60 the button will take you the https://www.twitch.tv/manager page (Twitch forwards here rather than erroring). This will be confusing, maybe making you think Splits I/O's "auto-highlight" feature actually is just a dumb link to Twitch that doesn't do anything.

I think the likelihood of this happening is low enough that the trade-off for code simplicity is worth it right now. You have to
1. opt not to use the button for 14 days
2. return to your splits after 14 days and before 60 days
3. finally press the button after 14 days of not pressing it
4. not have Twitch Prime
5. stream your runs
6. not be affiliated or partnered

If we get a lot of confusion or complaints about this, or if the feature is very successful and we want to add polish, then we can revisit! I'm also open to the idea of hardcoding 14 days, which would reduce confusion at the cost of exposure to the feature. Ideally in the future we already have this information about Twitch user types onhand and up-to-date in our database and can check it easily.

app/models/highlight_suggestion.rb
```diff
@@ -0,0 +1,68 @@
...
+    Twitch::Videos.recent(run.user.twitch.twitch_id).each do |video|
+      match = /^((\d+)h)?((\d+)m)?((\d+)s)?$/.match(video['duration'])
+      hours   = match[2].to_i.hours
+      minutes = match[4].to_i.minutes
+      seconds = match[6].to_i.seconds
+
+      video_start = DateTime.parse(video['created_at'])
+      video_end   = video_start + hours + minutes + seconds
+
+      next unless video['type'] == 'archive'
```
BatedUrGonnaDie: Instead of getting all videos, why not just select that we only want archives from helix?  The `type` parameter will accept `archive` which should get rid of uploads and highlights.

lib/twitch.rb
```diff
@@ -41,21 +41,34 @@ def get(id, cursor)
...
 
       def route(id, cursor)
-        User.route(id)["/follows/channels?limit=100&cursor=#{cursor}"]
+        Twitch.route["/users/follows?from_id=#{id}&after=#{cursor}"]
+      end
+    end
+  end
+
+  module Videos
+    def self.recent(id, limit: 10)
```
BatedUrGonnaDie: Limit is not actually passed into anything.

app/models/highlight_suggestion.rb
```diff
@@ -0,0 +1,68 @@
...
+    return '-' if milliseconds.nil?
+    time = explode_ms(milliseconds)
+
+    return format('%02d:%02d:%02d.%03d', time[:h], time[:m], time[:s], time[:ms]) if precise
+    format('%02d:%02d:%02d', time[:h], time[:m], time[:s])
+  end
+
+  # explode_ms returns a hash with the components of the given duration separated out. The returned hash is guaranteed
+  # to be ordered by component size descending (hours before minutes, etc.).
+  def explode_ms(total_milliseconds)
```
BatedUrGonnaDie: Same as above.

app/models/run.rb
```diff
@@ -24,6 +24,7 @@ class Run < ApplicationRecord
   # dependent: :delete_all requires there be no child records that need to be deleted
   # If RunHistory is changed to have child records, change this back to just :destroy
   has_many :histories, class_name: 'RunHistory', dependent: :delete_all
+  has_one :highlight_suggestion
```
BatedUrGonnaDie: Is there any merit into also associating this to a user directly?

lib/parser/livesplit_core_parser.rb
BatedUrGonnaDie: This should also be addressed just in case so we're not leaking memory.

jhobz:
Oh yeah I understand the reason for hardcoding the time range; totally agree with that. What I'm wondering is should there be a simple UX flow for if someone does indeed manage to fall into that situation where they are sent to the manager for a VOD that doesn't exist. That way if the user is confused as to why there's a button but they weren't able to highlight it, they can get a simple message about what happened.

ex:
splitsiohighlightex
and clicking that button would simply display a message explaining what likely happened (it could even hide the button, if you wanted, although that could bring some issues of its own too).

app/views/runs/_title.slim
```diff
@@ -72,3 +73,36 @@ h5
...
+  .btn-group.mr-2 role='group' aria={label: 'Calls to action'}
+    - if run.belongs_to?(current_user) && run.video_url.nil? && run.highlight_suggestion.present?
+      a.btn.btn-twitch.tip-top(
+        href='#'
+        data={toggle: :modal, target: '#highlight'}
+        title='Using the timestamps in your splits, we automatically found the stream you PBed in. Click to learn more.'
+      )
+        => icon('fab', 'twitch')
+        span Auto-Highlight Possible
+- if run.belongs_to?(current_user) && run.video_url.nil? && run.highlight_suggestion.present?
```
BatedUrGonnaDie: Maybe add in a little info bubble here about video storage durations if you don't wanna go ahead with any of the other things?

app/models/highlight_suggestion.rb
BatedUrGonnaDie: I bet I typed in my comment and forgot to actually hit the comment button.

I was saying is there a way to make this and the `format_ms` functions in such a way that the model and views would be able to access them, like a class method on the `Run` model or something?  It seems hacky to have the methods defined the exact same way in 2 different places because it will probably come back to bite someone at some point.

app/models/run.rb
BatedUrGonnaDie: Yeah, it was just something that I was thinking of, not really a huge issue at this exact moment.

app/models/highlight_suggestion.rb
glacials: Yeah I struggled with this. Class methods on `Run` is probably good enough. It just feels like `format_ms` is 100% a view-oriented function, so it should go in a helper. But here we are in a model performing the exact same thing just so we can pass this string to Twitch.

Putting it on a model would work, but it would be admitting defeat and having the source of truth for a view-related function be in a model :( but I think you're right overall, that's not as big a deal as duplicating the function.

 If only ActiveSupport::Duration supported sub-seconds. Maybe long-term we should make our own duration model or gem.

lib/parser/livesplit_core_parser.rb
```diff
@@ -69,14 +69,20 @@ def self.parse(run, fast: false)
...
         }
 
+        # See https://github.com/glacials/splits-io/pull/474/files#r241242051
+        [attempt.started, attempt.ended].compact.each(&:dispose)
```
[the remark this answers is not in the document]
CryZe: No, Attempt::started() and Attempt::ended() are both methods that create new owned objects that you need to manage, as indicated by their return values not being Ref / RefMut https://livesplit.org/livesplit-core/classes/attemptref.html#started

app/controllers/runs_controller.rb
```diff
@@ -83,7 +83,7 @@ def update
 
     if params[@run.id36][:video_url]
       if @run.update(video_url: params[@run.id36][:video_url])
-        redirect_to edit_run_path(@run), notice: 'Proof saved.'
+        redirect_back(fallback_location: edit_run_path(@run), notice: 'Video saved! 📹')
```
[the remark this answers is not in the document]
glacials: I think we should just leave them around to get cleaned up later. I don't see any advantage to introducing more code to clean them up sooner. 

From the user perspective, it doesn't really matter since the "Auto-Highlight Possible" button hides when there's a video URL present on the run. The only difference is whether it appears again if you later clear the video URL, which I don't think is really a big deal either way. So let's go with the simpler code option.

lib/twitch.rb
```diff
@@ -41,21 +41,36 @@ def get(id, cursor)
...
+      JSON.parse(get(id, type))['data']
+    end
+
+    class << self
+      def get(id, type)
+        route(id, type).get(Twitch.headers)
+      end
+
+      def route(id, type)
+        return Twitch.route["/videos?user_id=#{id}&type=#{type}"]
```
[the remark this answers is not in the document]
glacials: lol somehow I always forget about pagination. Thanks for bringing this up, did some googling and was able to adapt this into a lazy Enumerator, so we still get a clean interface but don't have to make more API calls than necessary.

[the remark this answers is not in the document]
BatedUrGonnaDie: To be fair I didn't catch the pagination until it started being a problem for me XD

Also that lazy enumerator is super cool, good to know for the future.

BatedUrGonnaDie:
Looks nice and flows well (at least it does to me).

PR #479, opened 2018-12-29
Combine segment related graphs together

BatedUrGonnaDie:
- Changes reset graph from pie to column chart, and then combine this new graph with the old pb and segment history graphs.
- Changes calculations for the mean graph to now ignore splits where the previous split was skipped (old formula only rejected the 0's of the skipped splits) to prevent polluting the data with inflated values.
- Since PB graph is now combined with other historic information graphs, it is no longer shown for timers other than livesplit.

The text on the new graph doesn't sound that great to me, especially the subtitle but I really couldn't think of anything better to put than the combination of the old subtitles.


app/javascript/graph_builder.js
```diff
@@ -19,11 +17,9 @@ document.addEventListener('turbolinks:load', function() {
   }).then(function(run) {
     document.getElementById('graph-spinner').hidden = true
     document.getElementById('graph-holder').hidden = false
-    build_pb_graph(run.run)
     if (run.run.program === 'livesplit') {
```
glacials: Completely unrelated, but technically other timers could supply this data now through the exchange format. I'd like developers of new timers to get the full suite of capabilities without first having to ask us for permission, is there a simple way we can change this check to "has segment histories"?

app/javascript/graphs/segment_graphs.js
```diff
@@ -0,0 +1,109 @@
...
+        }
+      },
+      fallbackToExportServer: false
+    },
+    chart: {
+      type: 'column',
+      zoomType: 'x'
+    },
+    title: {text: 'Individual Segment Stats'},
+    subtitle: {text: '(PB - Gold) | (Average - Gold) | (1 - (Finished Attempts / Total Attempts))'},
```
glacials: Yeah I agree this is awkward, and I think always has been. We should start leaning away from mathematical descriptions and more towards human-friendly stuff. Even for me it's hard to internalize what these bars really *mean* by looking at a simple equation like `PB - Gold`, and now with three this is overload.

The subtitle should be an ELI5 like "Tall is bad". Then we can hide some more words in mouseover tooltips or something like "Difference between PB and best", "Difference between average and best".

In the future maybe we can get more creative and find ways for graphs to speak for themselves. It'd be interesting if we eventually adapted the [fixed placement columns][1] chart to show how averages are approaching bests or something.

[1]: https://www.highcharts.com/demo/column-placement

app/javascript/graphs/segment_graphs.js
```diff
@@ -0,0 +1,109 @@
...
+
+      time = ((time / counter) - segment[shortestString]) / 1000
+    }
+    meanData.push([segment.name, time])
+
+    resetData.push([segment.name, (1 - (segment.histories.length / resetCounter)) * 100])
+    resetCounter = segment.histories.length
+  })
+
+  Highcharts.chart('segment-graphs', {
```
glacials: Can we use the three colors from the Splits I/O logo as the three bar colors?

app/javascript/graphs/segment_graphs.js
```diff
@@ -0,0 +1,109 @@
...
+        time += attempt[durationString]
+        counter++
+      })
+
+      time = ((time / counter) - segment[shortestString]) / 1000
+    }
+    meanData.push([segment.name, time])
+
+    resetData.push([segment.name, (1 - (segment.histories.length / resetCounter)) * 100])
+    resetCounter = segment.histories.length
```
glacials: I see a couple of bugs on my runs with the reset bar.

On [my Sunshine Any%][1] splits, some bars are negative:
screen shot 2018-12-29 at 12 19 58

On [my Portal Glitchless][2] splits, there's a 100% reset bar on my first segment, but I've reset 0 times in those splits (only 1 attempt and it was completed):

screen shot 2018-12-29 at 12 19 49

[1]: https://splits.io/1au
[2]: https://splits.io/3emj

jhobz: Just based on those screenshots, it might be good to have the grid lines very faintly resemble the color of the graph to which they refer, when applicable. E.g. the "75" grid line come from the Reset side would be slightly tinted green. Same with the labels. The more you can express without a legend, the easier graphs become to read I think.

BatedUrGonnaDie: @glacials How old are those splits?  The bug comes from the fact that you have 1 attempt but no segment history for any of the segments, and if you make a brand new livesplit file it will have a history entry even if you only have the 1 attempt for the segment.

@jhobz That's a really good idea, thanks.  I was having a hard time trying to display all the legends while not just being a mess of information.

glacials: Not sure when the Portal Glitchless splits were made, but I downloaded the blank splits file linked from Speedrun.com and ran my first and only attempt 12 days ago.

For the Sunshine splits, could it be that old versions of LiveSplit had absent segment histories for segments that weren't reached in that attempt? Not sure if I'm understanding that correctly though.

BatedUrGonnaDie: For the sms one, your noki 4 split has 2 real history entries, while noki 5 has around 30 different ones.  So somehow it didn't track that you got to a previous split on the run but somehow still got past it.

For the portal one I'm really not sure what's up with that one then, but somehow it was fixed when I was fixing something else.

app/javascript/graphs/segment_graphs.js
```diff
@@ -0,0 +1,132 @@
...
+      }
+    }, {
+      name: 'Difference of Averages and Golds',
+      data: meanData,
+      color: logoBlue,
+      tooltip: {
+        pointFormat: '<span style="color:{point.color}">\u25CF</span> {series.name}: <b>{point.y:.1f}s</b><br/>'
+      }
+    }, {
+      name: 'Reset Percentage',
```
glacials: "Percentage" isn't super descriptive here, plus this appears in the tooltip like `Reset Percentage: 50%` which is a bit redundant. Have you considered something like "Reset Rate" or "Failure Rate"?

app/javascript/graphs/segment_graphs.js
```diff
@@ -0,0 +1,132 @@
...
+    }, {
+      title: {text: 'Reset Attempts Per Split (%)'},
+      min: 0,
+      max: 100,
+      opposite: true,
+      tickColor: logoGreen,
+      gridLineColor: logoGreen
+    }],
+    series: [{
+      name: 'Difference of PB and Golds',
```
glacials: This appears exclusively in the tooltip for a single segment, so I think "Golds" should be "Gold", since in this context you're only looking at one segment.

app/javascript/graphs/segment_graphs.js
```diff
@@ -0,0 +1,132 @@
...
+      data: pbData,
+      color: logoYellow,
+      tooltip: {
+        pointFormat: '<span style="color:{point.color}">\u25CF</span> {series.name}: <b>{point.y:.1f}s</b><br/>'
+      }
+    }, {
+      name: 'Difference of Averages and Golds',
```
glacials: Same as above, with both "Averages" -> "Average" and "Golds" -> "Gold" here.

PR #481, opened 2019-01-04
Remove legacy parser code and expose everything lsc offers

BatedUrGonnaDie:
Shuffled a lot of the returned info around, and removed the dependency on the Split model which was used no where else besides here.  Also fixes an non-reversable migration from before because I ran into issues with my schema file from a previous migration being ran and then changed later but never reran for me.

I opted to stick with just a hash return value for simplicity sake.  I was thinking about making the default for key not found raise an error but I wanted to check with you first to make sure that is the behavior that you want first.  That way errors are reported a bit better instead of just getting nil's places when we expect values to be there.


lib/parser/livesplit_core_parser.rb
```diff
@@ -26,24 +26,37 @@ def self.parse(run, fast: false)
...
 
     run_object = {
       program: program,
       game: run.game_name,
+      game_icon: run.game_icon.presence,
       category: run.category_name,
-      srdc_id: run.metadata.run_id,
+      name: run.extended_name(false),
+      metadata: {
+        srdc_id: run_metadata.run_id.presence,
```
CryZe: What does .presence do?

BatedUrGonnaDie: If the object is `present?`, then it returns the object otherwise it returns `nil`.  I put it there to change the empty strings to `nil`s so we could just do `.nil?` checks to try and standardize what checks we perform.

Here is the api dock link for the method: https://apidock.com/rails/Object/presence

CryZe: Oh, that's kind of what I figured, but the documentation (which I found too) doesn't mention the empty string to nil conversion, so I wasn't sure.

BatedUrGonnaDie: Ah gotcha. I used [this](https://stackoverflow.com/questions/885414/a-concise-explanation-of-nil-v-empty-v-blank-in-ruby-on-rails/20663389#20663389) table that glacials gave me forever ago so I know what values are blank and present and stuff.

CryZe:
I kind of have one more question. What are the icons parsed for? It seems like you don't use them in any way. If you don't, I wouldn't get them out of livesplit-core, as that's possibly megabytes of data urls, which just seems like an unnecessary cost that you would be paying. Although it would be extremely neat if you could at some point see the icons on splits i/o, but at least at the moment and especially without the downscaling, this is just wasting some performance (probably not much but still).

BatedUrGonnaDie:
For right now at least the hash was going to be the end game unless you wanted to do a custom object.  I liked how simple it was and by having the error on missing key I thought that it would add back in the security while keeping the simplicity.  What did you mean exactly by saying that it adds a layer of magic?

glacials:
Oh, how do you plan on erroring on a missing key? I right away assumed you meant overriding `Hash#[]` but I didn't think about other ways that would be less magicky like `Hash#fetch`, or maybe something else you're thinking about.

BatedUrGonnaDie:
I was just gonna set the [default proc](https://ruby-doc.org/core-2.5.3/Hash.html#method-i-default_proc-3D) after creation, but you can also just create a new hash and pass a block in to set the default proc at hash creation if you think that looks better.  I was trying to stay away from `Hash#fetch` because that requires that the person checking the hash uses fetch, instead of just getting the free safetyl

glacials:
Oh cool I did not even realize that was an option! Okay so nevermind, that doesn't seem magicky at all and I think we should do it.

db/migrate/20190104033419_remove_seconds_columns_from_run.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveSecondsColumnsFromRun < ActiveRecord::Migration[5.2]
+  def change
+    remove_column :runs, :realtime_duration_s, :decimal
+    remove_column :runs, :realtime_sum_of_best_s, :decimal
+  end
+end
```
glacials: 👍 👍 👍

lib/parser/livesplit_core_parser.rb
```diff
@@ -25,25 +25,42 @@ def self.parse(run, fast: false)
...
+        platform_name: run_metadata.platform_name.presence,
+        region_name: run_metadata.region_name.presence,
+        uses_emulator: run_metadata.uses_emulator,
+        variables: {}
+      },
```
glacials: I'm confused by what goes in the `metadata` field vs. what doesn't. I see that you're inheriting how livesplit-core returns it, but still -- is there a rule of thumb for why e.g. `region_name` is a piece of metadata but `game_icon` isn't, or even `game` for that matter?

cc @CryZe @wooferzfg

lib/parser/livesplit_core_parser.rb
```diff
@@ -25,25 +25,42 @@ def self.parse(run, fast: false)
...
-      gametime_time: 0,
+      realtime_duration_ms: 0,
+      gametime_duration_ms: 0,
       realtime_sum_of_best_ms: 0,
       gametime_sum_of_best_ms: 0,
       total_playtime_ms: 0
-    }.tap { |run_tap| run_tap[:name] = "#{run_tap[:game]} #{run_tap[:category]}".strip }
+    }
+    run_object.default_proc = key_error_lambda
+    run_object[:metadata].default_proc = key_error_lambda
```
glacials: Is there a better way of setting up the `KeyError` proc on all nested hashes? I'd be concerned about adding another nested hash later and forgetting to do this vital thing of setting the default proc.

Either way looks like `run_object[:metadata][:variables]` doesn't have the proc set.

lib/parser/livesplit_core_parser.rb
```diff
@@ -62,91 +79,82 @@ def self.parse(run, fast: false)
...
-      segment = run.segment(i)
-      split = Split.new
-      split.name = segment.name
+        started_at: attempt_started && DateTime.parse(attempt_started.to_rfc3339),
+        ended_at:   attempt_ended   && DateTime.parse(attempt_ended.to_rfc3339),
 
-      split.realtime_end = segment.personal_best_split_time.real_time.try(:total_seconds) || 0
-      split.realtime_duration = [0, split.realtime_end - run_object[:realtime_time]].max
-      split.realtime_start = split.realtime_end - split.realtime_duration
+        pause_duration_ms: attempt.pause_time.try(:total_seconds).try(:*, 1000) || 0
```
glacials: For timers that don't track pause time, I'd rather this be `nil` than `0`, so we don't misrepresent runs as having not paused when they had.

lib/parser/livesplit_core_parser.rb
```diff
@@ -62,91 +79,82 @@ def self.parse(run, fast: false)
...
+      }
+      run_object[:history].last.default_proc = key_error_lambda
```
glacials: Appending an item to an array then immediately calling `Array#last` to get it back seems a bit smelly. If `a = {}; a.default_proc = key_error_lambda; history << a` doesn't feel idiomatic enough, maybe we could have a helper that accepts a hash and returns that hash with the default proc set?

lib/parser/livesplit_core_parser.rb
```diff
@@ -62,91 +79,82 @@ def self.parse(run, fast: false)
...
-        split.gametime_gold = false
+      history_iterator = segment.segment_history.iter
+      until (history_element = history_iterator.next).nil?
+        history_element_time = history_element.time
+        split[:history] << {
+          attempt_number: history_element.index,
+          gametime_duration_ms: (history_element_time.game_time.try(:total_seconds) || 0) * 1000,
+          realtime_duration_ms: (history_element_time.real_time.try(:total_seconds) || 0) * 1000
+        }
+        split[:history].last.default_proc = key_error_lambda
```
glacials: Same as above

lib/parser/livesplit_core_parser.rb
CryZe: The metadata is mostly for the additional info we store for speedrun.com, kind of. The distinction is fairly arbitrary though.

lib/parser/livesplit_core_parser.rb
BatedUrGonnaDie: I had initially avoided adding one on `run_object[:metadata][:variables]` since there are no guarantees that there are going to be any keys present (relies on the user to set them, so any or none can be present), but I think that adding the key safety and relying on whoever is checking that hash to be careful with access to it is better after thinking about it some more.

lib/parser/livesplit_core_parser.rb
BatedUrGonnaDie: If you want me to add all run information to the metadata key I can, but up till now I was just putting what LSC classified as "metatdata" in it.

lib/parser/livesplit_core_parser.rb
BatedUrGonnaDie: This should be fixed with the fix from the other stuff.  I agree it was a dumb way to do things and I'm glad you called me out on it.

lib/parser/livesplit_core_parser.rb
```diff
@@ -62,95 +75,105 @@ def self.parse(run, fast: false)
...
+
+  def self.traverse(obj)
+    traverse_hash(obj)
+  end
+
+  def self.traverse_hash(hash)
+    hash.default_proc = ->(_hash, key) { raise KeyError, "#{key} not found" }
+    hash.each do |_key, value|
+      traverse_hash(value) if value.is_a?(Hash)
+      traverse_array(value) if value.is_a?(Array)
```
glacials: Using `is_a?` is generally viewed as bad practice in Ruby because it stops duck typing from working (e.g. `ActiveRecord::Relation`s behaving like arrays). Prefer `respond_to?` to check if you can call `each` and/or `default_proc=`. If you can, great! Do it. Doesn't matter if it's a `Hash` deep down or not.

lib/parser/livesplit_core_parser.rb
```diff
@@ -62,95 +75,105 @@ def self.parse(run, fast: false)
...
+      run_object[:gametime_duration_ms] += split[:gametime_duration_ms] if split[:gametime_duration_ms].present?
       run_object[:splits] << split
     end
 
     run.dispose
+    traverse(run_object)
     run_object
   end
+
+  def self.traverse(obj)
```
glacials: Does this modify `obj` or create a new one and return it? I think you have to do something special in Ruby to modify an argument and have its changes persist in the caller.

If you do, I'd also add a `!` to the end of this method name to communicate that it does that.

lib/parser/livesplit_core_parser.rb
```diff
@@ -62,95 +75,105 @@ def self.parse(run, fast: false)
...
+
+  def self.traverse_hash(hash)
+    hash.default_proc = ->(_hash, key) { raise KeyError, "#{key} not found" }
+    hash.each do |_key, value|
+      traverse_hash(value) if value.is_a?(Hash)
+      traverse_array(value) if value.is_a?(Array)
+    end
+  end
+
+  def self.traverse_array(array)
```
glacials: `traverse_hash` and `traverse_array` look so similar, have you explored the various `each`-like methods each has to see if there's one loop that could work for both?

lib/parser/livesplit_core_parser.rb
BatedUrGonnaDie: I actually didn't know that, good to know for the future.

lib/parser/livesplit_core_parser.rb
BatedUrGonnaDie: It does modify the original hash, just forgot about the `!` and `?` stuff that ruby does.  I don't know about doing anything special though, since what it does currently modified it since I was testing it out in the console.

lib/parser/livesplit_core_parser.rb
BatedUrGonnaDie: Yeah, I wasn't super happy with it the way it was, but didn't want to spend a ton of time on it right away if you didn't like the direction I went with it and then have to scrap it.  Should be better now.

lib/parser/livesplit_core_parser.rb
```diff
@@ -62,95 +75,99 @@ def self.parse(run, fast: false)
...
+  def self.add_default_proc!(obj)
+    traverse_enumerable(obj)
+  end
+
+  def self.traverse_enumerable(obj)
+    obj.default_proc = ->(_hash, key) { raise KeyError, "#{key} not found" } if obj.respond_to?(:default_proc=)
+    return unless obj.respond_to?(:each)
+
+    obj.each do |value|
+      traverse_enumerable(value)
```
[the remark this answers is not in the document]
BatedUrGonnaDie: Yeah, it does assume we aren't doing anything crazy with the keys, but I don't think it should be a problem since part of this is to simplify the output.

lib/parser/livesplit_core_parser.rb
```diff
@@ -62,95 +75,105 @@ def self.parse(run, fast: false)
...
+        icon: segment.icon.presence,
+        realtime_gold: false,
+        realtime_skipped: false,
+        gametime_gold: false,
+        gametime_skipped: false,
+        history: []
+      }
+
+      split[:realtime_end_ms] = (segment.personal_best_split_time.real_time.try(:total_seconds) || 0) * 1000
+      split[:realtime_duration_ms] = [0, split[:realtime_end_ms] - run_object[:realtime_duration_ms]].max
```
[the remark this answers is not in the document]
BatedUrGonnaDie: I really feel like Github is deleting my comments now since this is more than once now it's happened.

At the end of each split iteration, it adds the current split duration to the running counter to get the actual duration of the run (as opposed to just checking the last split end time), which ends up being the same as the last splits end.  This just means we don't have to make a special case for the first iteration.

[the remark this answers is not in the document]
glacials: Ohhh ok, so `run_object[:realtime_duration_ms]` is the run duration _so far_, at the end of the last split. Got it! Also didn't realize before that this is basically unchanged from previous behavior 👍

PR #485, opened 2019-01-12
Fix times always 0 in api v3

BatedUrGonnaDie:


app/serializers/api/v3/run_serializer.rb
```diff
@@ -13,4 +13,8 @@ def sum_of_best
   def name
     object.to_s
   end
+
+  def time
+    object.realtime_duration_ms / 1000
```
glacials: Does this also need to be `to_f`ed and/or `|| 0`ed like above?

BatedUrGonnaDie: Excluding the `|| 0` causes a test to error and I'm not sure why, and the `to_f` was just trying to making it consistent with what is done in the detail serializer and that one was to be consistent with how segment times are made.

PR #487, opened 2019-01-18
Link games with Speedrun.com

### LINKED ISSUE #182: Link to speedrun.com page from game show page

glacials:
Gives games auto-associating Speedrun.com links, like the SpeedRunsLive ones they've always had. Prefer using Speedrun.com shortnames over SpeedRunsLive ones. Fixes #182.


app/models/speedrun_dot_com_game.rb
```diff
@@ -0,0 +1,28 @@
...
+    new(srdc_id: result['id'], game: game).from_result!(result)
+  end
+
+  def sync!
+    from_result!(SpeedrunDotCom::Game.from_id(srdc_id))
+  end
+
+  def from_result!(result)
+    update(
+      name:           result['names']['international'],
```
BatedUrGonnaDie: It might be worth storing the twitch name as well if you ever plan to add more direct links to things on twitch.  I'm not sure if it's present in the docs but for things like Yoshi's Island its there.

https://www.speedrun.com/api/v1/games/nd280rd0

app/models/speedrun_dot_com_game.rb
```diff
@@ -0,0 +1,28 @@
...
+      shortname:      result['abbreviation'],
+      url:            result['weblink'],
+      favicon_url:    result['assets']['icon']['uri'],
+      cover_url:      result['assets']['cover-large']['uri'],
+      default_timing: result['ruleset']['default-time'] == 'ingame' ? 'game' : 'real',
+      show_ms:        result['ruleset']['show-milliseconds']
+    ) && self
```
BatedUrGonnaDie: What is the purpose of the `&& self`?

app/views/games/categories/_title.slim
```diff
@@ -5,14 +5,16 @@ h1 = @category.game
...
-          span Edit
+  - if @category.game.srdc.try(:url).present?
+    a.btn.btn-dark.mr-2.tip href=@category.game.srdc.url title='See on Speedrun.com'
+      = image_tag(asset_path('srdc.png'), style: 'height: 0.8em')
+  - if @category.game.srl_uri.present?
+    a.btn.btn-dark.mr-2.tip href=@category.game.srl_uri title='See on SpeedRunsLive'
+      = image_tag(asset_path('srl.png'))
+  - if can?(:edit, @category.game)
+    a.btn.btn-outline-light href=edit_game_path(@category.game)
+      i.fas.fa-edit>
```
BatedUrGonnaDie: If we're trying to move to the FA helper this could be updated now.

app/javascript/tooltips.js
```diff
@@ -2,7 +2,7 @@ import tippy from "tippy.js"
 
 document.addEventListener('turbolinks:load', function() {
   tippy('.gold-split', {placement: 'left'})
-  tippy('.tip', {placement: 'right'})
+  tippy('.tip', {placement: 'top'})
```
BatedUrGonnaDie: What's the purpose of having both `.tip` and `.tip-top`?

app/models/speedrun_dot_com_game.rb
```diff
@@ -0,0 +1,28 @@
+require 'speedrundotcom'
+
+class SpeedrunDotComGame < ApplicationRecord
+  belongs_to :game
+
+  def self.from_game!(game)
+    return game.srdc if game.srdc.present?
+
+    result = SpeedrunDotCom::Game.search(game.name).first
```
BatedUrGonnaDie: Is it safe to assume the first result will always be exactly what you're searching for?

app/javascript/tooltips.js
glacials: `.tip` should be the default, like "I don't care where the tooltip is positioned, just use whatever is consistent with the rest of the site". Right now that's `top`, but maybe later we change it.

Whereas `.tip-top` is for "I specifically want the tooltip to be above the text", which could be the case if there's blank space above the text but not left, right, or below it.

app/models/speedrun_dot_com_game.rb
glacials: Not an absolute technical guarantee, but Speedrun.com's search sorts by similarity ([docs](https://github.com/speedruncomorg/api/blob/master/version1/games.md#example-requests)), so I'd say it's as much a guarantee as we need without having to put a human approval process in place.

PR #488, opened 2019-01-22
Makefile updates and cleanup

BatedUrGonnaDie:
Make the attach command behave the same as all the rest (deferring to sudo on linux) and remove an extra seed call and duplicated echo.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,12 +21,11 @@ build:
 	$(docker-compose) run web bundle install
 	$(docker-compose) build
 	$(docker-compose) run web rails db:migrate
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -9,24 +9,23 @@ else
...
+docker := docker
 ifeq ($(detectedOS),Linux)
   docker-compose := sudo -E docker-compose
+  docker := sudo -E docker
 endif
 
-ifeq ($(container),)
-  container := web
-endif
+container ?= web
```
glacials: Nice never knew about `?=`

BatedUrGonnaDie: Yeah, when i was going through and removing those first things I found another reference that listed a bunch of cool things like this.

Issue #119, opened 2015-02-13
Game moderators


glacials:
Gotta pick these automatically somehow, just not sure how yet.

PR #498, opened 2019-02-09
Add route-only exporting

glacials:
Allows exporting as timer files without history, and makes it the default behavior for exports.

This feature went through a roller coaster of implementations over the past couple of days, at first starting out as a distinct type of Thing (model, class, views, controller, database table, the whole nine), and after I was 90% done building that I realized it actually really sucked from a UX perspective.

This method is much simpler and leverages what users already do: upload a run, point at it, and say "here's my route". Nothing new to learn, no new pages to add or maintain, and no duplication of data between the run/segment database tables and almost-identical route/route segment database tables.

Once we get #119 done I'd also like to allow game moderators to pin runs to the game/category pages under a "routes" section with download links, to make them feel like proper routes, but for now this is as far as we get.


app/controllers/runs/exports_controller.rb
```diff
@@ -2,22 +2,6 @@ class Runs::ExportsController < Runs::ApplicationController
   before_action :set_run, only: [:index, :timer, :history_csv, :segment_history_csv]
```
BatedUrGonnaDie: The `:index` option should be removed as well.

Makefile
```diff
@@ -22,6 +22,7 @@ build:
 	$(docker-compose) build
 	$(docker-compose) run web rails db:migrate
 	@[ -e tmp/seed ] || make seed
+	$(docker-compose) run web skylight disable_dev_warning
```
BatedUrGonnaDie: Since this only needs to be ran once, should this be part of the build command that is meant to be ran many times?

glacials: Every time I `make build` it needs to be run again, so I think it does need to be here. Let me know if you have a different experience.

app/javascript/run_export.js
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  setRoutesOnly(checkbox.checked)
+
+  checkbox.addEventListener('change', (event) => {
+    setRoutesOnly(event.target.checked)
+  })
+
+  // Stop clicks on the checkbox's label text from closing the dropdown menu
+  document.getElementById('route-check-label').addEventListener('click', (event) => {
```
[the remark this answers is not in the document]
glacials: This one actually doesn't look like we can, but also looks like we don't need to.

We *can't* because making a listener like
```javascript
document.addEventListener('click', event => {
  if (event.target.matches('#route-check-label')) {
    event.stopPropagation()
  }
})
```
doesn't stop the propagation in time, because the event bubbles up to `document` _after_ Bootstrap has responded to it by closing the dropdown. So the net effect is the dropdown still closes when clicking on the "Route only" label.

We *don't need to* because attaching two or three or four listeners whose content is just
```javascript
function(event) {
  event.stopPropagation()
}
```
doesn't hurt anything in the end, as even if a user navigates back and forth like a hundred thousand times without a hard load, the function would still only be called once and the `stopPropagation()` would cancel the others. It's a bit messy for sure, but I don't think it will ever have any real measurable effect.

The alternative is to fix this label-closes-the-dropdown issue by completely disabling Bootstrap's autoclose behavior and rebuilding it from scratch with this one checkbox's label as the exception. And I'm confident I'd screw that up much worse than the small morale (maybe super tiny memory?) hit of the above solution.

Makefile
BatedUrGonnaDie: That might just be my bad.  I thought I remember seeing that warning say that it only needed to be ran once but I was probably misremembering.

app/javascript/run_export.js
```diff
@@ -0,0 +1,33 @@
...
+  const checkbox = document.getElementById('route-check')
+  if (checkbox === null) {
+    return
+  }
+
+  setRoutesOnly(checkbox.checked)
+
+  // We can't attach an event listener directly to the element inside a turbolinks:load event.
+  // See https://github.com/turbolinks/turbolinks#observing-navigation-events
+  document.addEventListener('click', event => {
```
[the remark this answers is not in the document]
glacials: Oh yeah, oops!

Issue #504, opened 2019-02-11
Split graphs sometimes don't adjust size correctly


glacials:
This might be one bug or two.

(EDIT: First bug is fixed!) First, sometimes the "segment duration over time" graph is about half-width compared to the container. This can happen even if the window is never resized, and has been seen happening after collapsing then re-expanding a segment. For a video see the Discord server's #bugs channel, 2019-02-11 09:37 AM PST, recorded on Windows 8.1 64-bit with Firefox 65.0.
![unknown](https://user-images.githubusercontent.com/438911/52582009-0d90dd80-2de1-11e9-947b-4ddd165576a8.png)

Second, if you expand a graph, widen the browser window (causing the graph to widen with it), then shrink the browser window, the graph doesn't reshrink with it and you're left with the left side of the page being horizontally scrollable because of the extra width:
![screen shot 2019-02-11 at 09 40 42](https://user-images.githubusercontent.com/438911/52582251-9d368c00-2de1-11e9-8f82-4c7a4129f2e2.png)

BatedUrGonnaDie:
This should be fixed with #506 being merged.

glacials:
Looks like the first half is fixed but not the second half, so they were unrelated after all. Re-opening just to dedicate to the second bug mentioned.

PR #506, opened 2019-02-13
Remove the double api call from graph building

### LINKED ISSUE #478: Double API call
The graphs on the run page make an API call to fetch the run's history twice identically. They should be consolidated down into one.

BatedUrGonnaDie:
Closes #478 .  I think the error message for when it fails to retrieve data was also busted since there was no panels left in the layout, and I added some spacing for the spinners while the graphs are still loading.


BatedUrGonnaDie:
I was having some issues with the anchor tag reloading the page before the JS was fully parsed, so I removed it and simulated it so that is no longer an issue.  While doing this I noticed that if a graph was created while the holder is hidden, it would always be half width, but on a window resize it would change.  Upon further investigation I found [this](https://api.highcharts.com/class-reference/Highcharts.Chart#reflow) which essentially states that only window resizes reflow the layout, so I have it manually call reflow on the chart in question when we expand it to force the correct size.  So this should also close #504 now as well.

app/javascript/graph_builder.js
```diff
@@ -20,16 +52,38 @@ document.addEventListener('turbolinks:load', function() {
...
+    const segId = toggler.dataset.segment
+    const row = document.querySelector(`.segment-graph-holder[data-segment="${segId}"]`).closest('tr')
+    row.hidden = !row.hidden
+    if (!row.hidden) {
+      const segCharts = Highcharts.charts.filter(chart => {
+        return chart.renderTo.id === `segment-graph-holder-${segId}`
+      })
+      segCharts.forEach(chart => { chart.reflow() })
+    }
+  }
```
glacials: Not sure if the JS linting community agrees but unless we know otherwise, let's stick to the Ruby/Rubocop rule of preferring guard clauses
```javascript
if (!toggler) {
  return
}
// the rest
```
over
```javascript
if (toggler) {
  // the rest
}
```

app/javascript/graph_builder.js
```diff
@@ -20,16 +52,38 @@ document.addEventListener('turbolinks:load', function() {
...
+  const toggler = event.target.closest('.segment-graph-toggler')
+  if (toggler) {
+    const segId = toggler.dataset.segment
+    const row = document.querySelector(`.segment-graph-holder[data-segment="${segId}"]`).closest('tr')
+    row.hidden = !row.hidden
+    if (!row.hidden) {
+      const segCharts = Highcharts.charts.filter(chart => {
+        return chart.renderTo.id === `segment-graph-holder-${segId}`
+      })
+      segCharts.forEach(chart => { chart.reflow() })
```
glacials: For short arrow functions like this you can omit the curly braces:
```javascript
segCharts.forEach(chart => chart.reflow())
```
And not relevant here, but for future reference omitting the curly braces makes a `return` implicit as well.

app/javascript/graph_builder.js
```diff
@@ -20,16 +52,38 @@ document.addEventListener('turbolinks:load', function() {
...
+document.addEventListener('click', event => {
+  const toggler = event.target.closest('.segment-graph-toggler')
+  if (toggler) {
+    const segId = toggler.dataset.segment
+    const row = document.querySelector(`.segment-graph-holder[data-segment="${segId}"]`).closest('tr')
+    row.hidden = !row.hidden
+    if (!row.hidden) {
+      const segCharts = Highcharts.charts.filter(chart => {
+        return chart.renderTo.id === `segment-graph-holder-${segId}`
+      })
```
glacials: If it's safe to assume there will only ever be one graph holder returned from this, we can use `find` instead of `filter`, which stops looking after the first result.

BatedUrGonnaDie: For know it's safe to assume it, I just didn't know find was a thing.  Will switch to that instead since it seems more explicit what we're trying to do.

app/javascript/graph_builder.js
BatedUrGonnaDie: I saw in the old file there was a fat arrow with no return and couldn't figure out how it was working.  Good to know for the future.

BatedUrGonnaDie:
I do not know why it was doing it at half width, I just know that it was applying the width in a style attribute on the top level chart node, and when I was resizing my console it would cause the graph to fit to the expanded size.

PR #508, opened 2019-02-14
Add Speedrun.com category links

glacials:
Adds associations between our categories and Speedrun.com's categories, in the same way as games from #487.

Right now, this is only used to change the button on the game/category page to link to the specific Speedrun.com category. Later, after adding Speedrun.com user links, these category links will help with powering auto-submissions to Speedrun.com.

### DIFF
--- a/app/models/category.rb
+++ b/app/models/category.rb
@@ -4,6 +4,8 @@ class Category < ApplicationRecord
   has_many :users, through: :runs
   has_many :rivalries, dependent: :destroy
 
+  has_one :srdc, class_name: 'SpeedrunDotComCategory', dependent: :destroy
+
   before_create :autodetect_shortname
 
   def self.global_aliases
@@ -46,6 +48,15 @@ class Category < ApplicationRecord
 
   def to_param
     id.to_s
+  end
+
+  def sync_with_srdc
+    if srdc.nil?
+      SpeedrunDotComCategory.from_category!(self)
+      return
+    end
+
+    srdc.sync!
   end
 
   def autodetect_shortname
--- a/app/models/speedrun_dot_com_category.rb
+++ b/app/models/speedrun_dot_com_category.rb
@@ -7,7 +7,7 @@ class SpeedrunDotComCategory < ApplicationRecord
     return category.srdc if category.srdc.present?
 
     if category.game.srdc.nil?
-      srdc_game = SpeedrunDotComGame.from_game!(category.game.srdc.srdc_id).persisted?
+      srdc_game = SpeedrunDotComGame.from_game!(category.game)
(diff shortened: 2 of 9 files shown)


app/models/category.rb
```diff
@@ -4,6 +4,8 @@ class Category < ApplicationRecord
   has_many :users, through: :runs
   has_many :rivalries, dependent: :destroy
 
+  has_one :srdc, class_name: 'SpeedrunDotComCategory', dependent: :destroy
```
BatedUrGonnaDie: `srdc` is used around the app to indicate speedrun.com by itself, should this be something like `srdcc` to indicate it's not just a reference to SRDC the site?

app/models/speedrun_dot_com_category.rb
```diff
@@ -0,0 +1,36 @@
+require 'speedrundotcom'
+
+class SpeedrunDotComCategory < ApplicationRecord
+  belongs_to :category
+
+  def self.from_category!(category)
+    return category.srdc if category.srdc.present?
+
+    if category.game.srdc.nil?
+      srdc_game = SpeedrunDotComGame.from_game!(category.game.srdc.srdc_id).persisted?
```
BatedUrGonnaDie: This line is actually throwing an error when running the job.  On line 9 you check if the game is nil, and then if it is you try and get the id from a nil object.

app/models/category.rb
glacials: It's not obvious what `srdcc` (or `srdcg`, or `srdcu`) stands for, so if we did I'd want to do something more like `srdc_category`. But imo the benefit doesn't outweigh the unwieldiness, because the context reveals the meaning, e.g.

```ruby
@game.category.srdc # a Speedrun.com category
@game.srdc          # a Speedrun.com game
@user.srdc          # a Speedrun.com user
current_user.srdc   # a Speedrun.com user
```

We'd never be faced with an object that has two Speedrun.com connections -- a `Category` only relates to a Speedrun.com category and a `Game` only relates to a Speedrun.com game, etc. If that turns out to be wrong and we eventually do have two connections in one object, we can update the relation names then.

We do this for other connections too:
```ruby
current_user.twitch
current_user.patreon
current_user.google
```

The one place context isn't as obvious is in instance methods, like inside `Category#sync_with_srdc!` where you find `srdc` naked because `self.` is implied. So far I think that's still not a huge enough deal to increase the name lengths, but I'm open to argument.

app/models/speedrun_dot_com_category.rb
glacials: Nice catch, think I brainfarted and thought I was calling `SpeedrunDotCom::Game.from_id` here or something. This method needs a `Game`, not a Speedrun.com game ID.

Issue #462, opened 2018-10-15
Discord link is not visible on mobile


glacials:
The Discord link is always hidden on mobile, even when dropping down the full menu.

jamacanbacn:
Issue can be closed now 👍

PR #509, opened 2019-02-15
Fix discord icon missing on mobile

jamacanbacn:
Not sure the original reason for d-none being used but I removed it and tested it through mobile and many responsive sizes.
https://github.com/glacials/splits-io/issues/462


jamacanbacn:
Accidentally pushed a different pr :) All good now

PR #510, opened 2019-02-19
Add a small suite of profiling tools

BatedUrGonnaDie:
The PR adds `derailed_benchmarks` and `rack-mini-profiler` to better allow comparing or changes made to various things.  Details for using both are explained very well in their readmes.

Derailed allows us to do dynamic benchmarking with production, but requires us to be able to run it locally, so I made a rake task that will allow us to toggle various connection information in order for it to be able to boot up.

Rack mini profiler adds a lot of different ways to track changes, some of which are covered by the rails chrome extension, but this also allows us to generate flamegraphs, track memory usage per request, and a host of other things through the sub gems that it uses.


BatedUrGonnaDie:
Almost forgot, rack mini profiler can also be ran in production (on a whitelisted request basis) to see all the timings like you would in development.  Is that something you would want included with this?

BatedUrGonnaDie:
The reason that I went with this approach initially was because it is quite similar to what rails does to allow caching in development.  If a file is present then it toggles the caching behavior in `config/environment/development.rb`.  I was also trying to keep all the individual config related components all together in one place (all db stuff in one file, all production stuff in one file).

I strayed away from adding more docker stuff because it would largely be identical and would be another file to maintain (I have forgotten to update `Dockerfile-production` once already when updating the base file).

I think adding a new environment is probably the way to go then, so I'll get to work on moving stuff to that.

PR #511, opened 2019-02-20
Add resize event handler to update segment graph sizes

### LINKED ISSUE #504: Split graphs sometimes don't adjust size correctly

BatedUrGonnaDie:
One thing to note is that setting the size on the graphs seems to be a pretty expensive operation.  To mitigate this a bit I did 2 things: only call `chart.setSize(...)` after not receiving anymore resize events for 250ms, and defer every call to `chart.setSize(...)` so that other things can have a chance to run off the stack.  Without the defer the page totally locks up for a few seconds, whereas with it you can at least scroll and interact with the page albeit in a very limited manner until it finishes all the resizes.  Without the debounce it makes resizing simply a nightmare with how many times it tries to resize everything even when you are still dragging the window.

The previous fix with the call to `reflow()` is still required for initial page load when the graphs are collapsed.  I did look into seeing if that was obsolete now but they seem to really be completely independent issues.

Truly fixes #504 this time.

### DIFF
--- a/app/javascript/graph_builder.js
+++ b/app/javascript/graph_builder.js
@@ -6,7 +6,7 @@ import {build_reset_graph} from "graphs/reset_graph.js"
 import {chartOptions} from "consts.js"
 import {Spinner} from 'spin.js'
 import Highcharts from 'highcharts'
-import _ from 'lodash'
+import _ from 'underscore'
 
 document.addEventListener('turbolinks:load', function() {
   if (document.getElementById('graph-holder') === null) {
@@ -101,6 +101,7 @@ document.addEventListener('click', event => {
   segChart.reflow()
 })
 
+// Use debounce to collect all resize events to fire it once instead of every single time
 window.addEventListener('resize', _.debounce(() => {
   Highcharts.charts.forEach((chart) => {
     const row = chart.renderTo.closest('tr')
@@ -108,6 +109,8 @@ window.addEventListener('resize', _.debounce(() => {
       return
     }
 
+    // Use defer to allow other JS to run on the stack in between chart sizing changes
+    // Without this the page will completely lock up for a bit while all charts are resized
     _.defer((size) => {
       chart.setSize(size)
     }, chart.container.closest('table').closest('.card').clientWidth)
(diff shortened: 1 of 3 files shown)


app/javascript/graph_builder.js
```diff
@@ -99,3 +100,16 @@ document.addEventListener('click', event => {
...
+window.addEventListener('resize', _.debounce(() => {
+  Highcharts.charts.forEach((chart) => {
+    const row = chart.renderTo.closest('tr')
+    if (row === null) {
+      return
+    }
+
+    _.defer((size) => {
+      chart.setSize(size)
+    }, chart.container.closest('table').closest('.card').clientWidth)
```
glacials: Can you also include the reason for this defer being necessary in a comment here? Since it's not obvious looking at the code on its own.

app/javascript/graph_builder.js
```diff
@@ -6,6 +6,7 @@ import {build_reset_graph} from "graphs/reset_graph.js"
 import {chartOptions} from "consts.js"
 import {Spinner} from 'spin.js'
 import Highcharts from 'highcharts'
+import _ from 'lodash'
```
glacials: Where is the lodash library coming from? I don't see any occurrence of "lodash" in our code.

BatedUrGonnaDie: So it turns out lodash and underscore are basically the same thing, with lodash being spawned off from underscore in the past.  I saw the references to `_.xxx` functions so I messed up thinking we had it, and because webpack-merge pulls it in it wasn't causing an error.  I corrected this and found an error in the upload file that I also fixed when removing underscore from the global namespace.

PR #512, opened 2019-02-21
Add likes to runs

glacials:
Lets users like runs:

screen shot 2019-02-20 at 17 28 39
screen shot 2019-02-20 at 17 28 50

Possible future things that went through my head while building this:
- Let people see a list of their likes
- Let people see a list of others' likes
- Let people like runs from within the PBs feed on the frontpage
- Show a feed of likes on the frontpage, like the current PBs feed (maybe next to it)

Open to critique on the position / color / size / behavior / existence of this feature!

I considered using [ryaanto/acts_as_votable](https://github.com/ryanto/acts_as_votable) for this, as it could be more easily applied to other models (games, categories, etc.) and overall looks like less work to set up. I decided against it because it seemed too powerful for what I wanted, and it also seemed to entangle itself into your models' tables if you need to implement caching later, which I wasn't a fan of.


jhobz:
Hmm... what's the overall user story for this feature? If the idea is to be able to curate lists, an alternative solution might be better. If the idea is celebration, a different icon or something might be more useful. Trying to get into the headspace for the feature in terms of providing feedback.

glacials:
Great question! Definitely celebration:

> As a logged-in user looking at a friend's PB/WR, I want to pat them on the back so they feel good about their accomplishment.

but also anti-isolation:

> As a user who shares splits with the world, I want some sort of confirmation that the world is seeing them so that I don't feel like I'm just sharing splits into the void.

This also helps some future use cases like discovery of splits, and sorting runs within a game/category or search results page by social signals. _I do not want to become Twitter the we-know-better-than-you algorithm monster, please slap me if I show signs of doing this._

This feature is not for curating lists. However I anticipate that some people will use it to more-or-less bookmark splits and expect to be able to return to them later. That's fine, I'm not building the feature for that but it's really not a big deal to build a list of likes.

I'm open to icon ideas. We have all the free icons from [Font Awesome][1] available to us and of course all emoji. Maybe a clapping hands emoji?

[1]: https://fontawesome.com/icons

jhobz:
Thanks for the clarification! Yeah, I wonder if perhaps something closer to the idea of "celebration" is more fitting here. Something that really says "CONGRATS!" The classic "like" button tends to fall into "I agree with this" or "I like this content" more than "Congratulations," at least in my mind. A few ideas...
🎉 👏 💯 🥂 🎊 🍰 (cake might be a neat reference to Portal, but runs the risk of being confused with Reddit's "cake day" tradition)
A custom "PB," "GG," "CONGRATS," or similar emote may even be the way to go. Something akin to [this one used by Speedgaming.](https://www.frankerfacez.com/emoticon/28071-GG)

Of course I may be totally overthinking things in the first place, but the kind of same old like button definitely feels more like trying to mimic social media and less like offering a unique celebration experience. Perhaps another alternative to the singular like button is adding reactions with the full emoji library and/or twitch emotes and/or custom emotes. The sum of all reactions could still be used for metric purposes. Idk, just spitballing.

app/javascript/like.js
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  likeButton.classList.toggle('btn-danger')
+  likeButton.classList.toggle('btn-outline-danger')
+
+  if (likeButton.dataset.liked === '1') {
+    likeButton.dataset.liked = '0'
+    likesCount.textContent = --likesCount.dataset.value
+
+    fetch(`${gon.run.id}/like`, {method: 'delete'})
```
BatedUrGonnaDie: If this errors, will it show any indication to the user?

app/javascript/like.js
```diff
@@ -0,0 +1,22 @@
...
+  } else {
+    likeButton.dataset.liked = '1'
+    likesCount.textContent = ++likesCount.dataset.value
+
+    fetch(`${gon.run.id}/like`, {method: 'put'})
```
BatedUrGonnaDie: Same as above.

app/views/runs/_title.slim
```diff
@@ -45,6 +45,14 @@ h5
             span Claim
         - else
           button.btn.btn-outline-secondary disabled=true Sign In to Claim
+    - if can?(:create, RunLike)
+      button#like-button.btn.tip(
+        data={liked: current_user.likes?(run) ? '1' : '0'}
+        class=(current_user.likes?(run) ? 'btn-danger' : 'btn-outline-danger')
```
BatedUrGonnaDie: 500 when logged out on page load since `current_user` will be nil, then `likes?` is called on the nil object.

app/views/runs/_title.slim
```diff
@@ -45,6 +45,14 @@ h5
             span Claim
         - else
           button.btn.btn-outline-secondary disabled=true Sign In to Claim
+    - if can?(:create, RunLike)
```
BatedUrGonnaDie: If I understand this correctly, this means the like counter will not be shown at all if logged out?  If that's the case I would argue that this should still be shown, just not have any functionality, or clicking it should prompt the user to login.

db/schema.rb
```diff
@@ -192,6 +192,16 @@
     t.index ["run_id"], name: "index_run_histories_on_run_id"
   end
 
+  create_table "run_likes", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.bigint "run_id", null: false
+    t.bigint "user_id", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["run_id", "user_id"], name: "index_run_likes_on_run_id_and_user_id", unique: true
```
BatedUrGonnaDie: With the indices, doesn't having the combined index make the 2 separate ones obsolete?

app/helpers/likes_helper.rb
```diff
@@ -0,0 +1,13 @@
...
+  MAX_NAMES = 4.freeze
+
+  def tooltip(run)
+    count = run.likes.count
+
+    if count <= MAX_NAMES
+      return "Liked by #{run.likes.limit(MAX_NAMES).joins(:user).pluck(:name).to_sentence}"
+    end
+
+    return "Liked by #{run.likes.limit(MAX_NAMES - 1).joins(:user).pluck(:name).to_sentence} and #{count - (MAX_NAMES - 1)} others"
```
BatedUrGonnaDie: Since it's only 4 names, why not take the `run.likes.limit(MAX_NAMES).joins(:user).pluck(:name)` and pull it out to a variable at the top, then use slicing of the array to only get the 3 names?  This would decrease line size for rubocop and let you reuse it for both branches.

app/javascript/like.js
BatedUrGonnaDie: Do you ever plan on exposing the list of celebrated runs a user has to themselves?  From a UX perspective not being able to find the run in your list of liked because it actually didn't save would be a major pain.

glacials: Ah, true. Okay, I'll re-hollow the button if it fails.

PR #513, opened 2019-02-21
Extract video_url from run table and create separate model

BatedUrGonnaDie:
This change will allow us to have videos attached to other models, like games/categories, going forward.

Much of this PR is adding tests for all the branches of `runs_controller#update`.  I noticed when I was adding to it that an error that was being thrown accidentally wasn't being picked up by the tests, so I took a look and noticed none of the branches besides category swapping was tested.  So I went ahead and added tests for most of the branches (mostly missing error updating the various fields) in order to make sure that futures changes would have a higher chance of picking up errors.

For the `safety_assured` call in the migration, I'm not really sure what is was trying to tell me to do, so in order to test it I just added it.  I noticed one of them before in a commit but didn't really understand it.

### DIFF
--- a/app/controllers/runs_controller.rb
+++ b/app/controllers/runs_controller.rb
@@ -86,11 +86,19 @@ class RunsController < ApplicationController
     end
 
     if params[@run.id36][:video_url]
-      if @run.update(video_url: params[@run.id36][:video_url])
+      if params[@run.id36][:video_url].blank?
+        @run.video.try(:destroy)
+        redirect_back(fallback_location: edit_run_path(@run), notice: 'Video deleted.')
+        return
+      end
+
+      video = Video.find_or_initialize_by(videoable: @run)
+      if video.update(url: params[@run.id36][:video_url])
         redirect_back(fallback_location: edit_run_path(@run), notice: 'Video saved! 📹')
       else
         redirect_to edit_run_path(@run), alert: @run.errors.full_messages.join(' ')
       end
+      return
     end
 
     if params[@run.id36][:archived]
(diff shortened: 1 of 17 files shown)


BatedUrGonnaDie:
The reason I specifically chose to go with a polymorphic relationship was to go along with the idea of canonical routes for categories.  A game moderator could set up an approved route, and then include a video as well so not only could a user have pre-generated splits, they could easily look and see what the run looks like all on site without needing to go anywhere.

glacials:
As it stands right now, the future of routes is leaning towards "pinned" Runs on a category page, and not a separate Route object. So in that scenario, the video would still belong to the run.

Apart from that I'm pretty scared of polymorphism in general because of two reasons:

1. Things that look the same now often end up being different later, leaving you with a bunch of blank fields.
2. The code can get smelly trying to figure out what type of object you're dealing with.

I'm definitely able to be convinced (am still on the fence about #512 using polymorphism), but if the only argument is for routes eventually having videos then I say let's skip the polymorphism.

BatedUrGonnaDie:
Ah, I thought routes were going to be self contained objects and such.  If they are going to be pinned runs then there isn't really a need for adding polymorphism at this point in time.  It isn't that hard to add it after the fact, so I would say unless you have another feature already planed that would leverage #512 then probably just stick with how it is now.

The next logical question would be do you think extracting videos out is a good idea?  The main reason I started this was to leverage other features off the back of it.  While there is some niceness to having helpers in the model for videos, does that justify the added complexity of another model for essentially one field?

glacials:
Yeah that's true, I guess for now it's probably not worth having the extra model.

BatedUrGonnaDie:
In that case I will just close this.  We can always refer back to it later if we find some other use for having separate videos.

PR #515, opened 2019-02-22
Add Speedrun.com user links

glacials:
Adds account links to Speedrun.com accounts. No benefit atm other than a link to your Speedrun.com profile from your Splits I/O profile page. As mentioned in #508, this is the next stepping stone to an auto-submit-to-Speedrun.com feature, and I didn't want to bundle it in the same branch since they're both pretty big features.

Speedrun.com has no OAuth support, so unfortunately we have to ask for the user's API key to do the hookup.

### DIFF
--- a/app/controllers/speedrun_dot_com_users_controller.rb
+++ b/app/controllers/speedrun_dot_com_users_controller.rb
@@ -1,22 +1,34 @@
 class SpeedrunDotComUsersController < ApplicationController
   def create
+    if cannot?(:create, SpeedrunDotComUser)
+      unauthorized
+      return
+    end
+
     if SpeedrunDotComUser.from_api_key!(params[:speedrun_dot_com_user][:api_key].strip, user: current_user)
-      if URI(request.referer).path == settings_path
-        redirect_back(fallback_location: settings_path, notice: 'Speedrun.com account linked! 🏆')
-        return
-      end
-      redirect_to("#{URI(request.referer).path}?srdc_submit=1")
+      redirect
       return
     end
 
     redirect_back(
       fallback_location: settings_path,
-      alert: "Couldn't link to Speedrun.com. Did you enter your API key correctly?"
+      alert:             "Couldn't link to Speedrun.com. Did you enter your API key correctly?"
     )
   end
 
   def destroy
     current_user.srdc.try(:destroy)
     redirect_back fallback_location: settings_path, notice: 'Speedrun.com account unlinked! 🚫🏆'
+  end
+
+  private
+
+  def redirect
+    if URI(request.referer).path == settings_path
+      redirect_back(fallback_location: settings_path, notice: 'Speedrun.com account linked! 🏆')
+      return
+    end
+
+    redirect_to("#{URI(request.referer).path}?srdc_submit=1")
   end
 end
--- a/app/controllers/speedrun_dot_com_users_controller.rb
+++ b/app/controllers/speedrun_dot_com_users_controller.rb
@@ -1,0 +1,22 @@
+class SpeedrunDotComUsersController < ApplicationController
+  def create
+    if SpeedrunDotComUser.from_api_key!(params[:speedrun_dot_com_user][:api_key].strip, user: current_user)
+      if URI(request.referer).path == settings_path
+        redirect_back(fallback_location: settings_path, notice: 'Speedrun.com account linked! 🏆')
+        return
+      end
+      redirect_to("#{URI(request.referer).path}?srdc_submit=1")
+      return
+    end
+
+    redirect_back(
+      fallback_location: settings_path,
+      alert: "Couldn't link to Speedrun.com. Did you enter your API key correctly?"
+    )
+  end
+
+  def destroy
+    current_user.srdc.try(:destroy)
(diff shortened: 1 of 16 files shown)


app/models/category.rb
```diff
@@ -51,11 +51,7 @@ def to_param
   end
 
   def sync_with_srdc
-    if srdc.nil?
-      SpeedrunDotComCategory.from_category!(self)
-      return
-    end
-
+    return SpeedrunDotComCategory.from_category!(self) if srdc.nil?
```
BatedUrGonnaDie: Rubocop will yell if you don't put an empty line after guard clauses.

app/models/game.rb
```diff
@@ -45,20 +45,12 @@ def to_param
   end
 
   def sync_with_srdc
-    if srdc.nil?
-      SpeedrunDotComGame.from_game!(self)
-      return
-    end
-
+    return SpeedrunDotComGame.from_game!(self) if srdc.nil?
```
BatedUrGonnaDie: Rubocop guard clause thing as above.

app/models/game.rb
```diff
@@ -45,20 +45,12 @@ def to_param
...
     srdc.sync!
   end
 
   def sync_with_srl
-    if srl.nil?
-      SpeedRunsLiveGame.from_game!(self)
-      return
-    end
-
+    return SpeedRunsLiveGame.from_game!(self) if srl.nil?
```
BatedUrGonnaDie: Rubocop guard clause thing as above.

app/models/speedrun_dot_com_user.rb
```diff
@@ -0,0 +1,26 @@
+require 'speedrundotcom'
+
+class SpeedrunDotComUser < ApplicationRecord
+  belongs_to :user
+
+  def self.from_api_key!(api_key, user:)
```
BatedUrGonnaDie: Was there a reason to leave the colon at the end of user?

app/models/speedrun_dot_com_user.rb
```diff
@@ -0,0 +1,26 @@
...
+    srdc_user = SpeedrunDotComUser.find_by(api_key: api_key)
```
BatedUrGonnaDie: If below on line 13 we're going to leave off the `SpeedrunDotComUser` for the `new` call, I feel like it should also be left off of the `find_by` call.

app/models/speedrun_dot_com_user.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def sync!
+    from_result!(SpeedrunDotCom::User.from_id(srdc_id))
+  end
+
+  def from_result!(result)
+    update(
+      name:           result['names']['international'],
+      url:            result['weblink'],
```
BatedUrGonnaDie: Rubocop will yell about the extra comma and the alignment of the hash items.

app/models/speedrun_dot_com_user.rb
```diff
@@ -0,0 +1,26 @@
...
+    ) && self
```
BatedUrGonnaDie: Did we ever figure out if the instance version of `update(...)` will return a true/false value or the instance itself?

db/migrate/20190214000759_create_speedrun_dot_com_users.rb
```diff
@@ -0,0 +1,14 @@
+class CreateSpeedrunDotComUsers < ActiveRecord::Migration[5.2]
+  def change
+    create_table :speedrun_dot_com_users, id: :uuid do |t|
+      t.belongs_to :user, foreign_key: true, null: false, index: {unique: true}
+
+      t.string :srdc_id, null: false, index: {unique: true}
+      t.string :name,    null: false
+      t.string :url,     null: false
+      t.string :api_key, null: false, index: {unique: true}
```
BatedUrGonnaDie: Is there a reason to make an index for `api_key`?  I thought those were for searches and it doesn't appear we're doing any lookups by key.

app/controllers/speedrun_dot_com_users_controller.rb
```diff
@@ -0,0 +1,22 @@
+class SpeedrunDotComUsersController < ApplicationController
+  def create
+    if SpeedrunDotComUser.from_api_key!(params[:speedrun_dot_com_user][:api_key].strip, user: current_user)
```
BatedUrGonnaDie: Should we force logins on these actions so that `current_user` will never be nil?

app/models/speedrun_dot_com_user.rb
glacials: Yeah, I tested it and it returns `true`/`false`, so this `&& self` is still necessary.

BatedUrGonnaDie: But if it returns false, wouldn't that immediately return the false value instead of self like you want?

glacials: Yeah, that's intended. We want to return something falsey if the operation failed, or the object itself if it succeeded.

BatedUrGonnaDie: Oh, that's my bad then.  I thought the intention was to return the object itself no matter what.

PR #516, opened 2019-02-24
Fix handlebars compile warning and nil method call

BatedUrGonnaDie:
Handlebars fix based off [this](https://github.com/wycats/handlebars.js/issues/953#issuecomment-239874313) comment.

For the `NoMethodError: undefined method 'run' for nil:NilClass`, `@run_like` will always be `nil` since we only set it in the `before_action` for the `destroy` method, but since `@run` is always defined just redirect directly back to it.

### DIFF
--- a/app/views/pages/faq.slim
+++ b/app/views/pages/faq.slim
@@ -210,7 +210,7 @@ article.mt-4
           info, among other things.
       .card-footer
         .btn-group.float-right
-          a.btn.btn-dark href='https://github.com/batedurdonnadie/salty_bot'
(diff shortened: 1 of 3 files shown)


app/views/pages/faq.slim
```diff
@@ -210,7 +210,7 @@ article.mt-4
           info, among other things.
       .card-footer
         .btn-group.float-right
-          a.btn.btn-dark href='https://github.com/batedurdonnadie/salty_bot'
```
glacials: Wow oops

BatedUrGonnaDie: I was just clicking random links for fun and stumbled on it ¯\_(ツ)_/¯

Issue #517, opened 2019-02-27
Rails 6 Compatibility Tracker


BatedUrGonnaDie:
With Rails 6 release right around the corner on April 30th, this issue will serve as a would be tracker to make sure that we can upgrade as soon as possible to take advantage of the various changes and improvements.

### Gem dependencies
- [x] `active_model_serializers` - has a hard limit of < 6, there is a PR open to fix depreciation warning but does not change the gemspec definition from < 6.  If blueprinter adds root node support before rails 6 drops, then this could most likely just be dropped in favor of that pending more benchmarking.

- [x] `delayed_job_active_record` - has hard limit of < 6, but they have a Rails 6 beta branch that will most likely be merged in when rails 6 drops officially, so this really should be a problem.

- [x] `meta_request` - has hard limit of < 6, with no branch of PR's.  Last commit was April 2018 so this may or may not get updated.

- [x] `administrate` - has hard limit of < 6, with no branch or PR's.  I assume this will be updated once rails 6 officially dropped as commits are still frequent.

- [x] `rspec-rails` - throws lots of errors when running the test suite, attempting to upgrade it to dev branch at this time resulted in even more errors.

### Code Changes
After running the update task, there are a few changes that are known know.  [Source](https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html)

- [ ] The new framework defaults are located in an initializer, these should be gone through one by one and made sure that they all work, then change the default config loadout.

- [ ] There is a new cookie option, we have had issue with this in the past so it should be double checked that this works properly in beta before approving.

- [x] ActionCable is now fully distributed in ES2015, which means we should update the code related to it to use proper imports.

- [x] Looks like `rails-ujs` npm package is now `@rails/ujs`.  Update this through yarn.

This is not a perfect list, just things that I was able to find at this stage.  ActiveRecord and ActiveSupport both had pretty lengthy changelogs as well.  They both removed and depreciated a lot things, but also added some cool new helper methods.  It may be worth while to go through what was added and see where we can use them to clean up old code to make it more readable/maintainable.

### Code Depreciations
As with every release, some things got depreciated.  These are not necessarily required to get rails 6 running, but we should take care of them as fast as possible so that we don't get into a situation where we're blocked by a dumb depreciation.  Once we start getting notices we can add them in this section to track them.

- [x] `force_ssl` in the controller is now depreciated, we use this in the API.

BatedUrGonnaDie:
[...]

PR #518, opened 2019-03-03
Swap blueprinter out for active model serializer

BatedUrGonnaDie:
Blueprinter is actively being maintained, has better performance over AMS, and is quite nice to work with.  Some stats from my tests:
Response Time: 35s -> 28s (95 percentile)
Response Time: 29s -> 24s (mean time)
Memory Level Off: 1106 -> 939 (15% savings)

Unfortunately it does not pass the current view down to associations unless you explicitly pass it, which means there is a bit of copy pasting that I would've liked to put in the default view.  One other big gotcha was that if you put an incorrect view name it will silently default back to default instead of throwing an error which was a bit of a pain in the butt.

This should also fix a couple various 500's that I found while testing.

Removes the AMS block from #517 .


app/controllers/api/v3/games_controller.rb
```diff
@@ -7,18 +7,18 @@ def index
...
 
   def show
-    render json: @game, serializer: Api::V3::GameSerializer, root: :game
+    render json: GameBlueprint.render(@game, view: :api_v3, root: :game, toplevel: :game)
   end
 
   private
 
   def set_game
-    @game = Game.joins(:srdc).includes(:categories).find_by(speedrun_dot_com_games: {shortname: params[:id]}) || Game.includes(:categories).find(params[:id])
```
glacials: Any specific reason for removing these `includes` calls? Does Blueprinter handle it for us now?

BatedUrGonnaDie:
Do you think future api version will share the same fields that are currently shared between v3 and v4?  If not then there will be even less fields in the default view to the point where it might not even be included into the new api.  Right now v3 is especially susceptible to changes since we don't have any schema matching like v4 does.  Fields could change that aren't checked or new fields could be added.

Also since view is not passed along to associations and has to be manually passed, it means that only the field calls can be put in the shared view.  I was also not able to find a way to override a field defined in an included view, which means that anything that has to change between versions needs to be kept out of the shared views.  The identifier as well will show up in EVERY view if it is defined inside of ANY view.  So if v4 defines an identifier, it will be present in v3 even if we call the v3 view and v3 does not include the v4 view.

Does any of this tilt the scales for you on the shared view front?

glacials:
> The identifier as well will show up in EVERY view if it is defined inside of ANY view. So if v4 defines an identifier, it will be present in v3 even if we call the v3 view and v3 does not include the v4 view.

Wow ok that's insane. That definitely tips the scales for me. Why does this happen?

> Do you think future api version will share the same fields that are currently shared between v3 and v4? 

Yeah, that's very likely. However my personal viewpoint on bumping API versions has been gliding more towards the "we should almost never bump versions" end of the spectrum recently. I think we should stay on v4 for many years to come.

BatedUrGonnaDie:
> Wow ok that's insane. That definitely tips the scales for me. Why does this happen?

I believe it is because the entire blueprint is parsed before anything actually is rendered.  Take the following code in the current run blueprint
```ruby
view :test do
  raise
end
```
will raise a runtime error when trying to render the blueprint, even though we are calling the `:api_v4` view and the `:test` view is not included in the current view.  I believe it just picks up an identifier statement which then it interprets for all views even if you are not using it in the current view.  Since the identifier is not required though, I was able to just make the `:id` column that would normally be the identifier a normal field in the blueprints in which it doesn't remain consistent across api versions.

glacials:
🤔 That's still super weird and seems like a landmine. Are we sure Blueprinter views are meant to be used like this, to differentiate API versions? The only example I see is "standard" vs. "extended" views, which seems at odds with what we're trying to do, specify independent versions.

BatedUrGonnaDie:
Uh, no I'm not sure if views were intended for this kind of use at all.  I was trying to condense the serializers down so that it would be one file per object, but I can see how it might be a landmine in the future for someone who doesn't know about some of these weird behaviors.  Maybe it is best to just have separate files for each version of the api then.

glacials:
We could also just kill v3, it's been deprecated for 11 months now.

BatedUrGonnaDie:
Doesn't Livesplit proper and SRDC still use v3?

glacials:
Ah right, ok let's not kill it then O_o

BatedUrGonnaDie:
I do think we should make some sort of progress towards killing off v3 since it does hold back other integrations (no gametime shown on SRDC splits).  Maybe we should add a depreciation notice to the v3 doc and add a kill date for it?  For this PR though if the plan is to remove v3 in the coming weeks/months then having separate files would probably be even better, because then we can just delete all the files related to v3 instead of having to go into the blueprints and remove just the v3 stuff.

glacials:
Yeah that's a really good point. Okay, let's do that -- deprecation notice, and separate out into different files.

BatedUrGonnaDie:
Alright, I'll extract into some kind of versioned layout and then fix that includes stuff in the controllers.

BatedUrGonnaDie:
It does feel a lot less hacky having each api live in their own blueprints, and it removes pretty much all of the pitfalls that I was dodging around.

I did end up changing the output of the convert endpoint to what it's supposed to be outputting.  The docs say that it is supposed to have the output as the run endpoints, so I went ahead and updated it to reflect that (with the null id being the exception).  I also tried to fix most of the rubocop errors that were present in files that I touched.

PR #522, opened 2019-03-15
Store run filesize in db (1/2)

BatedUrGonnaDie:
This needs to be done in 2 stages, one to create the column and have new runs put into it, then a second to get the content length of existing runs and save that, then switch to `sum`ing the db column.

### DIFF
--- a/db/migrate/20190314234116_add_filesize_to_runs.rb
+++ b/db/migrate/20190314234116_add_filesize_to_runs.rb
@@ -1,0 +1,10 @@
+class AddFilesizeToRuns < ActiveRecord::Migration[5.2]
+  def up
+    add_column :runs, :filesize, :bigint
+    change_column_default :runs, :filesize, 0
+  end
+
+  def down
+    remove_column :runs, :filesize
+  end
+end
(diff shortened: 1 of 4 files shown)


db/migrate/20190314234116_add_filesize_to_runs.rb
```diff
@@ -0,0 +1,10 @@
+class AddFilesizeToRuns < ActiveRecord::Migration[5.2]
+  def up
+    add_column :runs, :filesize, :bigint
```
glacials: It could be helpful in the future to include the units in the field name (like we do for `duration_ms`), so I (or you or anyone) won't have to remember that this is bytes.

db/migrate/20190315000310_backfill_add_filesize_to_runs.rb
```diff
@@ -0,0 +1,8 @@
+class BackfillAddFilesizeToRuns < ActiveRecord::Migration[5.2]
+  disable_ddl_transaction!
+
+  def change
+    Run.in_batches.update_all filesize: 0
```
glacials: Wow I never knew about `in_batches`! That seems a lot more flexible than `find_each`.

BatedUrGonnaDie: Strong migrations actually told me about it because it didn't like adding a new column default since that would rewrite the table.

PR #529, opened 2019-03-22
Upgrade splits.io to rails 6

BatedUrGonnaDie:
Fixes most of the depreciations notices, tries to standardize most files to what a fresh rails 6 install would like like, toggles most of the 6.0 defaults (excludes backwards incompatible ones), and updates packages and gems accordingly.


BatedUrGonnaDie:
Is there any chance to get this pushed into the beta pipeline without it being merged into master?  Was hoping to make sure some of the changes related to cookies are non-breaking when in production, and it would be nice if they did to be able to easy make changes here so we don't have broken stuff on master.

app/javascript/upload.js
```diff
@@ -63,6 +63,10 @@ const uploadAll = function(files) {
 
 document.addEventListener('turbolinks:load', function() {
   // Show dropzone-overlay when a file is dragged onto the page
+  if (document.getElementById('dropzone') === null) {
+    return
+  }
```
glacials: When would this trigger?

config/environments/production.rb
```diff
@@ -30,57 +22,73 @@
...
+  # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
 
-  # Store uploaded files on the local file system (see config/storage.yml for options)
+  # Store uploaded files on the local file system (see config/storage.yml for options).
   config.active_storage.service = :local
 
+  # Mount Action Cable outside main process or domain.
+  # config.action_cable.mount_path = nil
+  # config.action_cable.url = 'wss://example.com/cable'
+  config.action_cable.allowed_request_origins = ['https://splits.io', 'https://*.splits.io']
```
glacials: Do we also need localhost here, or is that implied?

BatedUrGonnaDie: This one is only for production, which shouldn't need a connection from localhost.  In development mode it auto-whitelists localhost.

app/javascript/upload.js
BatedUrGonnaDie: The admin pages.  I believe all of the JS changes that aren't directly related to action cable or the changed package names were from that.  I saw them when I went checking to make sure they were working properly and noticed them in the console.

PR #531, opened 2019-03-27
Remove Sprockets managing all JS

### LINKED ISSUE #530: N+1 query in patreon loop
Marking this for later, but I believe this line is generating an 2n+1 query because it loads each user's patreon and twitch user models separately.

https://github.com/glacials/splits-io/blob/e79ae85fe2a9a93119d460d570cc6e7ed067c764/app/views/layouts/application.slim#L129

BatedUrGonnaDie:
Ported most of the files over to webpacker, but removed a few that seemed like they were not doing anything.  I also went ahead and modernized a couple files, but there are still some that use jQuery and don't use proper imports most likely.

Also fixed #530 .


app/javascript/spinner.js
```diff
@@ -1,18 +1,47 @@
...
+  rotate: 90,
+  direction: 1,
+  color: '#FFFFFF',
+  speed: 0.5,
+  trail: 30,
+  shadow: false,
+  hwaccel: true
+}
+
+let globalSpinner = null
```
glacials: I don't see any place where we set `globalSpinner`, am I missing something?

app/javascript/convert.js
```diff
@@ -1,16 +1,17 @@
...
+    formData.set("format", $("#convert-format-select").val())
 
-    $("#dropzone-overlay").fadeTo(125, 0.9);
-    $("#droplabel").text("Converting file...");
-    window.isUploading = true;
-    window.showSpinner("#fff");
+    $("#dropzone-overlay").fadeTo(125, 0.9)
+    $("#droplabel").text("Converting file...")
+    window.isUploading = true
+    showSpinner
```
glacials: Looks like this might not be getting called because there are no parens.

app/views/layouts/application.slim
```diff
@@ -122,11 +121,12 @@ html
           = yield(:footer)
           .row
             .col-md-6
-              - if User.joins(:patreon).where('patreon_users.pledge_cents > 0').any?
+              - patreons = User.includes(:patreon, :twitch).joins(:patreon).where('patreon_users.pledge_cents >= 200')
```
glacials: When talking about users who contribute to the Patreon, use `patrons` over `patreons`

app/javascript/analytics.js
```diff
@@ -1,76 +1,76 @@
-$(function() {
+document.addEventListener('turbolinks:load', () => {
```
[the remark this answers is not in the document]
BatedUrGonnaDie: You are definitely correct.  I extracted the GA setup stuff so that the analytics reporting will still fire every page load now but not try and set it up multiple times.

PR #534, opened 2019-04-02
Box plots

glacials:
Adds box plots for each segment!

Screen Shot 2019-04-01 at 20 39 18


glacials:
Unfortunately it doesn't look like there's a way to make the whole vertical column act as a hover target for the tooltip. There is such a thing in other Highcharts chart types, but box plots don't seem to support it.

I decided to exclude statistical outliers to help with the super compressed boxes a bit, but they still happen here and there :/

PR #535, opened 2019-04-02
Switch from delaying methods to dedicated jobs

BatedUrGonnaDie:
This also includes a migration for cleaning up *most* of the broken jobs in the db.

Pros:
- First class integration with Rails
- Can change code inside job and next run will apply it (even if it errors on the first run)
- All delayed code is in one folder

Cons:
- Added complexity (more places to look for running code)
- Lose some Delayed Job specific callbacks (Rails has most of the same callbacks though)

### DIFF
--- a/app/jobs/discover_runner_job.rb
+++ b/app/jobs/discover_runner_job.rb
@@ -1,0 +1,13 @@
+class DiscoverRunnerJob < ApplicationJob
+  queue_as :discover_runner
+
+  def perform(run)
+    srdc_runner_id = SpeedrunDotCom::Run.runner_id(run.srdc_id)
+    return if srdc_runner_id.nil?
+
+    twitch_login = SpeedrunDotCom::User.twitch_login(srdc_runner_id)
+    return if twitch_login.blank?
+
+    run.update(user: User.find_by(name: twitch_login))
+  end
+end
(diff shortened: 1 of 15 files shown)


app/jobs/discover_runner_job.rb
```diff
@@ -0,0 +1,13 @@
...
+  queue_as :discover_runner
+
+  def perform(run)
+    srdc_runner_id = SpeedrunDotCom::Run.runner_id(run.srdc_id)
+    return if srdc_runner_id.nil?
+
+    twitch_login = SpeedrunDotCom::User.twitch_login(srdc_runner_id)
+    return if twitch_login.blank?
+
+    run.update(user: User.find_by(name: twitch_login))
```
glacials: Oops, never updated this before, but this should be something like
```ruby
run.update(user: User.joins(:twitch).find_by(twitch_users: {name: twitch_login}))
```
now.

app/models/highlight_suggestion.rb
```diff
@@ -42,7 +44,7 @@ def from_run(run)
             end
           )
           # 60 days is life of archives for Partners / Twitch Prime members
-          highlight_suggestion.delay(run_at: video_start + 60.days).destroy if highlight_suggestion.persisted?
+          HighlightCleanupJob.set(wait: 60.days).perform_later(highlight_suggestion) if highlight_suggestion.persisted?
```
glacials: Any simple way to make this 60 days from video start and not 60 days from upload?

BatedUrGonnaDie:
So delayed job is still going to be the thing that actually runs the job, but now we're using ActiveJob to schedule them and actually execute them if that makes sense.

app/jobs/discover_runner_job.rb
BatedUrGonnaDie: Worked perfectly!

PR #536, opened 2019-04-05
Calculate history stats in db

BatedUrGonnaDie:


app/models/run.rb
```diff
@@ -164,6 +164,26 @@ def publish_aging
...
+                  .select('
+                    segment_id,
+                    STDDEV_POP(segment_histories.realtime_duration_ms) AS standard_deviation,
+                    AVG(segment_histories.realtime_duration_ms) AS mean,
+                    PERCENTILE_DISC(.5) WITHIN GROUP (ORDER BY segment_histories.realtime_duration_ms) AS median,
+                    PERCENTILE_CONT(.1) WITHIN GROUP (ORDER BY segment_histories.realtime_duration_ms) AS percentile10,
+                    PERCENTILE_CONT(.9) WITHIN GROUP (ORDER BY segment_histories.realtime_duration_ms) AS percentile90,
+                    PERCENTILE_CONT(.99) WITHIN GROUP (ORDER BY segment_histories.realtime_duration_ms) AS percentile99
+                  ')
+  end
```
glacials: Does there happen to be any way to put this method on `Segment` rather than on `Run`, while keeping the benefit of doing it all in one query per run?

app/models/run.rb
```diff
@@ -164,6 +164,26 @@ def publish_aging
...
+                  .where.not(realtime_duration_ms: [0, nil])
+                  .group(:segment_id)
+                  .select('
+                    segment_id,
+                    STDDEV_POP(segment_histories.realtime_duration_ms) AS standard_deviation,
+                    AVG(segment_histories.realtime_duration_ms) AS mean,
+                    PERCENTILE_DISC(.5) WITHIN GROUP (ORDER BY segment_histories.realtime_duration_ms) AS median,
+                    PERCENTILE_CONT(.1) WITHIN GROUP (ORDER BY segment_histories.realtime_duration_ms) AS percentile10,
+                    PERCENTILE_CONT(.9) WITHIN GROUP (ORDER BY segment_histories.realtime_duration_ms) AS percentile90,
+                    PERCENTILE_CONT(.99) WITHIN GROUP (ORDER BY segment_histories.realtime_duration_ms) AS percentile99
```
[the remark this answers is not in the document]
BatedUrGonnaDie: PERCENTILE_DISC takes the closest element that is actually contained in the set, while PERCENTILE_CONT will calculate it and not necessarily be present in the set.  I thought percentiles were calculated and not strictly a number present but I may have messed that up, it's been a while since I took stats.

app/models/run.rb
BatedUrGonnaDie: Does it make sense to put this on `Segment` when it generates the data for the whole run?  I was struggling to figure out where it should actually end up since it touches so many different parts.  Moving it around should be pretty easy though I think.

glacials: I more mean is there a way to make e.g. `segment.stats` or whatever return that one segment's stats, but when called in a clever way (idk, like `segments.includes(:stats)`) we get all segments' stats loaded into memory and it is still run as one query. No idea if this is possible.

Also now that I'm thinking this through, it's feels a bit weird that this method returns an array of `SegmentHistory`s -- does that array include every segment history ever recorded in the run? If this method is coalescing data _about_ segment histories scoped to each segment, returning an array of `Segment`s seems more intuitive.

app/models/run.rb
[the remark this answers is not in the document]
glacials: Any specific reason for calculating 50th percentile discretely but 10th, 90th, and 99th continuously?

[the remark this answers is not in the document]
BatedUrGonnaDie: There is no built in median function, so the workaround is 50th percentile in discrete mode to calculate it.

app/models/run.rb
BatedUrGonnaDie: Ok gotcha I see what you mean.  It's definitely possible, could have the find line that is current in the view extracted to the `Segment` model, and have that call the underlying function on the `Run` model where it is now.  That was each `Segment` can access it easily but still keep it on `Run` since it generates data for the whole run.

Is it worth converting it just to a normal hash maybe?  The data doesn't really feel like it fits into any of the existing models and it's basically a frankenstein model object the way that it currently is?

BatedUrGonnaDie:
I just noticed that this doesn't respect timings, I'll make sure to add that back in with the next commit.

app/models/run.rb
glacials: Yeah frankenstein models do make me uneasy, a hash might be simpler to wrap my head around.

PR #537, opened 2019-04-07
Autoroutes

glacials:
Automatically determine the most popular route (i.e. arrangement of splits) for each category and place a download link to it on the game and category pages.

Screen Shot 2019-04-06 at 18 45 48

Came up with this one on the toilet an hour ago when I forgot to bring my Switch with me. TODO: Stop bringing distractions to the toilet.

### DIFF
--- a/app/models/category.rb
+++ b/app/models/category.rb
@@ -87,16 +87,18 @@ class Category < ApplicationRecord
     end
   end
 
-  # route returns a run whose route is the most popular in this category, as determined by runs sharing segment names
-  # and order.
+  # route returns a run whose route is the most popular in this category, as determined by the number of runs sharing
+  # its segment names and order. Returns nil if no runs exist in this category.
   def route
-    Run.find(
-      Run.select('MIN(runs.id) as id').
+    result = Run.select('MIN(runs.id) as id').
       joins(:segments).
       where(category: self).
       group(:segment_number, :name).
       order('COUNT(*) DESC').
-      first.id
-    )
+      first
+
+    return nil if result.nil?
+
+    Run.find(result.id)
   end
 end
--- a/app/models/category.rb
+++ b/app/models/category.rb
@@ -94,7 +94,7 @@ class Category < ApplicationRecord
       joins(:segments).
       where(category: self).
       group(:segment_number, :name).
-      order('COUNT(*) DESC').
+      order(Arel.sql('COUNT(*) DESC')).
       first
 
     return nil if result.nil?
--- a/app/models/category.rb
+++ b/app/models/category.rb
@@ -86,4 +86,17 @@ class Category < ApplicationRecord
       destroy
     end
   end
+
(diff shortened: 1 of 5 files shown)


app/views/runs/_title.slim
```diff
@@ -20,7 +20,7 @@ h5
 .btn-toolbar role='toolbar' aria={label: 'Run navigation'}
   .btn-group.m-2 role='group' aria={label: 'Run navigation'}
     .btn-group role='group'
-      = render partial: 'export_button', locals: {run: run, timing: timing}
```
glacials: This wasn't being used inside the partial.

app/views/runs/_export_button.slim
```diff
@@ -1,15 +1,16 @@
 button#export-link.btn.btn-outline-light.dropdown-toggle href='#' role='button' aria={haspopup: true, expanded: false} data={toggle: 'dropdown'}
   => icon('fas', 'download')
-  span Export
```
glacials: I'm a bit on the fence about using `defined?`, but I don't really see a better way to do defaults for partial arguments.

app/models/run.rb
```diff
@@ -38,7 +38,7 @@ class Run < ApplicationRecord
   scope :by_category, ->(category) { where(category: category) }
   scope :nonempty, -> { where('realtime_duration_ms != 0') }
   scope :owned, -> { where.not(user: nil) }
-  scope :unarchived, -> { where(archived: false) }
+  scope :unarchived, -> { where(archived: false).where.not(user: nil) }
```
glacials: Unrelated, but wanted to do this to help with unremovable pollution in run listings.

app/views/runs/_export_button.slim
BatedUrGonnaDie: [This](https://api.rubyonrails.org/classes/ActionView/Base.html) says that local variables are stored in `local_assigns`,  so you could do something like `local_assigns.has_key?(:button_text)` for the conditional, or `local_assigns[:button_text].presence || 'Export'` as a one liner.  It should be noted that it also says using `defined?` is an alternative to check if it is defined right under that section.

app/models/category.rb
```diff
@@ -86,4 +86,17 @@ def merge_into!(category)
...
+
+  # route returns a run whose route is the most popular in this category, as determined by runs sharing segment names
+  # and order.
+  def route
+    Run.find(
+      Run.select('MIN(runs.id) as id').
+      joins(:segments).
+      where(category: self).
+      group(:segment_number, :name).
+      order('COUNT(*) DESC').
```
BatedUrGonnaDie: This string throws a depreciation warning, should just need to be wrapped in `Arel.sql(...)`.

app/views/runs/_export_button.slim
glacials: Oh sweet! I like this better.

PR #539, opened 2019-04-08
Become our own OAuth client

glacials:
Makes the frontend an OAuth consumer of the backend. This gives it its own access token it can use to interact with the API, instead of shortcutting it by allowing the API to accept cookies under special circumstances.

This is a bit rough right now, as the frontend's access token needs to be refreshed once in a while. This refresh is done by sending the user through the OAuth flow. We can auto-accept the grant since we're a special client, but there's still a short flash of the in-between page.

We can clean this up incrementally later by doing things like using `Turbolinks.visit` instead of `window.location =` (although I couldn't immediately get it to work here), getting this JavaScript to execute super early on in pageload, or perhaps using an iframe instead of the whole browser.


app/controllers/sessions_controller.rb
```diff
@@ -1,4 +1,8 @@
 class SessionsController < ApplicationController
+  def in
+    head :ok
```
BatedUrGonnaDie: I did `head :ok` here so that when the page loads in an iframe it doesn't load any other resources.  I tried sandboxing the iframe originally but that caused warnings to be thrown in the console when the JS couldn't run and this seemed like a cleaner option.

PR #540, opened 2019-04-09
Browse games page

glacials:
Adds a new page, splits.io/games, to browse all verified games on the site (i.e. present on Speedrun.com). Also allows you to search the page locally by using its search field. This is live at https://beta.splits.io/games.

I added this to the navigation in the header, but to keep it from getting too cluttered I decided to remove the FAQ link for logged-in users and remove the Rivals link for logged-out users, so the overall number of elements up there is the same.

Added a link to the FAQ in the footer for logged-in users looking for it.


app/javascript/search.js
```diff
@@ -86,3 +86,23 @@ document.addEventListener('turbolinks:load', function() {
     }
   })
 })
+
+document.addEventListener('turbolinks:load', () => {
+  document.getElementById('search').addEventListener('input', event => {
```
BatedUrGonnaDie: This listener is being added every time any page loads, is there any way to use event delegation for this?

db/seeds.rb
```diff
@@ -9,4 +9,43 @@
 # See https://github.com/rails/rails/issues/35812
 ActiveJob::Base.queue_adapter = Rails.application.config.active_job.queue_adapter
 
-Run.create(s3_filename: SecureRandom.uuid)
+Game.create(name: 'Assassin\'s Creed')
```
BatedUrGonnaDie: Is there a reason to add a bunch of SRDC-less games in the seed file?

app/controllers/games_controller.rb
```diff
@@ -33,4 +37,12 @@ def set_game
     end
     redirect_to game_path(@game) if @game.srdc.try(:shortname).present? && params[:game] == @game.id.to_s
   end
+
+  def set_games
+    @game_sections = SpeedrunDotComGame.order(name: :asc).slice_when do |a, b|
+      a.name[0].downcase != b.name[0].downcase
+    end.map do |srdc_games|
```
BatedUrGonnaDie: Rubocop does not like chaining blocks, and here I would agree.  We're creating a completely new array just to throw it away and create a hash from them.

jhobz:
Agreed with Bated, this looks awesome!

Couple small design notes from the beta page:
- "." is being identified as its own section, but being sorted based on the letter following it, sticking it oddly in the middle of the list
image
- I'm guessing the 3-column layout has manual breakpoints for where each one starts? This is leading to a weirdly long blank spot in the middle column after "r" as the other two columns continue on with more letters
image
(CSS Grid could be particularly helpful for that last one)

app/javascript/search.js
glacials: Is this solved now with the guard clause? Still not quite sure I 100% understand how Turbolinks and event listeners interact and how event delegation fixes about it.

db/seeds.rb
glacials: Oops, I had populated the SRDC games at the command line after and forgot to add them here.

app/controllers/games_controller.rb
```diff
@@ -39,10 +43,9 @@ def set_game
   end
 
   def set_games
-    @game_sections = SpeedrunDotComGame.order(name: :asc).slice_when do |a, b|
-      a.name[0].downcase != b.name[0].downcase
-    end.map do |srdc_games|
-      [srdc_games.first.name[0].downcase, srdc_games]
-    end.to_h
+    @games = Hash.new([])
+    SpeedrunDotComGame.order('ASCII(name) ASC').each do |game|
```
glacials: @jhobz This `ORDER BY ASCII(name)` should fix the issue with games like .hack appearing in the `h` section. I tried to group all the symbols together into one section but it was too much complexity for the payoff, so there will be one section per symbol.

jhobz:
@glacials I don't think that's what happened with the columns because there's still multiple in the left and right columns after the middle column runs out. Here's a scrolling capture I took to better demonstrate what's going on: [link because image is very tall](https://puu.sh/Dc5fk/dd8782aba3.jpg)

ASCII sort: 👍

jhobz:
Thinking about it again, I guess it's actually possible since it has to keep the sections in alphabetical order and there are several long sections in a row. Is it possible book-read order fill might be more suitable in this case?

app/controllers/games_controller.rb
BatedUrGonnaDie: If you're still using vim there is a plugin(?) that will autorun rubocop.  I'm not sure what it's called since I don't use it anymore (tried and failed to move to it) but I still have it installed if you can tell me where to look.

app/javascript/search.js
BatedUrGonnaDie: Event delegation tells the page that you want to listen for this specific event, and if it fires then call my function.  When you assign an event handler inside of a `turbolinks:load` event, every time the load event is called you tell the page you want to listen for this specific event.  Since turbolinks doesn't perform hard refreshes of the page the old event listeners are still intact, essentially having the same listener function be called multiple times.

I'm having a hard time explaining this clearly in text, so if this doesn't clear it up for you I can explain over a voice call a lot easier and with more detail.

app/controllers/games_controller.rb
```diff
@@ -25,12 +30,22 @@ def authorize
   end
 
   def set_game
-    @game = Game.joins(:srdc).find_by(speedrun_dot_com_games: {shortname: params[:game]}) || Game.find_by(id: params[:game])
+    @game = Game.joins(:srdc).find_by(speedrun_dot_com_games: {shortname: params[:game]}) ||
+            Game.find_by(id: params[:game])
+  end
 
+  def redirect_if_not_found
```
BatedUrGonnaDie: Instead of creating a new method for this we should probably be consistent with other controller where it uses `Game.find` and then `rescue`s from the method if the record doesn't exist.

app/controllers/games_controller.rb
```diff
@@ -25,12 +30,22 @@ def authorize
...
       redirect_to games_path(q: params[:game])
       return
     end
     redirect_to game_path(@game) if @game.srdc.try(:shortname).present? && params[:game] == @game.id.to_s
   end
+
+  def set_games
+    @games = Hash.new([])
+    SpeedrunDotComGame.order('ASCII(name) ASC').each do |game|
+      @games[game.name[0].downcase] += [game]
```
BatedUrGonnaDie: Purely out of curiosity, why do you append an array with the game as its only element?

glacials: `+=` isn't technically an append, it creates a new array with the two values concatenated. Normally I'd append with `<<`, but using `+=` avoids the issue I talked with you about earlier:
```ruby
> h = Hash.new
=> {}
> h.default = []
=> []
> h['a'] << 'alice'
=> ["alice"]
> h
=> {}
> h.default
=> ["alice"]
> h['a']
=> ["alice"]
```
Using `+=` is more expensive because it copies the array, but avoids the above issue. It also requires the right-hand side be another array:
```ruby
> h['b'] += 'bob'
TypeError: no implicit conversion of String into Array
> h['b'] += ['bob']
=> ["alice", "bob"]
> h['b']
=> ["alice", "bob"]
```
However in writing this I realized we're doing this array copy for every game, so I'm changing this to be less expensive at the cost of an extra check.

app/javascript/search.js
glacials: Ok! That was a fantastic explanation, thanks for that! I think I've fixed this now, let me know if this change solves this issue.

BatedUrGonnaDie: Looks good now :+1:

app/controllers/games_controller.rb
BatedUrGonnaDie: Ah, that makes sense.  I poked around and found you can set the default proc of the hash to return an array which would allow you to do the first way.
```ruby
h = Hash.new { |hash, key| hash[key] = [] }
h['a'] << 'alice'
```

Not sure if that is cleaner than just doing the check as you are now.

PR #541, opened 2019-04-10
Add dropdown to show previous runs in pb tables

BatedUrGonnaDie:
Switch is only on the profile page for a user, and allows anyone to toggle it.  Wasn't sure if it should go there or on the homepage and only allow people to see their own archived runs.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -110,4 +110,16 @@ class User < ApplicationRecord
   def likes?(run)
     RunLike.find_by(user: self, run: run)
   end
+
+  # Retrieve all runs scoped for a paginated page excluding already loaded runs
+  def old_runs_for_page(paged_runs)
+    ids = []
+    categories = []
+    paged_runs.map do |run|
+      ids << run.id
+      categories << run.category_id
+    end
+
+    runs.where(category_id: categories).where.not(id: ids)
+  end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -63,15 +63,11 @@ class User < ApplicationRecord
         .union_all(runs.by_category(nil))
   end
 
-  def non_pbs
-    ids = []
-    categories = []
-    pbs.map do |run|
-      ids << run.id
-      categories << run.category_id
-    end
-
-    runs.where(category_id: categories).where.not(id: ids)
+  # Takes array of category_ids, will limit returned runs to only those from those categories
+  def non_pbs(categories = [])
+    query = runs.where.not(id: pbs).where.not(category_id: nil)
+    query = query.where(category_id: categories) if categories.present?
+    query
   end
 
   def runs?(category)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -63,7 +63,8 @@ class User < ApplicationRecord
         .union_all(runs.by_category(nil))
   end
 
-  # Takes array of category_ids, will limit returned runs to only those from those categories
+  # Will filter returned runs to only those from the supplied categories if provided
+  # Accepts either an ActiveRecord::Relation of categories or a PORO Array of category ids
   def non_pbs(categories = [])
     query = runs.where.not(id: pbs).where.not(category_id: nil)
     query = query.where(category_id: categories) if categories.present?
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -63,6 +63,17 @@ class User < ApplicationRecord
         .union_all(runs.by_category(nil))
   end
 
+  def non_pbs
+    ids = []
+    categories = []
+    pbs.map do |run|
+      ids << run.id
+      categories << run.category_id
+    end
+
+    runs.where(category_id: categories).where.not(id: ids)
+  end
+
   def runs?(category)
     runs.where(category: category).any?
   end
@@ -109,17 +120,5 @@ class User < ApplicationRecord
 
   def likes?(run)
     RunLike.find_by(user: self, run: run)
-  end
-
-  # Retrieve all runs scoped for a paginated page excluding already loaded runs
-  def old_runs_for_page(paged_runs)
-    ids = []
-    categories = []
-    paged_runs.map do |run|
-      ids << run.id
-      categories << run.category_id
-    end
-
-    runs.where(category_id: categories).where.not(id: ids)
   end
 end
--- a/app/views/shared/_run_table.slim
+++ b/app/views/shared/_run_table.slim
@@ -3,7 +3,7 @@
 / This allows us to blindly check the keys for items instead of wraping in `try`s
 - col_options = cols.to_h { |col| [col, []] }.merge(local_assigns.fetch(:col_options, {}))
 - runs = order_runs(runs).page(params[:page]).includes(:user, :game)
-- non_pbs_runs = user.non_pbs.group_by(&:category_id) if col_options[:time].include?(:archived)
+- non_pbs_runs = user.non_pbs(runs.map(&:category_id)).group_by(&:category_id) if col_options[:time].include?(:archived)
 - unless runs.any?
   center
     i No runs matched!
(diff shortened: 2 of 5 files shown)


BatedUrGonnaDie:
I really liked your mock up so I ended up just implementing that XD.  I definitely agree that this is a better user experience since it doesn't hide the option anymore on page that no one really goes to atm.  I still like having the history showing on the profile page because it's cool to see other's improvements throughout time and I wouldn't want to restrict it to just seeing your own histories.  Plus it gives people another reason to head to someone's profile.

app/views/shared/_run_table.slim
```diff
@@ -1,6 +1,7 @@
 - description ||= nil
 - cols ||= [:runner, :time, :name, :rival, :uploaded]
 - runs = order_runs(runs).page(params[:page]).includes(:user, :game)
+- all_runs = current_user.runs.order(:realtime_duration_ms) if cols.include?(:archived)
```
glacials: if the run table is paginated, we are still grabbing all pages' worth of past runs with this line when we only need the current page's.

app/views/shared/_run_table.slim
```diff
@@ -29,7 +30,17 @@
             - if cols.include?(:runner)
               td.align-right = user_badge(run.user)
             - if cols.include?(:time)
-              td.align-right = run.duration.format
+              td.align-right
+                - if cols.include?(:archived) && run.category_id.present?
+                  - old_runs = all_runs.filter { |r| r.id != run.id && r.category_id == run.category_id }
```
glacials: Is the first part of the filter, where we avoid displaying the run itself, moveable to the `all_runs` query up above? Seems like these runs are getting pulled from the database into memory twice, and the second time is being ignored.

app/views/shared/_run_table.slim
```diff
@@ -29,7 +30,17 @@
...
+                  - if old_runs.length.positive?
+                    span.dropdown-toggle.mr-2 id="#{run.id36}-history-dropdown" data-toggle='dropdown'
+                      = icon('fas', 'history')
+                    span.dropdown-menu.bg-dark aria-labelledby="#{run.id36}-history-dropdown"
+                      h6.dropdown-header Past Times
+                      - old_runs.each do |old_run|
+                        a.dropdown-item.text-light href="#{run_path(old_run)}" = old_run.duration.format
```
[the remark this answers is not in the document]
BatedUrGonnaDie: Nope!  Had in my mind that I needed something else in the link and added it like that.

BatedUrGonnaDie:
I went with adding a new options hash for columns, which will allow us in the future to stuff other column options in it as well.  I also set it so that all existing columns will get an empty array so that we can blindly check if an array has an option because wrapping it in a double `try` was looking really awful and made stuff really confusing.

I also changed the sorting from in db to creating a new array since this allows us to respect the timing of each individual run.

app/models/user.rb
```diff
@@ -110,4 +110,16 @@ def admin?
   def likes?(run)
     RunLike.find_by(user: self, run: run)
   end
+
+  # Retrieve all runs scoped for a paginated page excluding already loaded runs
+  def old_runs_for_page(paged_runs)
```
glacials: This function doesn't know anything about pages, so I think its name / comment / argument should also not discuss them at all. It takes an array of runs as input and spits out an array of runs as output; doesn't matter if the input array was generated by a pagination function or not.

app/views/shared/_run_table.slim
```diff
@@ -29,7 +32,20 @@
             - if cols.include?(:runner)
               td.align-right = user_badge(run.user)
             - if cols.include?(:time)
-              td.align-right = run.duration.format
+              td.align-right
+                / Make sure to not group runs with no category
+                - if col_options[:time].include?(:archived) && run.category_id.present?
+                  / Gather only runs with the same category and sort by default timing ascending
+                  - old_runs = all_runs.filter { |r| r.category_id == run.category_id }
+                  - old_runs = old_runs.sort_by(&Run.duration_type(run.default_timing))
```
glacials: Now that we have `User#old_runs_for_page`, if we have that return a hash rather than an array, we can migrate some of this logic out of the view (and avoid the n `filter` calls as a bonus).

app/models/user.rb
BatedUrGonnaDie: I'm not sure of a good name that describes what it does that isn't a million words long.  How does `runs_for_categories` sound and then have the comment describe that it takes an array of runs and gets all runs by the user for those categories?

glacials: Here are some ideas:
```ruby
user.previous_uploads_for(runs)
user.run_ancestors(runs)
user.runs_before(runs)

# If you're okay with changing the signature:
user.non_pbs(runs.map(&:category))
runs.ancestors # ancestors is a scope on Run
runs.previous_uploads
```
Whichever way you go yeah definitely explain what it returns in the comment, it's a complicated enough method.

BatedUrGonnaDie: Decided to just go with the `non_pbs` and make it the inverse of `pbs`, and just group by in the view so that the method itself is fairly simple in the model.

app/models/user.rb
```diff
@@ -63,6 +63,17 @@ def pbs
         .union_all(runs.by_category(nil))
   end
 
+  def non_pbs
+    ids = []
+    categories = []
+    pbs.map do |run|
```
glacials: This `map` can be an `each` if we're not going to use the result.

app/models/user.rb
```diff
@@ -63,6 +63,17 @@ def pbs
...
+      ids << run.id
+      categories << run.category_id
```
glacials: Do we need to harvest the categories here? `User#pbs` includes every non-nil category, so can we accomplish the same effect with `runs.where.not(category: nil)`?

app/models/user.rb
```diff
@@ -63,6 +63,17 @@ def pbs
...
+    end
+
+    runs.where(category_id: categories).where.not(id: ids)
```
glacials: Long shot, but any simple way to select non-PBs using just SQL?

BatedUrGonnaDie: Do you mean to remove the harvest steps before this line?

app/models/user.rb
BatedUrGonnaDie: In my simplifying I forgot the reason we harvest the categories, which is so that we don't load every run and just what is scoped.

BatedUrGonnaDie:
Ok, I believe this satisfies your all in sql request, keeps the loading only to those runs which are currently on the page, and simplifies the retrieval quite a bit.  Let me know what you think.

app/models/user.rb
```diff
@@ -63,6 +63,13 @@ def pbs
         .union_all(runs.by_category(nil))
   end
 
+  # Takes array of category_ids, will limit returned runs to only those from those categories
+  def non_pbs(categories = [])
+    query = runs.where.not(id: pbs).where.not(category_id: nil)
+    query = query.where(category_id: categories) if categories.present?
+    query
+  end
```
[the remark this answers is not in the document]
BatedUrGonnaDie: Do you want it to be able to take both?  Could have it accept either and then just convert the categories objects to ids before running the query.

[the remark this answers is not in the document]
glacials: Accepting just IDs is fine for now, no need to complicate things without a need. Unless ActiveRecord smartly combines queries when you give it an ActiveRecord::Relation as the value for `categories:`.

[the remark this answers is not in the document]
BatedUrGonnaDie: Rails smartly combined the query if you pass in an `ActiveRecord::Relation` of category objects.  Tested it out by passing both the `ActiveRecord::Relation` of `Category` objects and seeing if it was equal to the result if it was a PORO array of ids and they were.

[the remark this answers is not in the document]
glacials: Sweet, sounds good to me! In that case I retract my comment.

Issue #546, opened 2019-04-24
Boxes slowly eat up disk


glacials:
We just had the site go down because the production box ran out of disk space! Manifested as a 500 error that said
```
An unhandled lowlevel error occurred. The application logs may have details.
```
on a blank white page. The last deploy to production was 7 days ago and deploys replace boxes entirely, so that's probably how long it takes to have disk space creep up to full. We should find out what's growing (logs? cache? other?) and automatically prune it once in a while so disk space used is stable.

BatedUrGonnaDie:
Rack mini profiler writes to files by default, which may be a cause.  I'll try switching it to redis to see if that helps.

BatedUrGonnaDie:
Closed by #554

PR #547, opened 2019-04-24
Update gems to fix revision not found, use redis store for profiler

BatedUrGonnaDie:
Hopefully this will fix the production boxes running out of file system storage (#546 ), or help prevent some of it.


BatedUrGonnaDie:
Yep!  The dedicated clipboard JS file covers this page as well as the settings page, but I tested it just to make extra sure.

PR #550, opened 2019-05-07
Add onsite racing

BatedUrGonnaDie:
Adds the ability to race with your friends natively onsite, all within a browser or any tool which hooks into the new websocket api that this will be adding.

Current expected features that will land in version 1:
- Public races that anyone that has an account will be able to join.  ~These races can be a regular category race, a bingo race, a randomizer race, or any of those 3 with additional notes attached to it.~  General purpose races that can be attached to games, categories, or left to nothing if it doesn't fit.  The use of a notes field is used to differentiate between bingos or randomizers or any other race that doesn't fit nicely.
- ~Patrons~ Everyone will have the ability to create invite only and secret races.  Invite only races are going to be listed publicly so other users may watch, whereas secret races will only be viewable by those with a unique token.
- All races will have a chat room which will allow racers to easily, which will automatically close 30 minutes after the race has finished.
- Disconnection prevention which will allow your time to be saved and then uploaded whenever connection is re-established in the case of an internet outage.
- Races will be based on the honor system.
- Realtime splitting information.
- Ghosts that you can add to race against.

The idea is to get a solid base starting off, and then if it is popular in future iterations to add realtime splitting (added) and statistics (basic overview stats already implemented) on each race page so that you can have realtime information on where all racers are currently in the game.

EDIT: As we've worked on this some features have changed above.


app/models/concerns/raceable.rb
```diff
@@ -0,0 +1,71 @@
...
+    def finished?
+      entrants.all? { |entrant| entrant.finished? || entrant.forfeited? }
+    end
+
+    def entrant_for_user(user)
+      entrants.find_by(user_id: user.id)
+    end
+
+    def visibility!(type)
+      # When given an invalid type, the column defaults will make the race public
```
glacials: imo invalid types should error, otherwise we might write some code that slips through the cracks into production because it mostly looks like it works. Even if it's just an uncaught exception that bubbles up to cause a 500, that's better than a 200 that does the wrong thing.

db/migrate/20190425121602_add_racing.rb
```diff
@@ -0,0 +1,81 @@
+class AddRacing < ActiveRecord::Migration[6.0]
+  def change
+    create_table :standard_races do |t|
+      t.belongs_to :category, foreign_key: true, null: false
+      t.belongs_to :user,     foreign_key: true, null: false
+
+      t.string   :status_text, null: false
+      t.boolean  :listed,      null: false, default: true
+      t.boolean  :invite,      null: false, default: false
```
glacials: Any specific reason for making race publicity two booleans (rather than one enum)?

imo the default way of representing stuff in our database should be equivalent to how we represent it to users, unless there's a good argument not to.

Specifically here, I feel two fields is harder to reason about and needs special methods (the ones on Raceable) to abstract away the difficulty. They have four possible states but we only use three, so it's confusing to figure out which three are valid -- `invite` is in this weird undefined / unused state when `listed` is false. The fields are almost always written together and read together, so they're mostly behaving as if they're one field.

If we ever decide to add a fourth type of publicity (say, "open to my Twitch followers") there's no guarantee it will slot into this system, so we'd need to add another field for that. If we use an enum up front, we can add as many states as we want to later without making things more complicated.

app/channels/api/v1/global_race_channel.rb
```diff
@@ -0,0 +1,68 @@
...
+    race.notes = data['notes'] if data['notes'].present?
+    if race.save
+      transmit({
+        type: 'race_creation_successful',
+        data: {
+          message: 'Race has been created',
+          race:    Api::V4::RaceBlueprint.render_as_hash(race),
+          path:    Rails.application.routes.url_helpers.race_path(*race.url_params)
+        }
+      })
```
glacials: Is it possible / reasonable to extract this generic outer message for transmits to a blueprinter template as well? Having these hash literals sitting here among all this race logic makes me nervous something will accidentally get rustled, like a single character gets accidentally changed, and all race clients immediately break completely.

On a related note, I think we should have some guarantee that `type` is something we expect and not something one of us has typo'd. I don't know if that fits into blueprinter or not, but even something like a
```
MESSAGE_TYPES = %w[race_creation_successful ...].freeze
```
or whatever with a message-generator method that checks against it would help me sleep at night.

db/migrate/20190425121602_add_racing.rb
BatedUrGonnaDie: I had only briefly seen enums mentioned before in the rails docs, but that seems much better.  I'll take a look at changing that over instead of having a bunch of boolean fields.

app/channels/api/v1/global_race_channel.rb
```diff
@@ -0,0 +1,68 @@
...
+    when StandardRace
+      return if data['category_id'].blank?
+
+      category = Category.find_by(id: data['category_id'])
+      return if category.nil?
+
+      race.category = category
+    when BingoRace
+      return if data['game_id'].blank?
+      return if data['bingo_card'].blank?
```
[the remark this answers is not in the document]
BatedUrGonnaDie: Do you mean just attempting to put the params in no matter what, and then have the validator fail on create/save or whatever and then send back those error messages?

app/channels/api/v1/global_race_channel.rb
BatedUrGonnaDie: Yeah, I definitely want a sole location that defines all the types that can be sent, and then a way to make sure that any type of message sent back is in that array/list/enum.  I'm not sure if blueprinter is the correct solution for that since I haven't looked into it for non-instances of AR, but I'm sure something out there exists for this kind of thing.

Speaking along stability lines, do you think its worth rescuing any methods in the channels that send back some sort of status in the event that something errors?  Just like a basic hash with a type of something like `fatal_server_error` so that the client knows that something horrible went wrong and that it's request just isn't taking ages?

app/channels/api/v1/global_race_channel.rb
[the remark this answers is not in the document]
glacials: Yeah, exactly

app/channels/api/v1/global_race_channel.rb
glacials: Yeah, if the client expects a response that's a good idea.

BatedUrGonnaDie:
Chat design looks pretty cool, but I do have 2 concerns with it as is (not sure if you've already though of these).

The first being that if there is a lot of chat messages the screen will just stretch endlessly down.  Maybe some sort of overflow after X view height or something?

The other being displaying race information in the chat box pushing messages down out of sight.  I like both styles of having it and not, just I can see a situation where people are trying to talk and you have others joining and readying just shoving messages all over the place which could make it hard to follow a conversation.  Maybe a toggle for information or something like that?

app/models/concerns/raceable.rb
BatedUrGonnaDie: For stuff that can accept user input, is returning `nil` acceptable if it doesn't find anything that matches?  For internal only things I agree an error is the way to go.

app/channels/api/v1/global_race_channel.rb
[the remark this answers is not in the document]
BatedUrGonnaDie: Yeah this made the channel look so much nicer.

app/channels/api/v1/global_race_channel.rb
BatedUrGonnaDie: Let me know what you think of the current way I handled messages, and if it works for you then I'll make the rescue's do the same thing essentially as what the normal messages have.

glacials:
Both great points, I'll address them.

Also I just remembered, we should add a scope for API consumers to request to do race stuff. Right now the only scope is for uploading runs.

BatedUrGonnaDie:
I added a generic websocket sign in scope, but having a better name is something that would probably be for the best.

glacials:
Ah, hadn't gotten that far through the review yet I guess 😅

app/channels/api/v1/races_channel.rb
```diff
@@ -0,0 +1,228 @@
...
+    if entrant.destroy
+      transmit_user('race_leave_success', 'Race successfully left')
+      broadcast_race_update('race_entrants_updated', 'An entrant has left the race')
+    else
+      transmit_user('race_leave_error', entrant.error.full_messages.to_sentence)
+    end
+  end
+
+  def ready
+    return if current_user.nil?
```
glacials: Same as above

app/channels/api/v1/races_channel.rb
```diff
@@ -0,0 +1,228 @@
...
+  end
+
+  def unready
+    return if current_user.nil?
+
+    update_race_instance
+    if @race.started?
+      transmit_user('race_started_error', 'Cannot unready for race in progress')
+      return
+    end
```
glacials: This could be in a validation

app/channels/api/v1/races_channel.rb
```diff
@@ -0,0 +1,228 @@
...
+      transmit_user('race_ready_success', 'Entrant ready successful')
+      broadcast_race_update('race_entrants_updated', 'An entrant has readied up')
+      maybe_start_race
+    else
+      transmit_user('race_ready_error', entrant.errors.full_messages.to_sentence)
+    end
+  end
+
+  def unready
+    return if current_user.nil?
```
glacials: Same

app/channels/api/v1/races_channel.rb
```diff
@@ -0,0 +1,228 @@
...
+    # This is to try and make sure ff's have the most accurate time
+    forfeit_time = Time.now.utc
+    forfeit_time = Time.at(data['server_time']).utc if data['server_time'].present?
+    return if current_user.blank?
+
+    update_race_instance
+    unless @race.started?
+      transmit_user('race_started_error', 'Cannot abandon race that is not in progress')
+      return
+    end
```
glacials: Validation

app/channels/api/v1/races_channel.rb
```diff
@@ -0,0 +1,228 @@
...
+
+    if @race.finished?
+      transmit_user('race_finished_error', 'Cannot finish in a race that is already finished')
+      return
+    end
```
glacials: Validation

app/channels/api/v1/races_channel.rb
```diff
@@ -0,0 +1,228 @@
...
+    # This is to try and make sure done's have the most accurate time
+    done_time = Time.now.utc
+    done_time = Time.at(data['server_time']).utc if data['server_time'].present?
+    return if current_user.nil?
+
+    update_race_instance
+    unless @race.started?
+      transmit_user('race_started_error', 'Cannot finish in a race that is not started')
+      return
+    end
```
glacials: Validation

app/channels/api/v1/races_channel.rb
```diff
@@ -0,0 +1,228 @@
...
+
+    if @race.finished?
+      transmit_user('race_finished_error', 'Cannot finish in a race that is already finished')
+      return
+    end
```
glacials: Validation

app/channels/api/v1/races_channel.rb
```diff
@@ -0,0 +1,228 @@
...
+  end
+
+  def rejoin
+    return if current_user.nil?
+
+    update_race_instance
+    unless @race.started?
+      transmit_user('race_started_error', 'Cannot rejoin race that is not in progress')
+      return
+    end
```
glacials: Validation

app/channels/api/v1/races_channel.rb
```diff
@@ -0,0 +1,228 @@
...
+
+    if @race.finished?
+      transmit_user('race_finished_error', 'Cannot rejoin race that has already finished')
+      return
+    end
```
glacials: Validation

app/channels/application_cable/connection.rb
```diff
@@ -1,4 +1,36 @@
 module ApplicationCable
   class Connection < ActionCable::Connection::Base
+    identified_by :current_user, :onsite
+
+    def connect
+      # Identify if the user is 'onsite' so that pre-rendered html can be sent
+      self.current_user, self.onsite = find_verified_user
+      log_tag = current_user.try(:name) || SecureRandom.hex
```
glacials: For convention's sake, I've been using `SecureRandom.uuid` when needing a random identifier like this.

app/controllers/races_controller.rb
```diff
@@ -0,0 +1,29 @@
+class RacesController < ApplicationController
+  before_action :set_race, only: [:show]
+  before_action :set_race_gon, only: [:show]
+
+  def show
+    render template: 'races/show'
```
glacials: Is this line necessary? `races/show` should be the default render for `RacesController#show`

app/javascript/token.js
```diff
@@ -33,7 +38,7 @@ document.addEventListener('turbolinks:load', () => {
   const iframe = document.createElement('iframe')
   const params = {
     response_type: 'token',
-    scope: 'upload_run',
+    scope: 'upload_run+websocket_sign_in',
```
glacials: We talked about this offline but as a reminder -- let's change this to be a scope for `conduct_race` or similar.

app/models/chat_room.rb
```diff
@@ -0,0 +1,4 @@
+class ChatRoom < ApplicationRecord
+  belongs_to :race
+  has_many :chat_messages, dependent: :destroy
+end
```
glacials: What's the purpose of having this be its own model, rather than setting `belongs_to :race` on `ChatMessage`?

app/models/entrant.rb
```diff
@@ -0,0 +1,33 @@
...
+  def forfeited?
+    forfeited_at.present?
+  end
+
+  def done?
+    finished? || forfeited?
+  end
+
+  def place
+    return '-' unless finished?
```
glacials: Daydreaming about the future when we can fill this in even when not `finished?` due to having splits in races 😍😍

app/models/entrant.rb
```diff
@@ -0,0 +1,33 @@
...
+    return 'big fat x' if forfeited?
+
+    raceable.entrants.order(finished_at: :asc).pluck(:id).index(id) + 1
+  end
+
+  def duration
+    return nil unless finished?
```
glacials: In the past week or so I've merged some commits that apply a small twist on the [null object pattern][1] to Duration -- if you use a `Duration.new(nil)` rather than `nil`, it will mostly behave smartly when it's compared / rendered / used in arithmetic with other Durations.

[1]: https://medium.com/@franzejr/null-object-pattern-in-ruby-12a809bc75cc

app/models/concerns/raceable.rb
```diff
@@ -0,0 +1,62 @@
...
+    def started?
+      started_at.present?
+    end
+
+    def in_progress?
+      started? && !finished?
+    end
+
+    def finished?
+      entrants.all? { |entrant| entrant.finished? || entrant.forfeited? }
```
glacials: Is there a way to accomplish this in a query, without selecting all entrants into memory first?

app/models/user.rb
```diff
@@ -116,6 +121,15 @@ def likes?(run)
     RunLike.find_by(user: self, run: run)
   end
 
+  # races will returns a standard ruby array, NOT an AR relation
+  def races
+    standard_races + bingo_races + randomizer_races
+  end
+
+  def in_race?
+    !races.all?(&:finished?)
```
glacials: Can this be done in a query without loading all past races?

app/blueprints/api/v4/race_blueprint.rb
```diff
@@ -0,0 +1,25 @@
...
+    field :card do |race, _|
+      race.started? ? race.card : nil
+    end
+
+    association :game, blueprint: Api::V4::GameBlueprint
+  end
+
+  view :randomizer do
+    field :seed do |race, _|
+      race.started? ? race.seed : nil
```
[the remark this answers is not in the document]
BatedUrGonnaDie: As far as I'm aware, you withhold the seed/card so that people cannot start routing or planning before the race starts.  I'm not sure if there is a subset that would rather share it as bingos and randomizers were never really my thing.

Since our system is more focused on the honor system it would probably be fine have it revealed from race creation though.  We can always add in later some kind of time system that the card is revealed X seconds from start of race if that is preferable.  Would you prefer that it is shown from the beginning then?

app/models/chat_room.rb
BatedUrGonnaDie: I use chat room to hold a locked at time stamp so that you can only send messages upwards of X minutes after the end of a race.  Is that behavior desirable or something not really needed?

app/models/concerns/raceable.rb
BatedUrGonnaDie: We can probably query based on the status text, but that implies that we only ever have a few fixed status texts.  Another solution would be to have a finished at or closed at field directly on the race.  If you wanted the closed chat behavior i talked about in the other thing this could also be based on that finished at timestamp and could remove the chat room object if you still wanted the locking behavior.

app/controllers/races_controller.rb
BatedUrGonnaDie: Artifact of when it was 3 different controllers.  Based on the other discussion we had, are we going to make this 3 different actions, one for `race`, `bingo`, and `randomizer` respectively?

app/models/entrant.rb
BatedUrGonnaDie: I think I had half of those changes when I branched off, but having that built in support for `nil` will be much nicer to work with.

app/models/user.rb
BatedUrGonnaDie: See [this](https://github.com/glacials/splits-io/pull/550#discussion_r283105684)

app/channels/api/v1/global_race_channel.rb
```diff
@@ -0,0 +1,55 @@
...
+  def unsubscribed
+    stop_all_streams
+  end
+
+  def create_race(data)
+    return if current_user.blank?
+    return if data['race_type'].blank?
+
+    race_type = Raceable.race_from_type(data['race_type'])
+    return if race_type.nil?
```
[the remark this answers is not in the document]
BatedUrGonnaDie: Agreed, as well as the rescue from each method that should return a response in the event that we error in some part of the app we don't expect.

app/channels/api/v1/global_race_channel.rb
```diff
@@ -0,0 +1,55 @@
...
+    when RandomizerRace
+      game = Game.find_by(id: data['game_id'])
+
+      race.game = game
+      race.seed = data['seed']
+    end
+
+    race.owner = current_user
+    race.status_text = Raceable::OPEN_ENTRY
+    race.visibility = data['visibility'] if race.class.visibilities.key?(data['visibility'])
```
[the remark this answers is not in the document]
BatedUrGonnaDie: Since we defined it as an enum, rails throws an error when setting this value to anything not defined in our enum hash.  This was the workaround so that errors aren't being raised since they're quite slow.

app/blueprints/api/v4/race_blueprint.rb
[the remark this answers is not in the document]
glacials: Yeah I say let's first do the honor system then adjust if needed.

app/controllers/races_controller.rb
glacials: Yeah I do like the idea of passing the race type into the controller in a more structured way than a string in the params hash. I was thinking three different actions initially, like `race_show`, `bingo_show`, `randomizer_show`, but actually what would you think of a setup like this?

```ruby
# config/routes.rb
get '/races/:race',       to: 'races/races#show',       as: :race
get '/bingos/:race',      to: 'races/bingos#show',      as: :bingo
get '/randomizers/:race', to: 'races/randomizers#show', as: :randomizer

# app/controllers/races/application_controller.rb
class Races::ApplicationController < ApplicationController
  before_action :set_race,     only: [:show]
  before_action :set_race_gon, only: [:show]

  def show
  end

  private

  def set_race_gon
    # ...
  end
end

# app/controllers/races/races_controller.rb
class Races::RacesController < Races::ApplicationController
  def show
    render 'races/show'
  end

  private

  def set_race
    @race = Race.find(params[:id])
  end
end

# app/controllers/races/bingos_controller.rb
class Races::BingosController < Races::ApplicationController
  def show
    render 'races/show'
  end

  private

  def set_race
    @race = Bingo.find(params[:id])
  end
end

# app/controllers/races/randomizers_controller.rb
class Races::RandomizersController < Races::ApplicationController
  def show
    render 'races/show'
  end

  private

  def set_race
    @race = Randomizer.find(params[:id])
  end
end
```
This subfolder-based ApplicationController setup is how we do APIs and our nested non-API runs controllers (`runs/exports`, `runs/likes`).

I don't know if I caught what your three controllers were like before they were merged; did you initially try something like the above then find something bad about it? Or was it more copypasta than this?

app/models/chat_room.rb
glacials: I like that idea -- can we enforce it through logic rather than data? i.e. if the race's entrants are all finished or forfeited, you can only chat if the difference between the current time and the most recent finished_at/forfeited_at time <= X?

BatedUrGonnaDie: Yep, but you should probably take a look at [this](https://github.com/glacials/splits-io/pull/550#discussion_r283105684) reply in regards to figuring out if the race is finished.

app/controllers/races_controller.rb
BatedUrGonnaDie: It was almost identical but wasn't namespaced which seemed like it was just cluttering everything up and didn't inherit from a shared parent like that.  I think those 2 things would solve the style issues I was having with it while being a lot more consistent with how you would expect things to work.

app/models/concerns/raceable.rb
glacials: Not sure I follow your first sentence -- I mean can we do something like
```ruby
!Entrant.where(race: race, finished_at: nil).or(Entrant.where(race: race, forfeited_at: nil)).any?
```

BatedUrGonnaDie: Ah, thought you meant without querying entrants for some reason, not sure why though.  Does this looks alright?

```
    def finished?
      started? && Entrant.where(raceable: self, finished_at: nil, forfeited_at: nil).none?
    end
```

I tried to do with by `join`ing on raceable so that it could check `started_at` in the db as well, but that was throwing a polymorphic preload error.

glacials: Looks good to me! 1-2 queries is better than n. Also we may be able to clean the syntax a tiny bit by replacing `Entrant.where(raceable: self...` with `entrants.where(...` (I didn't catch that when I was suggesting it)

jhobz:
Just popping in to leave a general 👀 👌 ❗️ all over this PR. Carry on.

db/migrate/20190425121602_add_racing.rb
```diff
@@ -15,30 +15,30 @@ def change
...
       t.belongs_to :user, foreign_key: true, null: false
 
-      t.string   :status_text, null: false
+      t.integer  :status,      null: false, default: 0
       t.integer  :visibility,  null: false, default: 0
-      t.string   :join_token,  null: false, index: {unique: true}
+      t.string   :join_token,  null: false
       t.string   :notes,       null: true
-      t.string   :card,        null: false
+      t.string   :card_url,    null: true
```
BatedUrGonnaDie: Is there a reason to allow `card_url` and `seed` to be null?

glacials: I thought I remembered us talking about making card and seed optional, but now that I think about it I might be remembering us talking about making them exposed to race entrants from the beginning. I'll revert.

jhobz:
This obviously might not be necessary in the first PR on this feature, but in case it's one of those things that'll be easier to include early on, I'm mentioning it now: including bingo cards and/or a random seed name generator (for randos) could be really useful. Dream scenario would be to even [i]provide[/i] the seed file once all parties have joined (obviously tons of third-party blockers for that, but one can dream).

glacials:
@jhobz Haha I literally was just asking bated about this offline just before you posted this. Bingo cards are a bigger hurdle for later, but I want to learn more about randomizer seeds because I've never run a game that had randomizers. If we wanted to generate a random seed, what would it look like? Is the format different between games/programs? Or can we just generate any random string and expect it to be valid in most communities?

jhobz:
@glacials I'll throw you a chat over on Discord and we can summarize the results here for posterity. Real-time might be useful in this regard.

jamacanbacn:
Do you think it will be possible to set this up through some kind of websocket?
One of the great things about the SRL livesplit component is when you hit your final split it automatically .done's in chat so you don't have to rush to any page to finish.
That could help a lot with the "honor system" part if I am understanding that correctly.

👀 Excited for this let me know if you need any help testing it out (will maybe try pulling your code in at some point)

glacials:
Yep, WebSockets was a given for us! It's already built into this code.

app/javascript/count.js
```diff
@@ -1,7 +1,15 @@
...
 
-document.addEventListener('turbolinks:load', () => {
-  Array.from(document.querySelectorAll('[data-ms]')).forEach(el => {
-    window.setTimeout(() => el.dataset.ms += resolutionMS, resolutionMS)
-  })
+// We attach to load instead of turbolinks:load because this function re-finds DOM elements every tick, so we can re-use
+// the same function. We choose to re-find DOM elements every tick because ActionCable can replace our element with a
+// new one rendered server-side at any moment.
+window.addEventListener('load', () => {
+  setInterval(() => {
```
BatedUrGonnaDie: How accurate do we want the counting time to be?  `setInterval` doesn't guarantee the function be ran every X milliseconds, just that it will be run around that interval length.  When I was doing SRR stuff I found that even as high has 200-300ms intervals you can see the seconds be of variable length, and closer to 1000 I saw it actually skip numbers before.

app/models/concerns/raceable.rb
```diff
@@ -15,8 +15,14 @@ module Raceable
 
     # Return all current races, as defined by any raceable less than an hour old or has 2 or more entrants
     def self.current
-      where('updated_at > ?', 1.hour.ago)
-        .or(where(id: Entrant.having('count(*) > 1').group(:raceable_id).select(:raceable_id)))
+      case name # self.name refers to the class including this concern
```
BatedUrGonnaDie: Was this doing the thing where it doesn't know which table's column to compare against?

app/views/shared/_run_table.slim
```diff
@@ -38,11 +38,13 @@
...
-                  .text-warning data={ms: (now - race.created_at) * 1000}
-                    = Duration.new((now - race.created_at) * 1000).format
+                  .text-warning data={ms: (now - race.started_at) * 1000}
+                    = Duration.new((now - race.started_at) * 1000).format
+                - elsif race.finished?
+                  .text-success Finished
                 - else
                   - if race.public_visibility?
-                    .text-success Public
+                    .text-success Open
```
BatedUrGonnaDie: My concern with using `Open` for the visibility is that users might think that a race that is actually in progress is still join-able.  Do you think that that it's clear enough to not be a worry?

glacials: Yeah I agree, there is an `if` above this that catches in-progress and finished races, so this only says "Open" if the race is public entry *and* not yet started.

app/models/concerns/raceable.rb
glacials: Yeah, `updated_at` was an ambiguous column reference when I tried to do `@game.races.current`.

app/javascript/count.js
glacials: Oh wow, TIL! 🌠 How did you solve this in SRR? Should I implement something like this? https://medium.com/@swainson/accurate-javascript-timers-f71e0af5df32

app/javascript/channels/api/v1/race_channel.js
```diff
@@ -0,0 +1,270 @@
...
+
+      const doneTime = localStorage.getItem(`race-${window.gon.race.id}-done-time`)
+      if (doneTime) {
+        this.done(parseFloat(doneTime))
+        localStorage.removeItem(`race-${window.gon.race.id}-done-time`)
+      }
+    },
+
+    disconnected() {
+      isConnceted = false
```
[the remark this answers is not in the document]
BatedUrGonnaDie: My ability to typo words in unparalleled.

app/views/shared/_run_table.slim
BatedUrGonnaDie: Woops, must've missed that.

app/javascript/count.js
BatedUrGonnaDie: With the race start time as a data attribute I would use the timesync instance to calculate a duration.  That way it didn't rely on the timer being accurate since it was working with static times.  That link seems like it would accomplish things as well just in a different way, so whatever you think is easiest.

BatedUrGonnaDie:
@jamacanbacn to clarify the "honor system" part of the original message is that there will be no moderation in the event that someone cheats (or something of that nature).  We don't have a mod team or anything that can handle support for things like that at this time.

app/javascript/count.js
CryZe: I'd argue you just want 1s to be an integer multiple of the interval you choose here. And it needs to be lower than 1s to avoid "aliasing" / skipping. So 0.5s, 0.25s or 0.2s (avoid the last one due to floating point inaccuracies maybe) should be fine.

app/javascript/channels/api/v1/race_channel.js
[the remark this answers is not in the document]
glacials: Wow I just realized it's not just this occurrence, you typod every single occurrence in the exact same way including the declaration so this is Correct Code™ 😆

app/javascript/count.js
glacials: Okay just to spell this out for my own understanding, it seems like there are two issues with setInterval:

**Precision**: an interval of `n` ms might trigger after `n +/- c` ms instead. At its best this makes the timer tick `c` ms early/late, which may or may not be noticeable by a human. At its worst this can cause skipping, e.g. `00:00:00` ticks first at 999 ms, then at 1002 ms, which jumps the timer from `00:00:00` to `00:00:02`.

**Drift**: Bad precision over long periods of time (and/or other issues?) can make the timer drift from the actual elapsed time. e.g. 4 hours in your timer might say `04:32:10` when the actual elapsed time is `04:31:56`. Even just a 1ms error every second for 4 hours adds up to 14 seconds off.
 
Having 1s be a multiple of our interval of is a partway solution from what I can see. It avoids skipping, but it's still prone to drift and the first issue with precision -- the length of a second can vary by an amount *up to the interval duration*.

```
|--one real second--|
|-------------------|-------------------|
:   :   :   :   :   :   :   :   :   :   : <- where ticks should happen (200ms)
^   ^   ^   ^   ^  ^    ^   ^   ^   ^   ^ <- where ticks do happen (200ms +/- c)
\-----------------------/
 this "second" is 1200ms
                        \---------------/
                        this one is 800ms
```

The weird part is that the smaller the tick interval, the *higher the chance* of this happening. A real-world second that elapses in the mid-point between two ticks is relatively safe from falling off the edge of a tick due to variance. As the gap between ticks shrinks, the size of that "safe zone" shrinks. But the tick interval is still the upper bound for how off a second can be, so smaller is still probably better.

`timesync` solves drift, but I'm not sure whether it solves precision. It seems like the length of a displayed second might still vary due to setInterval.

I just found [dbkaplun/driftless][1] which recursively re-schedules ticks the closer they get to occurring to be more accurate. I'll mess around with this and `timesync` and with the function I linked before and see what works. `timesync` seems 👌👌 tho.

[1]: https://github.com/dbkaplun/driftless

BatedUrGonnaDie: @CryZe would having it be a multiple really help though since it's going to not only take time to execute the function, but could be interrupted by something else taking over causing it to not be ran precisely every multiple of the second?

@glacials Yeah, timesync solves the drift problem, but not the precision problem.  Driftless seems pretty cool, and it could solve the precision problem as long as stuff isn't happening earlier than it should be otherwise we're going to have the exact same problem where things just miss the new second and causes the next second to become "longer".

app/javascript/channels/api/v1/race_channel.js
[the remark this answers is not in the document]
BatedUrGonnaDie: I must've just autocompleted the wrong wrong spelling, I feel like there's no way that I could misspell something that many times in one try (although I type `destory` instead of `destroy` more than I would like to admit).

package.json
```diff
@@ -3,6 +3,7 @@
   "private": true,
   "dependencies": {
     "@rails/actioncable": "6.0.0-rc1",
+    "@rails/activestorage": "^6.0.0-alpha",
```
BatedUrGonnaDie: This should be changed to `6.0.0-rc1`.  For some reason on npm it shows the latest release as alpha but if you go to all versions then there is a pre verion with rc1.

app/models/randomizer.rb
```diff
@@ -2,6 +2,7 @@ class Randomizer < ApplicationRecord
   include Raceable
 
   belongs_to :game
+  has_many_attached :files
```
BatedUrGonnaDie: How come you went with `has_many` and not `has_one`?  (I know nothing about active storage)

app/views/races/_new.slim
```diff
@@ -58,9 +58,8 @@
           .tab-pane.text-light#randomizer-content role='tabpanel' aria={labelledby: 'randomizer-content-tab'}
             form#randomizer-races-form
               .form-group
-                label for='randomizer-seed-input' Seed
-                input.form-control#randomizer-seed-input type='text' placeholder='Seed' required=true
-                .valid-feedback Looks good!
-                .invalid-feedback Required.
+                label for='randomizer-seed-input' Seed name (optional)
+                input.form-control#randomizer-seed-input type='text' placeholder='Seed name'
+                small.form-text.text-muted You'll be able to attach a seed file once the race is created.
```
BatedUrGonnaDie: How come you went with attaching after creation instead of requiring it to proceed and having it be unchanging?

app/models/randomizer.rb
glacials: Mostly because there's really no downside to `has_many`, and I figured some people/games might need to share multiple files (though I don't know of any). If there's a downside I haven't thought of, open to changing.

app/views/races/_new.slim
glacials: This is one of my philosophies about building things (see: splitsdb vs splits.io). i.e. don't require any upfront work to get going, and always let people change their mind.

Specific to this scenario, I was thinking people would feel "I want to do a randomizer!", then get people to join them, and actually generating a seed file would be the last thing on their mind. So I don't want to put a hurdle up in front of people that says "people can't join you until you finish your chores".

I also want to allow people to change their mind -- maybe they were going to share the file in some other way or not at all, then once everyone has joined the race (perhaps someone they don't know joins) they decide to share it using Splits.io. the alternative would be to force them to recreate the race or settle for some other way of sharing files.

app/models/randomizer.rb
BatedUrGonnaDie: Ah, good idea I didn't even think of needing to share multiple files.

app/views/races/_new.slim
BatedUrGonnaDie: Ok, that does make a lot of sense.  100% agree with the sentiment of creating the race before even thinking about creating the seed file.

I would like to make an argument for having a seed file be present for a randomizer race to start though.  I think being able to go back to a race and be able to download their seed file is a cool concept and having it be there for historical purposes also sounds nice.  Obviously you could upload a blank file or something, but it would encourage sharing the seed file for others to be able to see.

glacials: I agree having the seed file consistently be on Splits.io for historical purposes is a great goal. I agree we should encourage it. I disagree that we should force it.

If we force users to upload a seed file and they don't want to / it doesn't make sense to, they will either choose not to race using Splits.io at all (which we need to avoid early on for racing to catch on through the network effect), or like you said they'll upload some random file to get past the hurdle, which by itself is its own hurdle -- both in the fact that you have to go through the effort to do so, but also because you have to learn that it's possibility, and not that you've hit a brick wall by not having a file to upload.

My biggest worry is that you and I personally don't understand the entirety of the randomizer community, and for one reason or another some communities straight up won't be uploading files at all. Maybe it's the standard that only the seed name is shared and seed files are generated locally, or maybe the randomizer program has its own built-in share feature that's 1000% better than what we can provide, or maybe even the game itself allows you to input the seed, so there's literally no file to be shared at all (see: Crypt of the Necrodancer).

I want to make all races reproducible for historical purposes. Not by forcing users to upload seed files but by making the experience of sharing your seed file through us far better than any other way of doing so; and for communities where sharing a seed file doesn't even make sense, talk to those communities (later, after we deploy this) and figure out if there is a low-friction way to preserve that history if the seed name field isn't enough.

BatedUrGonnaDie: Those are all very compelling arguments.  I can't really disagree with any of it so I retract my desire to force a seed file :+1:

app/channels/api/v1/race_channel.rb
```diff
@@ -0,0 +1,298 @@
...
+      transmit_user('race_rejoin_success', 'Entrant rejoin successful')
+      broadcast_race_update('race_entrants_updated', 'An entrant has rejoined the race')
+    else
+      transmit_user(entrant.error_status! || 'race_rejoin_error', entrant.errors.full_messages.to_sentence)
+    end
+  rescue StandardError => e
+    Rails.logger.error([e.message, *e.backtrace].join($RS))
+    Rollbar.error(e, 'Uncaught error for Api::V1::RaceChannel#rejoin')
+    transmit_user('fatal_error', 'A fatal error occurred while processing your message')
+  end
```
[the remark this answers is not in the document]
BatedUrGonnaDie: You can also come back from a forfeit, so I wasn't sure what to really call it that worked for both.  "Undone" seemed too focused on just if you finish with a time.  Definitely open to changing it though, but naming things is a very weak point for me.

app/models/run.rb
```diff
@@ -207,7 +207,7 @@ def segment_history_stats(timing)
   end
 
   def recommended_comparison(timing)
-    query = Run.where(category: category).where.not(user: nil, category: nil)
+    query = Run.where(category: category).where.not(user: nil).where.not(category: nil)
```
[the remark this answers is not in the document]
BatedUrGonnaDie: Depreciation warnings make me sad :cry:

app/blueprints/api/v1/websocket_message_blueprint.rb
```diff
@@ -0,0 +1,3 @@
+class Api::V1::WebsocketMessageBlueprint < Blueprinter::Base
```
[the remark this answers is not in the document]
BatedUrGonnaDie: I was actually gonna ask about this soon-ish.  We could always just make separate blueprints for each version, but code organization wise things might need to be addressed?  I can see a situation where eventually the WebSocket version and the HTTP version are in the same-ish numbers and sharing folders and that seems like a bad idea if they aren't changing together.

app/blueprints/api/v4/chat_message_blueprint.rb
```diff
@@ -0,0 +1,5 @@
+class Api::V4::ChatMessageBlueprint < Blueprinter::Base
+  fields :body, :entrant
```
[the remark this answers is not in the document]
BatedUrGonnaDie: None at all!  Slipped my mind, but it would be good to have this so that people can put timestamps on the messages if they so desire.

app/blueprints/api/v4/race_blueprint.rb
```diff
@@ -0,0 +1,26 @@
+class Api::V4::RaceBlueprint < Blueprinter::Base
+  fields :id, :status, :visibility, :notes, :started_at, :created_at, :updated_at
+
+  field :type do |race, _|
```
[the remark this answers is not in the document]
BatedUrGonnaDie: Oh interesting.  I always just went with the `_` because rubocop yells at you to do that if you have an unused block parameter.  I'll give that a shot.

app/javascript/channels/api/v1/race_channel.js
```diff
@@ -0,0 +1,263 @@
...
+        this.forfeit(parseFloat(forfeitTime))
+        localStorage.removeItem(`race-${window.gon.race.id}-forfeit-time`)
+      }
+
+      const doneTime = localStorage.getItem(`race-${window.gon.race.id}-done-time`)
+      if (doneTime) {
+        this.done(parseFloat(doneTime))
+        localStorage.removeItem(`race-${window.gon.race.id}-done-time`)
+      }
+    },
```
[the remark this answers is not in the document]
BatedUrGonnaDie: This is a super cool idea!  It should matter if the event exists though, since we could just emit our own even upon connection in our channel right?  Then we just listen for our own custom event and handle it accordingly yeah?

app/controllers/api/v4/time_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Api::V4::TimeController < Api::V4::ApplicationController
+  def create
+    render status: :ok, json: {
+      status: :ok,
+      id:     params[:id],
```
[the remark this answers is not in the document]
BatedUrGonnaDie: [Timesync](https://github.com/enmasseio/timesync#protocol) wants them to match up the response to it's request.

app/channels/application_cable/connection.rb
```diff
@@ -1,4 +1,39 @@
...
+    end
+
+    def oauth_token=(token)
+      @oauth_token = token
+    end
+
+    protected
+
+    def find_verified_user
+      onsite = ['splits.io', 'localhost'].include?(request.host)
```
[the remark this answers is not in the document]
BatedUrGonnaDie: You were right it didn't work for a beta, but I think I found a better solution.  `onsite = ['splits.io', 'localhost'].include?(request.domain(1))`.  This will take the first 2 items from the right, so the `splits.io` part, or if there isn't that many the entire thing (`localhost`).  This way it doesn't matter if we ever change the subdomain or add another or something like that.

BatedUrGonnaDie:
Sounds good to me to make it more clear for Entrants.  I did kind of the same thing with how races belong to people by having them be called "owners" over needing to know the difference between `race.user` and `race.users`.

I think it's a good idea to allow changing bingo as well.  Like you said for randomizers you most likely would create the bingo before you have a card ready to go, so being able to update it later would be pretty beneficial.

app/channels/api/v1/race_channel.rb
```diff
@@ -0,0 +1,298 @@
...
+  # Ends the race if all entrants have either finished or forfeited, otherwise does nothing
+  def maybe_end_race
+    update_race_instance
+    return if !@race.started? || !@race.entrants.all?(&:done?)
+
+    @race.update(status: :ended)
+    broadcast_race_update('race_ended', 'All entrants have finished')
+    GlobalRaceUpdateJob.perform_later('race_finished', 'A race has finished', @race)
+  end
+end
```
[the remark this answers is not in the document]
BatedUrGonnaDie: I don't know of any "accepted" solution to this.  I haven't really seen anything that wasn't a toy/trivial example of using action cable.  I think we kind of have free range to try things out and experiment.  I just think it might be beneficial to avoid having a ton of channels that need to be subscribed to if that is going to increase the complexity for a 3rd party dev.  Maybe we could leverage models more or something along those lines?

Related, but why are large files considered always bad?  I feel like having everything related to something all in one place would be a good thing since you wouldn't have to try and figure out a chain of callbacks or something along those lines

app/controllers/races/randomizers_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    @race.files.attach(params[:randomizer][:files])
+    Api::V1::RaceChannel.broadcast_to(@race, Api::V1::WebsocketMessageBlueprint.render_as_hash(
+      Api::V1::WebsocketMessage.new('new_attachment', {
+        message: 'The race owner has attached a new file',
+        race: Api::V4::RaceBlueprint.render_as_hash(@race, view: @race.type),
+        attachments_html: ApplicationController.render(partial: 'races/attachments', locals: {race: @race})
+      })
+    ))
+  end
```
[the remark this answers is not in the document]
BatedUrGonnaDie: My idea had been to put everything related to races over the websocket just so that it was clear where to go for it.  I also wanted to force 3rd parties to use the websockets to make an integration so that the integration would be better user experience, if that makes sense?

I'm having a bit of hard time trying to put my thoughts into text that is somewhat coherent.  If you want I can dump my thoughts in a call.

app/javascript/channels/api/v1/race_channel.js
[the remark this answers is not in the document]
glacials: Yeah I think that's right! I've never emitted an event in JS though so I'm not 100%.

app/blueprints/api/v4/race_blueprint.rb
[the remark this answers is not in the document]
glacials: Ohh ok, I didn't realize Rubocop yelled about it. Feel free to disregard what I said then.

app/channels/api/v1/race_channel.rb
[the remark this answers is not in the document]
glacials: That's a really good question! There are some practical disadvantages, like readability.
```ruby
class ThingsController < ApplicationController
  def index
  end

  def create
  end

  def show
  end
end
```
This controller clearly tells you what it does -- list, create, and show Things. But if it grows to 400 lines, it doesn't. You have to scroll a lot, have a good memory, good tooling, etc.

And there's a theoretical disadvantage -- large files can be indicative of *other* bad things.

e.g. Run started small but eventually grew a lot of ways to render itself. If we'd started Splits.io today and knew we wanted 19 ways of rendering a Run, we'd build the rendering engine to be independent from the start. But we didn't know that, we slowly grew it from nothing, and it's hard to see how complex it is when you're in the weeds.

But what you *can* see is the file size. When it starts getting long, it's a good idea to dive in a bit and see if you've overlooked whether a new model/controller/etc. wants to be split out. For us, we had controller actions for `download`, `run_history_csv`, and `segment_history_csv` bloating their controllers and kind of all doing the same thing -- exporting a run.

The weirdest part about this type of situation is none of us even realized how similar they were in UX -- the download button for the different timer types was completely separate from the "export to CSV" ones.

By forcing ourselves to think about whether there was a new controller begging to be created, we came up with Run Export, and alongside that came one button to handle all types of exports.
```
app/
  controllers/
    runs_controller.rb
    runs/
      exports_controller.rb
```
It also gave us a nice birds-eye view in the directory structure, just like we have in simple controllers. The code got cleaner and the user experience did as well.

I do agree channels will be a different beast, as splitting up one channel into two causes a very different experience for consumers of the channel, where in controllers it does not.

I do notice some repeated code in similar methods, e.g. ready/unready are basically identical. Could we share code between these methods more, or even merge them into one `set_ready(ready)`-esque method? How about for other entrant state-change type methods?

app/blueprints/api/v1/websocket_message_blueprint.rb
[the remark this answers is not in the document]
glacials: Yeah let's keep it simple and change them together, sorry for my original recommendation which turned out to be bad

app/channels/api/v1/race_channel.rb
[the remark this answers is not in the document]
BatedUrGonnaDie: Thanks for the info.  I can definitely see now why having concise files would make things better in the long run.

My only concern with combining the methods (which may be baseless) was that I wanted each action to be explicit.  You knew what you were doing by what method was being called.  Originally in SRR I had leaving a race before it started and forfeiting be the same method, and that seemed like it would be confusing to consumers.  You had to track the state of the race to know what the method was going to do.  I probably took that a little too far though here, since by passing it a value to some of the methods you maintain that explicit nature without needing to redefine everything.

app/controllers/races/bingos_controller.rb
```diff
@@ -6,6 +6,6 @@ def show
   private
 
   def set_race
-    @race = Bingo.find(params[:race])
+    @race = Bingo.where('LEFT(id::text, ?) = ?', race_params[:race].length, race_params[:race]).order(created_at: :asc).first
```
BatedUrGonnaDie: Might be a good idea to move this to some shared method so that each controller can use the same lookup.

app/channels/api/v1/race_channel.rb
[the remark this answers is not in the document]
glacials: Yeah I totally agree we shouldn't use one "command" to mean different things based on state, like a `toggle_ready` or the example you mentioned. The consumer should know exactly what effect its command will have if successful, even if it is out of sync with the server state.

The current setup is great as an API design, but I do think we need to find a way to share some of the duplicated code, however that is. Like you said we should feel free to invent something.

config/routes.rb
```diff
@@ -160,13 +160,16 @@
...
-      get '/randomizers/:race', to: 'races/randomizers#show'
+      Raceable::RACE_TYPES.map { |raceable| raceable.type.to_s.pluralize }.each do |type|
+        get   "/#{type}",       to: "races/#{type}#index"
+        post  "/#{type}",       to: "races/#{type}#create"
+        get   "/#{type}/:race", to: "races/#{type}#show"
+        patch "/#{type}/:race", to: "races/#{type}#update"
+
+        post   "/#{type}/:race/entrants",          to: "races/entrants/#{type}#create"
+        patch  "/#{type}/:race/entrants/:entrant", to: "races/entrants/#{type}#update"
+        delete "/#{type}/:race/entrants/:entrant", to: "races/entrants/#{type}#destroy"
```
glacials: Thoughts on doing PATCH/DELETE `/#{type}/:race/entrant` instead, and inferring the entrant based on auth? Can't see us ever needing to have one account manage multiple entrants in the same race, and this would make it simpler for consumers (including our frontend), and less prone to auth bugs on the backend.

spec/support/views/api/v4/races/show.json
```diff
@@ -0,0 +1,12 @@
...
+  "required": [
+    "race"
+  ],
+  "additionalProperties": false,
+  "properties": {
+    "race": {
+      "$ref": "../../../../models/api/v4/race.json"
+    }
+  }
+}
```
glacials: wew schemas 👍

config/routes.rb
BatedUrGonnaDie: I actually like it more.  I thought about it but it seemed less REST-ful but if you don't think that's a problem I'm more than willing to head that direction.  Inferring the entrant also means clients don't have to track their ID's, just manage the state that they're currently in.

app/controllers/api/v4/races/entrants/application_controller.rb
```diff
@@ -0,0 +1,97 @@
...
+
+      @column = param if params[param].present?
+    end
+    return unless @column.nil?
+
+    render status: :bad_request, json: {
+      status: :bad_request,
+      error:  "Specify at least one parameter to update: #{columns}"
+    }
+  end
```
glacials: Why the restriction on updating only one at a time?

BatedUrGonnaDie: In order to keep it explicit exactly what is getting updated, and because updating 2 things at once doesn't really make sense (how can you both forfeit and finish at the same time).

BatedUrGonnaDie:
There is some kind of JS error being thrown from vue when logged out.  I think the cause is something to do with gon.

glacials:
Okay I can take a look.

glacials:
I'm getting all these while logged out (yellow one also while logged in). No cause found yet but unrelated I oddly seem to get re-logged-in after navigating around races, lmao. I think when we call `current_user=` after seeing an OAuth token, it's also setting a cookie.

BatedUrGonnaDie:
I didn't realize calling `current_user=` would overwrite the session stuff.  It's getting called in the entrants controller and probably the main race controller.  We could just switch it to @user then or something along those lines.

glacials:
^ This commit diverts `current_user`/`current_user=` uses in the API to its own variable which is scoped only to that request.

app/controllers/api/v4/races/application_controller.rb
```diff
@@ -5,14 +5,36 @@ class Api::V4::Races::ApplicationController < Api::V4::ApplicationController
...
+
+  def create
+  end
+
+  def show
+  end
+
+  def update
+  end
+
```
glacials: Oops sorry for removing these, didn't realize they were for Rubocop.

BatedUrGonnaDie: I didn't either until you removed it.  I just have a linter plugin for sublime that runs on every save so I try and stay on top of it in files I touch.

glacials:
I've added the ability for consumers to associate a run with a race entry, for runs to exist without being complete, for third party consumers to update them as splits happen, and for all of the above to display nicely on the site both in the context of a run and of a race.

The most recent commit has some failing tests which I will fix later, just felt the need to push this before being afk for a bit.

Here are some screenshots!

Screen Shot 2019-06-19 at 13 07 02
Screen Shot 2019-06-19 at 13 07 07
Screen Shot 2019-06-19 at 13 57 01

app/controllers/api/v4/application_controller.rb
```diff
@@ -88,7 +88,7 @@ def validate_user
   end
 
   def set_race
-    @race = Race.find(params[:race])
+    @race = Race.friendly_find(params[:race])
```
BatedUrGonnaDie: I thought we were going to have the API take full ID's only?

app/models/race.rb
```diff
@@ -51,7 +51,9 @@ def self.active
   end
 
   def self.friendly_find(slug)
-    where('LEFT(id::text, ?) = ?', slug.length, slug).order(created_at: :asc).first
+    race = where('LEFT(id::text, ?) = ?', slug.length, slug).order(created_at: :asc).first
+    raise ActiveRecord::RecordNotFound if race.nil?
```
BatedUrGonnaDie: Since this method now raises on not finding, should it follow Rails find convention and be a bang method?  (ex: `find_by!`)

app/models/race.rb
```diff
@@ -91,6 +93,20 @@ def maybe_start!
     return if started? || entries.where(readied_at: nil).any? || entries.count < 2
 
     update(started_at: Time.now.utc + 20.seconds)
+
+    # Schedule ghost splits and finishes
+    entries.ghosts.find_each do |entry|
+      entry.update(finished_at: started_at + (entry.run.duration(Run::REAL).to_sec))
+      Api::V4::RaceBroadcastJob.set(wait_until: started_at + entry.run.duration(Run::REAL).to_sec).perform_later(
+        self, 'race_entries_updated', 'A ghost has finished'
```
BatedUrGonnaDie: I feel like ghosts should have their own status, and not just be entries updating since it isn't a true entrant.  That way 3rd parties can ignore the ghosts if they choose too.  What do you think?

config/routes.rb
```diff
@@ -166,14 +166,16 @@
       get   '/races/:race', to: 'races#show',   as: 'race'
       patch '/races/:race', to: 'races#update'
 
-      get    '/races/:race/entry', to: 'races/entries#show'
-      put    '/races/:race/entry', to: 'races/entries#create'
-      patch  '/races/:race/entry', to: 'races/entries#update'
-      delete '/races/:race/entry', to: 'races/entries#destroy'
+      get    '/races/:race/entries/:id', to: 'races/entries#show',   as: 'race_entry'
```
BatedUrGonnaDie: How come we're switching back to ID based?

glacials: Since a ghost is an Entry, that means one user can own multiple entries. So consumers need some way of differentiating one from another, which means we can't do `/races/:race/entry` anymore.

app/controllers/api/v4/application_controller.rb
glacials: Oh oops, I don't think I realized this was an API controller. I'll revert.

app/models/race.rb
glacials: Yeah I was torn here because bang -> raise is the overall Rails convention, but [`find`](https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find) breaks that convention by raising, and I saw this method as closely related to `find`.

Probably bang is better though, so there's no question about it.

app/models/race.rb
glacials: I actually like the idea of ghosts not needing any special code client-side, except maybe a styling change if `entry.ghost == true`. But I'm open to hear any practical advantages of separating them!

For disadvantages: if ghosts had a separate status, that's another event clients have to manually listen for (and might opt not to) and make display changes based on. But ultimately ghosts should be mixed into the list of real runners, so then you basically end up listening for 2 events that have the same effect (re-render the list of runners).

However I do think the extreme version of this is worth considering, where ghosts don't send any events at all, and are just static fields that point to the run. The clients would look at the times in the run and time their display changes correctly so that the splits and finish times happen when they are supposed to.

I think that extreme is a cleaner concept and has a nice offline-friendly mode, but still requires a lot more work from clients to ultimately display something that could have just behaved as a runner the whole time, so not worth it overall.

app/validators/entry_validator.rb
```diff
@@ -1,87 +1,74 @@
...
-      # Reject if `finished_at` is already set
-      if record.forfeited_at.present? && record.finished_at.present?
-        record.errors[:forfeited_at] << 'Cannot forfeit a race you have already finished'
-      end
+    # Reject both finished_at and forfeited_at being set
+    if record.finished_at.present? && record.forfeited_at.present?
+      record.errors[:base] << 'Cannot finish and forfeit from the same race'
     end
   end
 end
```
glacials: 👏 Fantastic refactor

app/models/race.rb
BatedUrGonnaDie: I never really understood why all the `find_by` methods returned nil on not finding, but `find` would raise.  Seems like an easy thing to have be standardized.

app/models/race.rb
BatedUrGonnaDie: I feel like going the extreme route where they just get the run is not very "integrated".  As a 3rd party consumer I would feel like it's something that is just tacked on so I don't think that is a great way to go personally, but I'm sure someone else may disagree.

I don't really have any clear cut advantages to having them be separate, it's just a sort of gut feeling I have.

BatedUrGonnaDie:
As a note for later, whenever this is merged we should have another commit that then removes all the run broadcast since upload jobs so they don't just sit in the database until the end of time.

config/application.rb
```diff
@@ -24,5 +24,9 @@ class Application < Rails::Application
     config.to_prepare do
       Doorkeeper::AuthorizationsController.layout 'application'
     end
+
+    config.action_cable.disable_request_forgery_protection = true
+    config.action_cable.url = '/api/cable'
```
BatedUrGonnaDie: Just wanna make sure this is the URL we want to mount the cable at.

config/initializers/content_security_policy.rb
```diff
@@ -16,7 +16,16 @@
 
 #   # Specify URI for violation reports
 #   # policy.report_uri "/csp-violation-report-endpoint"
-# end
+
+  # from https://github.com/rails/webpacker#vue
+  Rails.application.config.content_security_policy do |policy|
```
BatedUrGonnaDie: We're already in the security policy, so this do wrapper can be removed.

db/schema.rb
```diff
@@ -169,6 +215,21 @@
...
+  create_table "races", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.bigint "category_id"
+    t.bigint "user_id", null: false
+    t.integer "visibility", default: 0, null: false
+    t.string "join_token", null: false
+    t.string "notes"
+    t.datetime "started_at", precision: 3
+    t.datetime "created_at", precision: 3, null: false
+    t.datetime "updated_at", precision: 3, null: false
+    t.bigint "game_id"
```
BatedUrGonnaDie: We should make this be a foreign key as well.

db/schema.rb
```diff
@@ -73,6 +103,22 @@
     t.index ["priority", "run_at"], name: "delayed_jobs_priority"
   end
 
+  create_table "entries", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.uuid "race_id"
```
BatedUrGonnaDie: We should make this be a foreign key as well.  And maybe an index?

db/schema.rb
```diff
@@ -58,6 +78,16 @@
     t.index ["name"], name: "index_categories_on_name"
   end
 
+  create_table "chat_messages", id: :uuid, default: -> { "gen_random_uuid()" }, force: :cascade do |t|
+    t.uuid "race_id"
```
BatedUrGonnaDie: We should make this be a foreign key as well.

BatedUrGonnaDie: We should make this be a foreign key as well.  And maybe an index?

app/channels/application_cable/channel.rb
```diff
@@ -1,4 +1,6 @@
 module ApplicationCable
   class Channel < ActionCable::Channel::Base
+    delegate :oauth_token, :oauth_token=, to: :connection
```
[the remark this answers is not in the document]
glacials: Tried removing these and got this error whenever a race update was attempted.
```
web_1        | There was an exception - NoMethodError(undefined method `oauth_token=' for #<ApplicationCable::Connection:0x00007f97866cfd70>)
web_1        | /app/app/channels/application_cable/connection.rb:27:in `find_verified_user'
web_1        | /app/app/channels/application_cable/connection.rb:7:in `connect'
web_1        | /usr/local/bundle/gems/actioncable-6.0.0.rc1/lib/action_cable/connection/base.rb:171:in `handle_open'
web_1        | /usr/local/bundle/gems/actioncable-6.0.0.rc1/lib/action_cable/server/worker.rb:59:in `block in invoke'
web_1        | /usr/local/bundle/gems/actioncable-6.0.0.rc1/lib/action_cable/server/worker.rb:41:in `block in work'
web_1        | /usr/local/bundle/gems/activesupport-6.0.0.rc1/lib/active_support/callbacks.rb:112:in `block in run_callbacks'
web_1        | /usr/local/bundle/gems/actioncable-6.0.0.rc1/lib/action_cable/server/worker/active_record_connection_management.rb:16:in `block in with_database_connections'
web_1        | /usr/local/bundle/gems/actioncable-6.0.0.rc1/lib/action_cable/connection/tagged_logger_proxy.rb:24:in `block in tag'
web_1        | /usr/local/bundle/gems/activesupport-6.0.0.rc1/lib/active_support/tagged_logging.rb:80:in `block in tagged'
web_1        | /usr/local/bundle/gems/activesupport-6.0.0.rc1/lib/active_support/tagged_logging.rb:28:in `tagged'
web_1        | /usr/local/bundle/gems/activesupport-6.0.0.rc1/lib/active_support/tagged_logging.rb:80:in `tagged'
web_1        | /usr/local/bundle/gems/actioncable-6.0.0.rc1/lib/action_cable/connection/tagged_logger_proxy.rb:24:in `tag'
web_1        | /usr/local/bundle/gems/actioncable-6.0.0.rc1/lib/action_cable/server/worker/active_record_connection_management.rb:16:in `with_database_connections'
web_1        | /usr/local/bundle/gems/activesupport-6.0.0.rc1/lib/active_support/callbacks.rb:121:in `block in run_callbacks'
web_1        | /usr/local/bundle/gems/actioncable-6.0.0.rc1/lib/action_cable/engine.rb:62:in `block (4 levels) in <class:Engine>'
web_1        | /usr/local/bundle/gems/activesupport-6.0.0.rc1/lib/active_support/execution_wrapper.rb:88:in `wrap'
web_1        | /usr/local/bundle/gems/actioncable-6.0.0.rc1/lib/action_cable/engine.rb:57:in `block (3 levels) in <class:Engine>'
web_1        | /usr/local/bundle/gems/activesupport-6.0.0.rc1/lib/active_support/callbacks.rb:121:in `instance_exec'
web_1        | /usr/local/bundle/gems/activesupport-6.0.0.rc1/lib/active_support/callbacks.rb:121:in `block in run_callbacks'
web_1        | /usr/local/bundle/gems/activesupport-6.0.0.rc1/lib/active_support/callbacks.rb:139:in `run_callbacks'
web_1        | /usr/local/bundle/gems/actioncable-6.0.0.rc1/lib/action_cable/server/worker.rb:40:in `work'
web_1        | /usr/local/bundle/gems/actioncable-6.0.0.rc1/lib/action_cable/server/worker.rb:58:in `invoke'
web_1        | /usr/local/bundle/gems/actioncable-6.0.0.rc1/lib/action_cable/server/worker.rb:53:in `block in async_invoke'
web_1        | /usr/local/bundle/gems/concurrent-ruby-1.1.5/lib/concurrent/executor/ruby_thread_pool_executor.rb:348:in `run_task'
web_1        | /usr/local/bundle/gems/concurrent-ruby-1.1.5/lib/concurrent/executor/ruby_thread_pool_executor.rb:337:in `block (3 levels) in create_worker'
web_1        | /usr/local/bundle/gems/concurrent-ruby-1.1.5/lib/concurrent/executor/ruby_thread_pool_executor.rb:320:in `loop'
web_1        | /usr/local/bundle/gems/concurrent-ruby-1.1.5/lib/concurrent/executor/ruby_thread_pool_executor.rb:320:in `block (2 levels) in create_worker'
web_1        | /usr/local/bundle/gems/concurrent-ruby-1.1.5/lib/concurrent/executor/ruby_thread_pool_executor.rb:319:in `catch'
web_1        | /usr/local/bundle/gems/concurrent-ruby-1.1.5/lib/concurrent/executor/ruby_thread_pool_executor.rb:319:in `block in create_worker'
```

app/javascript/vue/race.js
```diff
@@ -0,0 +1,136 @@
...
+    this.error = false
+
+    const headers = new Headers()
+    if (localStorage.getItem('splitsio_access_token')) {
+      headers.append('Authorization', `Bearer ${localStorage.getItem('splitsio_access_token')}`)
+    }
+
+    let url = `/api/v4/races/${this.raceId}`
+    let join_token
+    if (window.gon.race) {
```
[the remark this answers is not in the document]
glacials: I think the localstorage solution is better than having the server pass the join token to the client over and over

1. if a join token is in the URL, save it to localstorage
2. if a join token is in localstorage, use it in the connection

app/javascript/vue/race.js
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    this.globalSubscription = consumer.subscriptions.create('Api::V4::GlobalRaceChannel', {
+      connection() {},
+
+      disconnected() {},
+
+      received(data) {
+        switch(data.type) {
+          // TODO: update races on game page with this info
```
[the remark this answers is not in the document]
glacials: yeah i agree, let's not increase scope more. and we can easily add it later.

app/models/concerns/unparsed_run.rb
```diff
@@ -157,5 +157,10 @@ def write_segment_histories(segs)
 
       SegmentHistory.import(histories)
     end
+
+    def zero_to_nil(parsed_number)
+      parsed_number = parsed_number.presence
```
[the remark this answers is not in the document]
glacials: It converts blank values like `''` to `nil` just for added assurance that we are only dealing with numbers or `nil`.

app/models/race.rb
```diff
@@ -0,0 +1,169 @@
...
+  # note: this can potentially send multiple ending messages if called on an already finished race
+  def maybe_end!
+    return if !started? || entries.where(finished_at: nil, forfeited_at: nil).any?
+
+    Api::V4::RaceBroadcastJob.perform_later(self, 'race_ended', 'The race has ended')
+    Api::V4::GlobalRaceUpdateJob.perform_later(self, 'race_ended', 'A race has ended')
+  end
+
+  # Returns this race's entry for the given user, not including ghosts.
+  def entry_for_user(user)
```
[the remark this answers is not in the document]
glacials: I can do you one better, how about we make it a class method on `Entry` so we can call `race.entries.for_user(user)`!

config/application.rb
glacials: No issues here

config/initializers/content_security_policy.rb
glacials: I have no idea why, but removing this causes errors when loading charts on the run page related to this setting.

app/models/race.rb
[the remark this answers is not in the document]
BatedUrGonnaDie: Wait, you can call class methods on relation collections?

glacials:
Yep! A class method is the exact same thing as a scope.

app/models/race.rb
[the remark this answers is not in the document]
glacials: Yep! A class method is the exact same thing as a scope.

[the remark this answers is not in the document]
BatedUrGonnaDie: Dang, never knew about that.  That's super cool.

app/helpers/races_helper.rb
```diff
@@ -0,0 +1,19 @@
...
+      nil
+    else
+      raise 'invalid status'
+    end
+  end
+
+  # Overrides the default helper because Race#to_param returns a user-friendly ID
+  def api_v4_race_path(race)
+    super(race.id)
+  end
```
glacials: this felt super dirty to do, open to better ideas if you have them

app/javascript/tooltips.js
```diff
@@ -1,10 +1,14 @@
...
   tippy('.gold-split', {placement: 'left'})
   tippy('.tip', {placement: 'top'})
   tippy('.tip-top', {placement: 'top'})
   tippy('.tip-bottom', {placement: 'bottom'})
   tippy('.tip-right', {placement: 'right'})
   tippy('.tip-left', {placement: 'left'})
-})
+}
+
+export { applyTips }
```
glacials: what was the reason for this change?

app/controllers/api/v4/races/chat_messages_controller.rb
```diff
@@ -0,0 +1,41 @@
+class Api::V4::Races::ChatMessagesController < Api::V4::ApplicationController
+  before_action :set_user, only: %i[create]
+  before_action :validate_user, only: %i[create]
+  before_action :set_race, only: %i[index create]
+  before_action :set_messages, only: %i[index]
+
+  def index
+    render status: :ok, json: Api::V4::ChatMessageBlueprint.render(paginate(@messages), root: :chat_messages)
+  end
```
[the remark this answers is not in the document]
BatedUrGonnaDie: I don't think we should allow it.  I think all secret stuff should only be available to those that are allowed to be in the race.

app/blueprints/api/v4/race_blueprint.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  field :join_token do |race, options|
+    options[:join_token] ? race.join_token : nil
+  end
+
+  association :owner, blueprint: Api::V4::UserBlueprint
+  association :entries, blueprint: Api::V4::EntryBlueprint
+  association :chat_messages, blueprint: Api::V4::ChatMessageBlueprint do |race, options|
+    options[:chat] ? race.chat_messages.order(created_at: :desc).limit(100) : []
```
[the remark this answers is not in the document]
BatedUrGonnaDie: Yep!  I was able to get rid of one on the normal load by including the twitch and google users, but for some reason I can't get it to work on the api request.

app/javascript/tooltips.js
BatedUrGonnaDie: To apply the tips when we receive new html from ActionCable (user badges mostly).

app/helpers/races_helper.rb
BatedUrGonnaDie: The straightforward answer would be to either have the API take friendly URLs as well, or just forgo them completely.

config/initializers/content_security_policy.rb
BatedUrGonnaDie: I found the exact same behavior and it makes no sense, so we'll just leave it as is I guess.

app/helpers/races_helper.rb
glacials: yeah, but
1. i hate super long and unintelligible user-facing urls
2. i regret ever making api resources accessible by anything but id (like games and runners) and want to move away from it long-term

so i think this might be the best of bad options :/

app/controllers/api/v4/races/chat_messages_controller.rb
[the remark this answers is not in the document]
glacials: ohh ok i didn't realize we were catching it in `set_race`. considering we want to pretend like the race doesn't exist if the user doesn't have access, i think it's right to keep it in `set_race`.

app/blueprints/api/v4/race_blueprint.rb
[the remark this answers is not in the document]
glacials: hmm, yeah i see the n+1 happening. i also can't find the cause of this. also can't reproduce it in a rails console :/

PR #551, opened 2019-05-11
Support refreshing Twitch tokens

glacials:
Add automatic support for refreshing Twitch access tokens when they're expired or otherwise not valid.

See https://dev.twitch.tv/docs/authentication/#refreshing-access-tokens

The Twitch API has always returned refresh tokens, but access tokens never actually expired so this was never critical. I'm adding it for future proofing and to solve bugs where the access token needs refreshing for other reasons (e.g. the user changes their Twitch password).

### DIFF
--- a/app/models/twitch_user.rb
+++ b/app/models/twitch_user.rb
@@ -22,12 +22,13 @@ class TwitchUser < ApplicationRecord
     end
 
     twitch_user.assign_attributes(
-      access_token: auth.credentials.token,
-      name:         auth.info.nickname,
-      display_name: auth.info.name,
-      email:        auth.info.email,
-      avatar:       auth.info.image || TwitchUser.default_avatar,
-      url:          auth.info.urls.Twitch
+      access_token:  auth.credentials.token,
+      refresh_token: auth.credentials.refresh_token,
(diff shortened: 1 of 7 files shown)


app/models/twitch_user.rb
```diff
@@ -39,9 +40,29 @@ def self.default_avatar
 
   def videos
     Twitch::Videos.recent(twitch_id, type: :archive, token: access_token)
+  rescue RestClient::Unauthorized
+    refresh_tokens!
```
BatedUrGonnaDie: If we catch an error here and go to refresh the token, the original intended action is going to not be performed, correct?  Maybe we could check the return value of `refresh_tokens!` and then if it is successful call `retry` in the rescue block so that it process the original request for data?

app/models/twitch_user.rb
```diff
@@ -39,9 +40,29 @@ def self.default_avatar
...
   end
 
   def followed_ids
     Twitch::Follows.followed_ids(twitch_id, token: access_token)
+  rescue RestClient::Unauthorized
+    refresh_tokens!
```
BatedUrGonnaDie: Same thing as above.

PR #574, opened 2019-08-01
Fix comparing values with nil durations

BatedUrGonnaDie:
I believe I was able to keep the logic consistent with what it was before, just want to double check.  Turns out that by `including Comparable` and defining `<=>` in a certain way implements a bunch of comparing methods [automagically](https://ruby-doc.org/core-2.6/Comparable.html).

One behavior now is `Duration.new(0) != Duration.new(nil)`, and I'm not sure how that was handled before.  However `Duration.new(0) == Duration.new(0) && Duration.new(nil) == Duration.new(nil)` which I believe is new behavior?

### DIFF
--- a/app/models/duration.rb
+++ b/app/models/duration.rb
@@ -57,7 +57,7 @@ class Duration
   end
 
   def !=(other)
-    !to_ms == other.to_ms
+    !(self == other)
   end
 
   def <=>(other)
--- a/app/models/duration.rb
+++ b/app/models/duration.rb
@@ -1,4 +1,6 @@
 class Duration
+  include Comparable
+
   def initialize(milliseconds)
     @ms = milliseconds
   end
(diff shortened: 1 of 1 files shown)


app/models/duration.rb
```diff
@@ -50,87 +52,58 @@ def format_casual(num_units: 2, sign: :negatives)
...
-    ms = duration.respond_to?(:to_ms) ? duration.to_ms : duration
-
-    to_ms > ms
+    d
   end
 
-  def >=(duration)
-    self > duration || self == duration
+  def !=(other)
+    !to_ms == other.to_ms
```
glacials: Is this right? Feels like `!to_ms` would always return `true` or `false`.

BatedUrGonnaDie: Nope!  I meant to put `!(self == other)` and leverage the automatically set up `==` method since we know the reverse is just as simple as normal equality.  I'm not sure why it's not covered by Comparable to be honest.

BatedUrGonnaDie:
After fixing `!=` to be correct, I can't notice any pages that aren't acting correctly locally.  I'll merge this and then double check a bunch of compare changes when it gets to the beta push stage.

PR #577, opened 2019-08-05
Allow game lookup by ID in APIv4

### LINKED ISSUE #489: Allow looking up games by ID in API
Allow looking up games in the API by ID, not just shortname.

Since shortnames can change, they're bad ways to keep track of games off-platform. This is worse now that shortnames belong only to third-party game links (Speedrun.com and SpeedRunsLive).

Even on-platform, in the run edit page where you can change your run's game/category, we have to weirdly depend on Speedrun.com shortnames as under-the-hood values for each game even though they are not shown anywhere, because it's the only way we can populate the categories list using the API.

BatedUrGonnaDie:
Closes #489 

Piggybacking 2 other API doc tweaks onto this.

### DIFF
--- a/app/controllers/api/v4/application_controller.rb
+++ b/app/controllers/api/v4/application_controller.rb
@@ -55,11 +55,7 @@ class Api::V4::ApplicationController < ActionController::Base
   end
 
   def set_game
-    if params[:game_value] == 'id'
-      @game = Game.find(params[:game])
-    else
-      @game = Game.joins(:srdc).find_by!(speedrun_dot_com_games: {shortname: params[:game]})
-    end
+    @game = Game.joins(:srdc).find_by!(speedrun_dot_com_games: {shortname: params[:game]})
   rescue ActiveRecord::RecordNotFound
     render not_found(:game)
   end
--- a/app/controllers/api/v4/application_controller.rb
+++ b/app/controllers/api/v4/application_controller.rb
@@ -55,7 +55,11 @@ class Api::V4::ApplicationController < ActionController::Base
   end
 
   def set_game
-    @game = Game.joins(:srdc).find_by!(speedrun_dot_com_games: {shortname: params[:game]})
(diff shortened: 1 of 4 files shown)


app/controllers/api/v4/application_controller.rb
```diff
@@ -55,7 +55,11 @@ def set_category
   end
 
   def set_game
-    @game = Game.joins(:srdc).find_by!(speedrun_dot_com_games: {shortname: params[:game]})
+    if params[:game_value] == 'id'
+      @game = Game.find(params[:game])
+    else
+      @game = Game.joins(:srdc).find_by!(speedrun_dot_com_games: {shortname: params[:game]})
+    end
```
glacials: this is safe and efficient, but i think the interface can be simpler. have you considered something like

1. let `params[:game]` be an ID or a shortname without telling us which, we take the extra 1-query hit and keep the understanding of how to use the API dead simple, and accept the possibility of collisions. or
2. merge the lookup-by-ID function into the `/v4/games?search=:game` endpoint, where if you supply just an ID we return an array containing only that game.

BatedUrGonnaDie: I was really trying to avoid the collision angle because as a developer using it I would not want to pass in what I thought to be a game ID and then get back some other result.  Plus the fact that it might be a valid game ID at first but then a new shortname gets added with the same ID which could then change the result.

I'm not sure I understand what you're proposing for option 2.  Do you mean the flag option to the search endpoint?  Or do you mean accept the collisions there since we'd be returning multiple results so it wouldn't be a problem?  Or am I totally off the mark with both of those.

BatedUrGonnaDie: For posterity, we're going with having the game search check and see if an ID exists with that exact string, and if it does insert the matching record included as part of the search (but include the search results as well to preserve original behavior).

PR #584, opened 2019-08-18
Fix double escaping in window titles

### LINKED ISSUE #570: Apostrophes in game titles aren't properly transformed for window titles
image

BatedUrGonnaDie:
When using `content_for`, the string is automatically escaped and converted to a `SafeBuffer`, whoever when using string concatenation this seems to be ignored, and another round of escaping is performed.

My original fix for this was just just escape the `Game`'s string value, but I feel like we would just hit the same issue down the line with some other resource.  So instead I opted to mark the title as `html_safe`, since the (potentially user) provided titles would already have been escaped, and we control the value of `site_title`.

Closes #570

### DIFF
--- a/app/views/layouts/application.slim
+++ b/app/views/layouts/application.slim
@@ -1,7 +1,7 @@
 doctype html
 html
   head
-    title = content_for?(:title) ? "#{yield(:title)} - #{site_title}".html_safe : site_title
+    title = content_for?(:title) ? yield(:title) + " - #{site_title}" : site_title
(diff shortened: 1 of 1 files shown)


BatedUrGonnaDie:
I agreed that I didn't like the `html_safe` call, so I went back to looking at behavior of `SafeBuffer`.  Turns out that if you manually concatenate to another string (instead of interpolation) then it respects it.  Let me know what you think of this.

PR #585, opened 2019-08-19
Add Race.with_ends

glacials:
Adds a `Race.with_ends` scope similar to `Segment.with_ends` and `SegmentHistory.with_ends`.

This method adds a computed field to the returned models called `ended_at` which contains the end time of the race, inherited from its slowest entry's finish or forfeit time (or null if at least one entry is still going).


glacials:
Didn't even know that was a function! Will do.

PR #586, opened 2019-08-20
Clean up started but abandoned races

glacials:
After a race has been going for 48 hours, forcefully forfeit all entries that have not finished or forfeited already.

The longest recorded 602 time is 47 hours :)

### DIFF
--- a/app/jobs/race_cleanup_job.rb
+++ b/app/jobs/race_cleanup_job.rb
@@ -1,0 +1,10 @@
+class RaceCleanupJob < ApplicationJob
+  # Adding or deleting a job? Reflect the change in the QUEUES environment variable in docker-compose.yml and
+  # docker-compose-production.yml.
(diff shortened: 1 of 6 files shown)


jhobz:
It might be good to have a contingency for the occasional scenario in which one might _actually_ want a race that long (like a big team relay race or something). Like a chatbot warning to which you can respond? (Think Netflix's "are you still watching?") Obviously this leaves it open to abuse, though. That seems like a very low chance of happening to me, but I would understand making a call based on that factor..

glacials:
I like the idea of asking people if they're still active, but the < 1% of cases is not worth the 10x-100x code it would take to implement. If people hit this 48h limit and complain, we can raise the limit. Right now I only care about eventually removing cruft, not necessarily promptly removing it.

jhobz:
Yeah, that's totally valid.

PR #588, opened 2019-08-23
Tweak race display behaviors

BatedUrGonnaDie:
Fixed `finished` scope selecting races with any entrant being done, only count real entrants for abandonment purposes, race table column created -> finished.

### DIFF
--- a/app/models/race.rb
+++ b/app/models/race.rb
@@ -33,14 +33,14 @@ class Race < ApplicationRecord
   end
 
   def self.finished
-    joins(:entries).where.not(entries: {finished_at: nil, forfeited_at: nil}).distinct
+    where.not(id: unfinished)
   end
 
   # unabandoned returns races that have had activity (e.g. creation, new entry, etc.) in the last hour
   # or have more than 2 entries. this includes races that have finished
   def self.unabandoned
     where('races.updated_at > ?', ABANDON_TIME.ago).or(
-      where(id: Entry.having('count(*) > 1').group(:race_id).select(:race_id))
+      where(id: Entry.where(ghost: false).having('count(*) > 1').group(:race_id).select(:race_id))
     )
   end
 
--- a/app/models/race.rb
+++ b/app/models/race.rb
@@ -40,7 +40,7 @@ class Race < ApplicationRecord
   # or have more than 2 entries. this includes races that have finished
   def self.unabandoned
     where('races.updated_at > ?', ABANDON_TIME.ago).or(
-      where(id: Entry.where(ghost: false).having('count(*) > 1').group(:race_id).select(:race_id))
+      where(id: Entry.nonghosts.having('count(*) > 1').group(:race_id).select(:race_id))
     )
   end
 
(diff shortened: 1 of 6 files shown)


app/models/race.rb
```diff
@@ -33,14 +33,14 @@ def self.unfinished
   end
 
   def self.finished
-    joins(:entries).where.not(entries: {finished_at: nil, forfeited_at: nil}).distinct
+    where.not(id: unfinished)
```
glacials: Would it be more efficient to do `with_ends.where(ended_at: nil)`?

app/models/race.rb
```diff
@@ -33,14 +33,14 @@ def self.unfinished
...
   end
 
   # unabandoned returns races that have had activity (e.g. creation, new entry, etc.) in the last hour
   # or have more than 2 entries. this includes races that have finished
   def self.unabandoned
     where('races.updated_at > ?', ABANDON_TIME.ago).or(
-      where(id: Entry.having('count(*) > 1').group(:race_id).select(:race_id))
+      where(id: Entry.where(ghost: false).having('count(*) > 1').group(:race_id).select(:race_id))
```
glacials: Can use `Entry.nonghosts` as syntactic sugar here.

app/views/runs/index.slim
```diff
@@ -27,7 +27,7 @@
     .row.mx-2
       .col-md-6.mb-3: .card.shadow
         = render partial: 'shared/race_table', locals: { \
-          races: current_user.races.finished.order(created_at: :desc), \
+          races: current_user.races.with_ends.finished.order(ended_at: :desc), \
```
glacials: Do you think it would be worth it to move this sorting to inside the partial, so we don't have to repeat it every time we render the partial?

In fact on that note, should we also move the distinction between `races` and `active_races` inside the partial, and just pass in one `races` ActiveRecord::Relation that the partial then splits up using `.finished` and `.unfinished`?

app/models/race.rb
BatedUrGonnaDie: With the way `with_ends` is setup, we can't use it in a where clause.  I tried to get it to work, and was able to by using a from clause but as soon as you chained another scope it just fell apart.  Since this was the only thing I had in mind to use `ended_at` in a where clause, I just decided to go with this basic way of checking instead.

glacials: Ohh ok gotcha so this is where you were having trouble with that

BatedUrGonnaDie: If you want to tinker around with it, something as simple as `Race.select('*').from(Race.with_ends).where('ended_at IS NOT NULL')` was what kind of worked.  I'm not sure if it was because of the other calls to `with_end` or just the other scopes in general that were messing things up.

app/views/runs/index.slim
BatedUrGonnaDie: So I tried tinkering around with this a bunch, and was not able to get things working very well.  If you move just the `order(ended_at: :desc)` to the parial then you start getting column doesn't exist errors, if you move both `.with_ends.order(ended_at: :desc)` to the partial, then you start getting errors about the group by clause.

I also spent some time tinkering around with just passing in one race object, but I wasn't able to get the behavior the same as it was now, specifically on the active races page.  Maybe we can revisit this in the future though.

glacials: Okay sounds good, thanks for explaining!

PR #592, opened 2019-08-25
Fix various API bugs

glacials:
Fixed various API bugs and documentation inconsistencies. (Thanks @CryZe!)

- Add `GET /api/v4/runner` to retrieve the currently authenticated runner
- Move documentation notes on each model's ID type (e.g. UUID, base 36 string) to within that model's section
- Fix docs on `/api/v4/races/:race/entry` -> `/api/v4/races/:race/entries/:entry`
- Fix JSON schema enums being specified wrong
- Fix some missing timers in test files
- Add text for the OAuth grant page for the `manage_race` scope
- Allow `twitch_id`, `twitch_name` to be null now
- Fix race visibility being a number in docs

I have not yet tried to reproduce the bug where an entry's `created_at` is null.

Also for now I'm leaving out the bug where a user with the race page open who changes their entry from the API, causes the website to not know that the change was made, as that's a bit more involved a fix.


app/controllers/api/v4/runners_controller.rb
```diff
@@ -13,4 +14,13 @@ def index
   def show
     render json: Api::V4::UserBlueprint.render(@runner, root: :runner)
   end
+
+  def me
+    if current_user.nil?
+      render not_found(:runner, message: 'No authenticated runner to display')
```
BatedUrGonnaDie: 404 seems a bit weird here, what do you think about a 401?

PR #594, opened 2019-08-30
Finalize rails 6 update

### LINKED ISSUE #517: Rails 6 Compatibility Tracker

BatedUrGonnaDie:
Updates a ton of gems and packages.  Just did blanket updates for both yarn and bundle.

Closes #517 

Also, is it possible to highjack the beta branch to make sure that the cookie changes will not break production?  There was 2 this time and the extra assurance would be nice.


Gemfile
```diff
@@ -45,7 +45,7 @@ group :production do
 end
 
 # administration
-gem 'administrate'
+gem 'administrate', github: 'thoughtbot/administrate'
```
BatedUrGonnaDie: Administrate doesn't have a release that will `bundle install` when using rails 6, but master does work.  They have a plan to release a version sometime soon to fix this.

app/blueprints/api/v3/category_blueprint.rb
```diff
@@ -1,5 +1,5 @@
 class Api::V3::CategoryBlueprint < Blueprinter::Base
   fields :id, :name, :created_at, :updated_at
 
-  association :game, blueprint: Api::V3::GameBlueprint, if: ->(_, options) { options[:toplevel] == :category }
+  association :game, blueprint: Api::V3::GameBlueprint, if: ->(_, _, options) { options[:toplevel] == :category }
```
BatedUrGonnaDie: Blueprinter has a depreciation warning now, because they changed the conditionals to take 3 parameters now instead of 2.

app/controllers/concerns/authenticatable.rb
```diff
@@ -1,8 +1,8 @@
 module Authenticatable
   private
-  
+
   def sign_in(user)
-    self.current_user = user
+    create_auth_session(user)
```
BatedUrGonnaDie: When updating authie, I noticed now that it recommends using this method over calling `self.current_user = user`

app/controllers/runs_controller.rb
```diff
@@ -1,7 +1,5 @@
 require 'uri'
 
-require 'speedrundotcom'
```
BatedUrGonnaDie: Not sure if this was no longer needed because of the new autoloader, or because at some point `/lib` was added to the autoloading path.

app/models/concerns/speedrun_dot_com_run.rb
```diff
@@ -1,8 +1,6 @@
 require 'active_support/concern'
 
-require 'speedrundotcom'
-
-module SRDCRun
+module SpeedrunDotComRun
```
BatedUrGonnaDie: The new autoloader didn't like this old class name, so I updated it to be consistent with all the models that reference speedrun.com stuff.

lib/parser/livesplit_core_parser.rb
```diff
@@ -1,10 +1,10 @@
 begin
-  require File.expand_path('LiveSplitCore', __dir__)
+  require_relative('LiveSplitCore')
 rescue LoadError
-  raise 'Please follow the steps to install livesplit-core in the readme'
+  raise 'LiveSplitCore install not found'
 end
 
-class Parser
+class Parser::LivesplitCoreParser
```
BatedUrGonnaDie: The new autoloader also did not like the way this was defined before.  Since `/lib` is the top level namespace, anything under it needs to be namespaced according to its directory structure.

spec/controllers/api/v4/runs_controller_spec.rb
```diff
@@ -175,7 +175,7 @@
       it 'returns the correct content-type header' do
         request.headers['Accept'] = 'application/wsplit'
 
-        expect(response.content_type).to eq('application/wsplit')
+        expect(response.media_type).to eq('application/wsplit')
```
BatedUrGonnaDie: `response.content_type` now returns the full content type header as is, while `response.media_type` now returns just the MIME component as previous versions of rails did.

BatedUrGonnaDie:
I totally forgot that in development mode it was only loading classes that were accessed in a request.  When I went to test on beta I hit a autoloading error, and this is where I ended up.  Since I had to change a lot of names and removed some stuff, another review seems appropriate.

lib/rails_extensions/read_only.rb
```diff
@@ -1,7 +0,0 @@
-if Rails.configuration.read_only
-  module ActiveRecord::Base
-    def readonly?
```
BatedUrGonnaDie: I couldn't find any reference to this method outside of the definition.  Looks like everything was just checking the environment variable.

lib/parser/livesplit_core_version
```diff
@@ -1 +1 @@
-v0.10.0
\ No newline at end of file
+v0.11.0
```
BatedUrGonnaDie: Since I needed LSC to be in the `Parser::` namespace, I took this opportunity to just update it as well.

lib/programs/exchange_format.rb
```diff
@@ -1,4 +1,4 @@
-module ExchangeFormat
+module Programs::ExchangeFormat
```
BatedUrGonnaDie: Since `lib/` is now a top level namespace, zeitwerk is expecting `programs/` to define a new namespace.  All programs now reside in the namespace, and allows us to easily get all constants of this namespace in `run.rb` instead of needing to manually name all programs there every time we change something.

lib/assets/bower_components/bootstrap-toggle/.bower.json
```diff
@@ -1,41 +0,0 @@
-{
```
BatedUrGonnaDie: I'm pretty positive we don't use bower anymore, and it was causing an issue for some reason (not really sure why).  Seemed like a good time to kill this folder.

config/puma.rb
```diff
@@ -16,6 +16,9 @@
 #
 environment ENV.fetch("RAILS_ENV") { "development" }
 
+# Specifies the `pidfile` that Puma will use.
+pidfile ENV.fetch("PIDFILE") { "tmp/pids/server.pid" }
```
BatedUrGonnaDie: I found this from [Rails diff](railsdiff.com) which is a pretty cool website for checking config changes between rails versions.

config/initializers/inflections.rb
```diff
@@ -14,3 +14,8 @@
 # ActiveSupport::Inflector.inflections(:en) do |inflect|
 #   inflect.acronym 'RESTful'
 # end
+
+ActiveSupport::Inflector.inflections(:en) do |inflect|
+  inflect.acronym('SplitterZ')
```
BatedUrGonnaDie: It didn't make sense to change the file name for these 2 programs to me, since logically they couldn't be considered words.  Instead by making them an inflector zeitwerk will use this definied camelization instead of needing the underscores.  I can do this for any of the other programs if you think that would be a better idea than renaming the files

lib/programs/exchange_format.rb
glacials: 👍 Love it!

config/initializers/inflections.rb
glacials: Can we add a comment to each of these programs' files explaining that? I'm afraid we'll add a program later that needs this, forget about it, and be confused for hours or days about why it doesn't work the same way the others do.

lib/tasks/livesplit_core.rake
```diff
@@ -50,5 +50,10 @@ task :update_lsc do
   mv("#{DEST_FOLDER}/liblivesplit_core.so", "#{PARSER_FOLDER}/liblivesplit_core.so")
   mv("#{DEST_FOLDER}/bindings/LiveSplitCore.rb", "#{PARSER_FOLDER}/LiveSplitCore.rb")
   rm_rf DEST_FOLDER
+  data = File.read("#{PARSER_FOLDER}/LiveSplitCore.rb")
+  data.sub!('module LiveSplitCore', 'module Parser::LiveSplitCore')
+  File.open("#{PARSER_FOLDER}/LiveSplitCore.rb", 'w') do |f|
+    f.write(data)
+  end
```
glacials: Another possible solution to this problem in case you haven't seen it, though I have no opinion either way:

https://stackoverflow.com/a/8762740

config/initializers/inflections.rb
BatedUrGonnaDie: I can totally see that happening as well, will do.

lib/tasks/livesplit_core.rake
BatedUrGonnaDie: Thanks for the link, I ended up liking that way a bit more 👍

PR #595, opened 2019-09-02
Update the docker development environment

BatedUrGonnaDie:
A lot of this PR is based around [this article](https://evilmartians.com/chronicles/ruby-on-whales-docker-for-ruby-rails-development), so I won't repeat most of what it talks about.

Because this new build is using volumes to store all installed gems and npm packages, we should be able to just cache those volumes through some kind of action(?) which would speed test times up immensely as most of the time was spent installing gems.

Fun fact: this started out as my just updating the ruby version to 2.6.4.


Dockerfile
```diff
@@ -1,20 +1,21 @@
-FROM ruby:2.6.0
+FROM ruby:2.6.4-alpine
```
BatedUrGonnaDie: Alpine is a minimal size linux distro.  Prior to this PR, the image size of each image (multiply sizes by 3) was 1938MB.  By using this build process instead using `FROM ruby:2.6.4` it bring it down to 963MB.  By switching to alpine we bring it all the way down to 308MB.

Gemfile
```diff
@@ -29,10 +29,6 @@ group :development, :hot do
   # profiling
   gem 'derailed_benchmarks'
 
-  # tests
-  gem 'guard', require: false
```
BatedUrGonnaDie: Just removing some gems that weren't needed anymore to clean up root directory.

Makefile
```diff
@@ -18,40 +18,40 @@ endif
 container ?= web
 
 build:
-	$(docker-compose) run web bundle install
-	$(docker-compose) run web yarn install
 	$(docker-compose) build
-	$(docker-compose) run web rails db:migrate
+	$(docker-compose) run --rm web bundle install -j $(shell expr $(shell getconf _NPROCESSORS_ONLN) - 1)
```
BatedUrGonnaDie: By adding `--rm` to these commands, after the command exits it will no longer leave an unnamed image floating around taking up space until you prune dangling images.

docker-compose-production.yml
```diff
@@ -62,3 +62,5 @@ services:
     image: worker
     logging: *default-logging
     ports: []
+    volumes:
+      - .:/app
```
BatedUrGonnaDie: Since web has this volume, worker should as well.

docker-compose.yml
```diff
@@ -28,13 +28,21 @@ x-logging:
...
+  - packs:/app/public/packs
+  - rails_cache:/app/tmp/cache
 
 services:
   db:
     image: postgres
     logging: *default-logging
     volumes:
       - ./db/init.sql:/docker-entrypoint-initdb.d/10-init.sql
+      - db:/var/lib/postgresql/data
```
BatedUrGonnaDie: By adding volumes here and in redis, data will now be persisted after a `docker-compose down`.  The way to get rid of the volumes is by doing `docker-compose down --volume`.  Unfortunately I'm not sure how to hook these up to the current containers, so you might need to do a `docker-compose down` first to get them to work.  Once you do that for sure it saves it.

docker-compose.yml
```diff
@@ -90,9 +96,20 @@ services:
     links:
       - s3:s3.localhost
     logging: *default-logging
+    volumes: *default-volumes
     ports: []
     stdin_open: true
     tty: true
   redis:
     image: redis
```
glacials: Same as above re: version numbers; also looks like there's an `-alpine` Redis image too.

docker-compose.yml
```diff
@@ -28,13 +28,21 @@ x-logging:
...
+  &default-volumes
+  - .:/app:cached
+  - bundle:/bundle
+  - node_modules:/app/node_modules
+  - packs:/app/public/packs
+  - rails_cache:/app/tmp/cache
 
 services:
   db:
     image: postgres
```
glacials: Does this default to `:latest`? If so, the article recommends using version numbers.

docker-compose.yml
```diff
@@ -28,13 +28,21 @@ x-logging:
   options:
     max-size: 1m
     max-file: "1"
+x-volumes:
+  &default-volumes
+  - .:/app:cached
+  - bundle:/bundle
+  - node_modules:/app/node_modules
+  - packs:/app/public/packs
+  - rails_cache:/app/tmp/cache
```
glacials: Should we add `:cached` to any other of these? I'm not quite sure how it works but the article recommends adding it on any generated content.

docker-compose.yml
```diff
@@ -90,9 +96,20 @@ services:
...
+    volumes:
+      - redis:/data
+
+volumes:
+  db:
+  redis:
+  bundle:
+  node_modules:
+  packs:
+  rails_cache:
```
glacials: General docker-compose.yml comments:

- The linked article suggests using `tmpfs` for speed, is that something we should do?
- The article also uses the same `image:` for rails, runner, sidekiq, and webpacker (inherited from `app`/`backend`), where we specify a different name for each. Would we get any benefit from combining them like they do?

docker-compose.yml
BatedUrGonnaDie: If you wanted to completely start over from nothing.  If we want to add it that's fine, but I think a normal `docker-compose down` is still useful because there were a lot of time I wanted to do that but didn't want to lose all the data in my DB.

appspec.yml
```diff
@@ -1,18 +0,0 @@
...
-      timeout: 600
-      runas: root
-  ApplicationStart:
-    - location: scripts/application_start.sh
-      timeout: 600
-      runas: root
-  ApplicationStop:
-    - location: scripts/application_stop.sh
-      timeout: 60
-      runas: root
```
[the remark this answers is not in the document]
BatedUrGonnaDie: I thought I looked, but when you say buildspec that sounds more like what I was looking for.  I'll go take another look to see what I find.

docker-compose.yml
glacials: Alright np, if you get value out of it then let's leave it as-is

docker-compose.yml
BatedUrGonnaDie: I remember seeing somewhere that `:cached` was used mostly for source code, which is why I followed the article on only having it on `/app`

docker-compose.yml
BatedUrGonnaDie: `tmpfs` was one of the few things from the article that I really didn't understand what it was doing and what the consequences of enabling it was, so I decided to just ignore it for now.

The benefits of using the base image is that there is less duplication.  Right now each container a few of the same components (like `stdin_open` and `tty`).  The downside was that you cannot do `docker-compose up` any longer because there are services that do nothing.  However I took a look at the repo showing off these changes and they actually moved them to `x-...` config options, so I'm gonna take a look at moving to that because then we get the benefits with no downsides.

docker-compose.yml
BatedUrGonnaDie: Good catch!  I checked my images with `docker images` and redis, postgres, and fake-s3 are all on latest so I'll add a version pinning them to the tag that corresponds to latest as of today.

docker-compose.yml
BatedUrGonnaDie: I saw a `postgres-alpine` as well when checking the tag versions, I'll definitely see if they work out of the box.

docker-compose.yml
glacials: For the second point, I'm talking less about the inheritance of the abstract services and more about the fact that several of their non-abstract services are based on the same [`image`][1].

e.g. theirs is like:
```yaml
services:
  rails:
    image: example-dev:1.0.0
    ...
  runner:
    image: example-dev:1.0.0
    ...
  webpacker:
    image: example-dev:1.0.0
    ...
```
and ours is like
```yaml
services:
  web:
    image: web
    ...
  worker:
    image: worker
    ...
  webpacker:
    image: webpacker
    ...
```
My first reaction was maybe they are able to build once and reuse many times and we are not, but I thought I remembered you showing me once that we do reuse images, so I'm just confused what the difference is between the two setups I guess.


[1]: https://docs.docker.com/compose/compose-file/#image

BatedUrGonnaDie: Oh, I believe the benefit to that is that when you switch branches to work on something that might need to change the Dockerfile in some way, you can bump the version and that way anyone testing that branch only needs to build it once versus needing to build it every time they switch to it, and then when they switch back to another (ie: master) having to rebuild off master.

glacials: I think I understand the advantage of specifying a version number, but what's the advantage of specifying the same string for all the services?

BatedUrGonnaDie: Not an advantage per se, but since all of them use the base image you can just tie them all to the same number since you would need to increment them all if you changed the Dockerfile in any way.

BatedUrGonnaDie:
Docker has decided we're no longer friends and is throwing random errors randomly.  It appears to be related to volumes being mounted and unmounted, and I think depending on how fast it gets to the bundle install influences what error, if any.  Might take a bit longer to finish than I originally thought.

appspec.yml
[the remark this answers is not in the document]
BatedUrGonnaDie: I was going to investigate this, but after having so many problems with docker the last week I will just defer figuring out about this file to a later date since it's not necessary to proceed with the rest of the PR.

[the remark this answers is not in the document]
glacials: Yeah no worries, punt it!

glacials:
Okay I'll check it out!

BatedUrGonnaDie:
Beta deploys successfully once more!

So I took the time to make the Dockerfile for production and have it use a builder.  I also took what you had for development and tweaked it slightly because I was getting some errors and a lot of things in weird places.  For testing now you should be able to cache `bundle/` and `node_modules/` between runs and everything "should" work correctly.

Let me know what you think.

docker-compose.yml
BatedUrGonnaDie: Pretty sure this question went straight over my head a couple times.  Re-reading it now after working with the production dockerfile I think I understand the question now, but I don't have an answer.  I'm not sure what that name is actually used for in the end.

BatedUrGonnaDie:
Oh, I forgot to mention this but now production is a fully self-contained image, which if things work like I think they do would allow us to cache this to AWS image registry and pull that down before building for faster production builds.

So for dev/testing we can cache the two folders to speed up builds, and production should be able to cache the final image to a registry to speed up those builds.

Dockerfile-production
```diff
@@ -0,0 +1,63 @@
...
+RUN yarn install
+COPY . .
+RUN DATABASE_URL=nulldb://host/db SECRET_KEY_BASE="$(bundle exec rails secret)" REDIS_URL=redis://redis:6379/1 \
+    bin/rails assets:precompile
+
+RUN rm -rf node_modules/ tmp/cache app/assets vendor/assets spec
+
+#############
+
+FROM ruby:2.6.4-alpine
```
glacials: tim-and-eric-mind-blown

BatedUrGonnaDie: A quick breakdown is you use the first to pull in all dependencies and build them (a few more packages in the apk install).  Then once they're built you don't need the compiler, and in production we don't need anything from node_modules since its all been compiled so we can throw it all away and start in a clean fresh build.  Then we just copy over all the compiled assets and gems and we're good to go!

moorecp:
I just noticed this PR. You might want to roll back to ruby 2.6.3 until 2.6.5 comes out due to a string corruption bug in 2.6.4: https://bugs.ruby-lang.org/issues/16136

BatedUrGonnaDie:
Hmm, yeah that seems pretty serious.  I'll swap that tonight along with the env arg for stripe change that needs to be done.

Thanks for the heads up, appreciate it!

Issue #568, opened 2019-07-11
YouTube embeds


glacials:
Attaching a YouTube video to a run should embed the video in the page like it does for Twitch videos.

BatedUrGonnaDie:
Closed by #603

PR #603, opened 2019-09-13
Add support for displaying Youtube videos for runs like Twitch videos

moorecp:
This addresses #568. It pulls in the [Youtube IFrame API](https://developers.google.com/youtube/iframe_api_reference) and duplicates the functionality of the Twitch JS.

I monkey-patched `String` in order to clean up some of the `video_url` related code. If you'd rather that not happen or be done differently, I'm open to suggestions.


lib/patches/string.rb
```diff
@@ -0,0 +1,28 @@
+class String
```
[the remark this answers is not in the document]
moorecp: @BatedUrGonnaDie I added a `Video` PORO and assoicated spec and got rid of the `String` monkey patch. This seems like a better solution.

app/javascript/youtube.js
```diff
@@ -0,0 +1,34 @@
+const async = require('async')
+
+window.onYouTubeIframeAPIReady = () => {
```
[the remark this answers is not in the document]
moorecp: I went ahead and moved the YouTube API shim to the bottom of `<body>`. I don't think there should be a load issue, but I'm willing to discuss it.

app/models/video.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def youtube?
+    /^(?:https?:\/\/)?(?:www\.)?youtu(?:be\.com|\.be)\/.+/i.match?(url)
+  end
+
+  def id
+    if youtube?
+      # https://www.youtube.com/watch?v=asdf1234&feature=related
+      # https://youtube.com/asdf1234
```
BatedUrGonnaDie: As far as I'm aware this is not a format for ID's, but for channels.

app/javascript/youtube.js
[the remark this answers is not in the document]
BatedUrGonnaDie: I'm not 100% on the loading thing, it's just that turbolinks and DOM related things do not play well together usually and we've had problems with it in the past.

app/models/run.rb
```diff
@@ -218,6 +218,12 @@ def possible_timesave(timing)
     duration(timing) - sum_of_best(timing)
   end
 
+  def video
+    return nil unless video_url
```
BatedUrGonnaDie: @glacials Do you think it's worth it to just pass in `nil` to video like how `Duration` handles it?  Since we aren't printing anything it seems pointless, but having things be consistent is also nice sometimes.

app/views/shared/_youtube_api.slim
```diff
@@ -0,0 +1,6 @@
+- if [@run, *@compare_runs].compact.map(&:video).compact.map { |video| video.youtube? }.any?
```
BatedUrGonnaDie: Do you know if this will have any negative effects if it's run more than once before a page reloads?  Turbolinks doesn't clear the page completely, just replaces the body and merges the head so there's a chance for it to download the API stuff more than once if a user navigates to a page with a YT video multiple times in one session.

app/models/video.rb
moorecp: Seems plausible. I'll see if there are any possible video URLs that don't use the `v` query string parameter. A quick glances makes it look like that's not the case, so that would simplify the URL check and `video_id` lookup.

app/models/run.rb
moorecp: I'll take a look at `Duration`. Consistency is important.

app/views/shared/_youtube_api.slim
moorecp: Just thinking it through, I feel like there could be weirdness with Turbolinks. If we were to load the API JS more than once, I don't think the end user would be negatively impacted, but it feels like there might be a potential situation where the API JS *doesn't* load like expected.

I'll do some searching to see if people have mentioned specific interactions between Turbolinks and the YouTube IFrame API JS.

app/javascript/youtube.js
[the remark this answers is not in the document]
moorecp: I moved the API JS to the `<head>` tag along side the Twitch JS and then tied in to the Turbolinks events so that things should play nicely with partial page loads and not do stuff more than once.

app/models/video.rb
```diff
@@ -0,0 +1,34 @@
...
+      id = /.+\/([^\/]+$)/.match(url)
+      id ? id[1] : nil
+    elsif twitch?
+      # https://www.twitch.tv/videos/12345678
+
+      id = /.+\/([^\/]+$)/.match(url)
+      id ? id[1] : nil
+    end
+  end
+end
```
glacials: I agree the complexity of video stuff is now at the point that splitting it into a class is a good idea! I'd go one step further and say this should be its own ActiveRecord model, database table, etc. for full separation. It might only be one or two columns to start, but I think it's worth it for the simpler mental model.

app/assets/stylesheets/youtube.sass
```diff
@@ -0,0 +1,6 @@
+.youtube-player
+  position: absolute
+  top: 0
+  right: 0
+  bottom: 0
+  left: 0
```
glacials: Since this is the exact same as `.twitch-player`, maybe we can merge them into one `.video-player` class, or at least into one file together if you think that's too aggressive.

app/views/runs/show.slim
```diff
@@ -21,9 +21,12 @@
   - if [@run, *@compare_runs].map(&:video_url).compact.any?
     .card-deck.mx-1
       - [@run, *@compare_runs].each do |run|
-        - if run.video_url.present? && URI.parse(run.video_url).host.match?(/^(www\.)?(twitch\.tv)$/)
+        - if run.video.twitch?
```
glacials: Great refactor!

app/javascript/youtube.js
```diff
@@ -0,0 +1,50 @@
...
+  if (window.YT !== undefined && typeof(window.YT.Player) === "function") {
+    setupYouTubeVideos()
+    return true
+  }
+  setTimeout(waitForYouTubeAPI, 100)
+}
+
+document.addEventListener('turbolinks:load', function() {
+  waitForYouTubeAPI()
+})
```
glacials: The flow of code in this file feels a bit hard to follow, especially with the polling for the YouTube JS being ready. Is there a reason it needs to be a different format from what's in `twitch.js`? (i.e. a single `turbolinks:load` listener)

app/models/run.rb
glacials: Yeah it might be worth trying it out to see how it feels if it's not too much work. For reference this is called the [null object pattern][1]. I've been really liking `Duration` ever since we made that change. However if doing this becomes more complicated alongside my other comment re: making Video a proper ActiveRecord model, then don't worry about it for now.

[1]: https://en.wikipedia.org/wiki/Null_object_pattern

app/models/video.rb
moorecp: Sounds good. Are you thinking you want to go as far as using STI and having a base `Video` class with child classes of `class Twitch < Video` and keeping their specific logic contained therein?

app/assets/stylesheets/youtube.sass
moorecp: That sounds fine. I considered it originally but didn't as that would affect the existing Twitch JS code, but it shouldn't be a huge deal to tweak it to be able to figure out if we need to load the Twich player vs the YouTube player.

app/javascript/youtube.js
moorecp: The reason for the polling listener triggered by `turbolinks:load` is it was possible to hit `YT.Player(...)` before that object was fully initialized and a JS error of `YT.Player is not a constructor` would get thrown. The YouTube API JS that is included in the `<head>` is just a shim that then loads the actual API JS code asynchronously, so lots of load-order shenanigans are possible.

I'd like to get rid of the complexity and what amounts to basically busy waiting and am open to suggestions.

app/models/run.rb
moorecp: The current version does take `nil` and no longer does this check. But it sounds like things are moving to a real AR model, so this will go away.

moorecp:
> As a general comment, I'd ask that we avoid force pushing during PRs if possible, as some of the comments I'm late to the party for are giving 404s when I try to click through for context.

I almost never force push, but was keeping the commit history squashed to make it better for merging because I underestimated the scope of the PR and didn't expect as much interest in it. I'll hold off of squishing things down until the end in the future.

BatedUrGonnaDie:
From what I can tell from other projects, they usually have you do one final squash at the end right before merging.  We usually just take GitHub's option to squash and commit though and have all the messages from the commits be in the sub-message (if that's what it's called).

app/models/video.rb
glacials: My first gut reaction is that STI sounds a bit overkill here, but I'll leave the decision to you and what feels cleanest. It wouldn't be the end of the world if it was still a generic `url` field that we inspect in `Video` methods, since that's the simplest baby step for the data. We can always further refine it later.

app/javascript/youtube.js
glacials: In that case, is it possible the [code][1] in `twitch.js` encounters the same problem?

Also I think there's an `onload="doAThing()"` attribute you can add to `<script>` tags, maybe that would help with getting rid of the polling at least.

[1]: https://github.com/glacials/splits-io/blob/master/app/javascript/twitch.js#L10

app/assets/stylesheets/youtube.sass
glacials: Ohh then nevermind, I didn't realize `.twitch-player` was being used to select elements, thought it was just for this CSS here. No worries then.

app/javascript/youtube.js
moorecp: No, the Twitch code is fine. The script that loads is the complete JS.

The `onload` attribute probably wouldn't help since we need the script that YouTube downloads after the initial script loads.

glacials: Ahh I see, the script we include downloads another script! Sounds crazy and awkward to deal with.

Their [iframe player API reference][1] seems to say the script tries to call a global function called `onYouTubeIframeAPIReady` when it's loaded, can we take advantage of that?

[1]: https://developers.google.com/youtube/iframe_api_reference#Getting_Started

app/assets/stylesheets/youtube.sass
BatedUrGonnaDie: If it's just being used for element selection, we could always just use something like `data-youtube` and `data-twitch` attributes.

app/javascript/youtube.js
moorecp: I think it won't play nicely with Turbolinks. On the initial page load, that function would probably get called after the DOM finished loading, but on subsequent partial loads, I'm don't think it would even get called since the JS wouldn't get reloaded. Even if it did, I don't think you can rely on the order of operations due to the async nature of the API JS.

app/models/video.rb
moorecp: The current version is now just what you described. `Video` is a simple AR model that acts basically like the PORO version was.

Doing this involved many changes in getting rid of `Run.video_url`, but I think I pretty well got them all and the test suite is still green. Due to `strong_migrations`, there will need to be 1 deploy done before you can drop `video_url` on `runs`, but the column is currently being ignored via ActiveRecord.

app/assets/stylesheets/youtube.sass
moorecp: The CSS has been combined and it now uses `data` attributes to do the correct thing for element selection.

glacials:
@BatedUrGonnaDie: Nope! When Rails boots, it discovers all columns for all tables so that during ActiveRecord calls like `Run.first`, it can do `SELECT id, category_id, program, ... FROM runs` instead of `SELECT * FROM runs` (I think because it's slightly more efficient?).

So if this branch deleted `video_url` then got deployed to beta, production would start erroring on nearly every request with something like `unknown column video_url`. What we actually want to do is tell both beta and production to ignore the column, _then_ it's safe to delete it.

moorecp:
Yup, I wrote the drop column migration, but I haven't committed it for just that reason.

app/views/runs/show.slim
```diff
@@ -18,15 +18,15 @@
 = render partial: 'runs/full_timeline', locals: {timing: timing, run: @run, compare_runs: @compare_runs}
 
 - if @compare_runs.none? || (@compare_runs.any? && current_user.try(:patron?, tier: 3))
-  - if [@run, *@compare_runs].map(&:video_url).compact.any?
+  - if [@run, *@compare_runs].map(&:video).compact.map(&:url).compact.any?
```
glacials: For the future when `@compare_runs` can be any number of runs, this turns into an n+1 query for all those videos. Can it be eliminated?

app/models/video.rb
```diff
@@ -0,0 +1,49 @@
...
+
+  def video_type
+    if twitch?
+      TWITCH
+    elsif youtube?
+      YOUTUBE
+    else
+      nil
+    end
+  end
```
glacials: `video.video_type` sounds redundant, but I know `type` is reserved, so can we use something like `provider` instead?

app/views/users/pbs/export/panels/index.erb
```diff
@@ -3,12 +3,12 @@
 <% longest[:game_name] = (['game'.length] + @runs.map { |run| run.game.name.length }).max || 'Game'.length %>
 <% longest[:category_name] = (['category'.length] + @runs.map { |run| run.category.name.length }).max || 'Category'.length %>
 <% longest[:time] = (['pb'.length] + @runs.map { |run| "[#{format_ms(run.duration_ms(run.default_timing))}](#{link(run)})".length }).max || 'PB'.length %>
-<% longest[:video_url] = (['video'.length] + @runs.map { |run| "[Watch](#{run.video_url})".length }).max || 'Video'.length %>
-<% if @runs.where.not(video_url: nil).count > 0 %>
+<% longest[:video_url] = (['video'.length] + @runs.map { |run| "[Watch](#{run.video&.url})".length }).max || 'Video'.length %>
+<% if @runs.joins(:video).where.not(videos: { url: nil }).count > 0 %>
```
glacials:
```ruby
@runs.joins(:video)
```
seems like it should already only return runs with videos, so I think the `where` is redundant.

db/migrate/20190915141533_create_videos.rb
```diff
@@ -0,0 +1,10 @@
+class CreateVideos < ActiveRecord::Migration[6.0]
+  def change
+    create_table :videos do |t|
+      t.references :run, index: true
+      t.string :url
```
glacials: Do you think these fields should be non-nullable?

db/migrate/20190915142043_migrate_run_video_urls_to_videos.rb
```diff
@@ -0,0 +1,22 @@
+class MigrateRunVideoUrlsToVideos < ActiveRecord::Migration[6.0]
+  # video_url is included in ignored_columns in Run, so we can't use standard ActiveRecord stuff to create Videos or rollback the migration
+  def up
+    sql = "INSERT INTO videos (run_id, url, created_at, updated_at)
+           SELECT id, video_url, now(), now()
+           FROM runs
+           WHERE
+             video_url IS NOT NULL
+             AND video_url <> ''
+          ".squish
```
glacials: I'd actually slightly prefer that we don't carry over `created_at` and `updated_at`, as they might be wrong when talking strictly about when the video was added.

app/helpers/application_helper.rb
```diff
@@ -5,7 +5,7 @@ def site_title
 
   def order_runs(runs)
     dir = %w[asc desc].include?(params[:order]) ? params[:order] : 'desc'
-    runs.order(by(params[:by]) => dir)
+    runs.joins(:video).order(by(params[:by]) => dir)
```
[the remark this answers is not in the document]
moorecp: The main reason was I wasn't thinking. Apart from that, no, it should be `includes`.

app/models/user.rb
```diff
@@ -74,7 +74,10 @@ def pb_for(timing, category)
   end
 
   def pbs
-    runs.where.not(category: nil).select('DISTINCT ON (category_id) *').order('category_id, realtime_duration_ms ASC')
+    # TODO: Put the SELECT * back in after Run.ignored_columns is removed
+    # runs.where.not(category: nil).select('DISTINCT ON (category_id) *').order('category_id, realtime_duration_ms ASC')
```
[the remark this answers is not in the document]
moorecp: Once the column is actually deleted from the database, `SELECT *` will union properly with `runs.by_category` again. It's just due to `runs.by_category` going through ActiveRecord and thus `video_url` is left out of the `SELECT` it builds, but it is included in `SELECT *`.

In its current state with `Run.ignored_columns = ["video_url"]`, this is broken with `SELECT *` and the `UNION`.

app/views/users/pbs/export/panels/index.erb
moorecp: Agreed.

db/migrate/20190915141533_create_videos.rb
moorecp: Probably. What is the point of a `Video` without a `URL`.

db/migrate/20190915142043_migrate_run_video_urls_to_videos.rb
moorecp: They're not being carried over. Both fields are being set to the current time on the DB server.

app/models/video.rb
moorecp: I like `provider` a lot more than `video_type`.

db/migrate/20190915142043_migrate_run_video_urls_to_videos.rb
glacials: Oh sorry about that, I misread.

PR #604, opened 2019-09-14
Ignore segments following a skipped split when calculating segment history stats

### LINKED ISSUE #562: Standard deviation may be affected by skipped splits
It seems like our standard deviations might be operating on collapsed segments rather than actual segments, causing them to spike really high when the previous segment was skipped.

See: https://twitter.com/Gikkman/status/1145410450508656640

moorecp:
Closes #562.

When calculating segment_history_stats, skipped splits would cause time to accumulate in later splits.

I didn't want to turn this into a direct SQL query, but ActiveRecord was mishandling the result of the query if compiled using ActiveRecord notation for most of it, despite that fact it was generating identical raw SQL.

    SegmentHistory.joins(segment: :run)
      .joins(%Q{LEFT JOIN (SELECT segment_histories.id AS id, segment_histories.attempt_number AS attempt_number, segments.segment_number as segment_number FROM segment_histories INNER JOIN "segments" ON "segments"."id" = "segment_histories"."segment_id" WHERE segments.run_id = #{id} AND NOT (("segment_histories"."realtime_duration_ms" = 0 OR "segment_histories"."realtime_duration_ms" IS NULL))) AS other_histories ON other_histories.attempt_number = segment_histories.attempt_number AND other_histories.segment_number = segments_segment_histories.segment_number - 1})
      .where(segment: {runs: {id: id}})
      .where.not(Run.duration_type(timing) => [0, nil])
      .where("segments_segment_histories.segment_number = 0 OR (other_histories.attempt_number = segment_histories.attempt_number AND (other_histories.segment_number = segments_segment_histories.segment_number - 1)")
      .group(:segment_id)
      .select(stats_select_query(timing))

Because of that, I just wrote the SQL manually and executed it.

I'm leaving the PR as a draft for now looking for feedback on the solution and while I look at trying to get some tests written that would fail under the old method and pass with the new query.

### DIFF
--- a/app/models/run.rb
+++ b/app/models/run.rb
@@ -217,41 +217,69 @@ class Run < ApplicationRecord
   private
 
   def segment_history_stats_query(timing)
-    timing_string = case timing
-                    when Run::REAL
-                      'real'
-                    when Run::GAME
-                      'game'
-                    else
-                      raise 'Unsupported timing'
-                    end
-
-    sql = %Q{
+    sql = nil
+    case timing
+    when Run::REAL
+      sql = Run.sanitize_sql_array [%Q{
       SELECT
         segment_id,
-        STDDEV_POP(segment_histories.#{timing_string}time_duration_ms) AS standard_deviation,
-        AVG(segment_histories.#{timing_string}time_duration_ms) AS mean,
-        PERCENTILE_DISC(.5) WITHIN GROUP (ORDER BY segment_histories.#{timing_string}time_duration_ms) AS median,
-        PERCENTILE_CONT(.1) WITHIN GROUP (ORDER BY segment_histories.#{timing_string}time_duration_ms) AS percentile10,
-        PERCENTILE_CONT(.9) WITHIN GROUP (ORDER BY segment_histories.#{timing_string}time_duration_ms) AS percentile90,
-        PERCENTILE_CONT(.99) WITHIN GROUP (ORDER BY segment_histories.#{timing_string}time_duration_ms) AS percentile99 FROM segment_histories
+        STDDEV_POP(segment_histories.realtime_duration_ms) AS standard_deviation,
+        AVG(segment_histories.realtime_duration_ms) AS mean,
+        PERCENTILE_DISC(.5) WITHIN GROUP (ORDER BY segment_histories.realtime_duration_ms) AS median,
+        PERCENTILE_CONT(.1) WITHIN GROUP (ORDER BY segment_histories.realtime_duration_ms) AS percentile10,
+        PERCENTILE_CONT(.9) WITHIN GROUP (ORDER BY segment_histories.realtime_duration_ms) AS percentile90,
+        PERCENTILE_CONT(.99) WITHIN GROUP (ORDER BY segment_histories.realtime_duration_ms) AS percentile99
+      FROM segment_histories
       INNER JOIN segments ON segments.id = segment_histories.segment_id
       INNER JOIN runs ON runs.id = segments.run_id
       LEFT JOIN (
         SELECT segment_histories.id AS id, segment_histories.attempt_number AS attempt_number, segments.segment_number as segment_number
         FROM segment_histories
         INNER JOIN segments ON segments.id = segment_histories.segment_id
-        WHERE segments.run_id = #{id}
-        AND NOT ((segment_histories.#{timing_string}time_duration_ms = 0 OR segment_histories.#{timing_string}time_duration_ms IS NULL))
+        WHERE segments.run_id = :run_id
+        AND NOT ((segment_histories.realtime_duration_ms = 0 OR segment_histories.realtime_duration_ms IS NULL))
       ) AS other_histories ON other_histories.attempt_number = segment_histories.attempt_number AND other_histories.segment_number = segments.segment_number - 1
-      WHERE runs.id = #{id}
-        AND NOT ((segment_histories.#{timing_string}time_duration_ms = 0 OR segment_histories.#{timing_string}time_duration_ms IS NULL))
+      WHERE runs.id = :run_id
+        AND NOT ((segment_histories.realtime_duration_ms = 0 OR segment_histories.realtime_duration_ms IS NULL))
         AND (segments.segment_number = 0 OR
           (other_histories.attempt_number = segment_histories.attempt_number
           AND other_histories.segment_number = segments.segment_number - 1)
         )
       GROUP BY segment_id
-    }.squish
-    ActiveRecord::Base.connection.execute(sql).to_a
+    }.squish, run_id: id]
+
+    when Run::GAME
+      sql = Run.sanitize_sql_array [%Q{
+      SELECT
+        segment_id,
+        STDDEV_POP(segment_histories.gametime_duration_ms) AS standard_deviation,
+        AVG(segment_histories.gametime_duration_ms) AS mean,
+        PERCENTILE_DISC(.5) WITHIN GROUP (ORDER BY segment_histories.gametime_duration_ms) AS median,
+        PERCENTILE_CONT(.1) WITHIN GROUP (ORDER BY segment_histories.gametime_duration_ms) AS percentile10,
+        PERCENTILE_CONT(.9) WITHIN GROUP (ORDER BY segment_histories.gametime_duration_ms) AS percentile90,
+        PERCENTILE_CONT(.99) WITHIN GROUP (ORDER BY segment_histories.gametime_duration_ms) AS percentile99
+      FROM segment_histories
+      INNER JOIN segments ON segments.id = segment_histories.segment_id
+      INNER JOIN runs ON runs.id = segments.run_id
+      LEFT JOIN (
+        SELECT segment_histories.id AS id, segment_histories.attempt_number AS attempt_number, segments.segment_number as segment_number
+        FROM segment_histories
+        INNER JOIN segments ON segments.id = segment_histories.segment_id
+        WHERE segments.run_id = :run_id
+        AND NOT ((segment_histories.gametime_duration_ms = 0 OR segment_histories.gametime_duration_ms IS NULL))
+      ) AS other_histories ON other_histories.attempt_number = segment_histories.attempt_number AND other_histories.segment_number = segments.segment_number - 1
+      WHERE runs.id = :run_id
+        AND NOT ((segment_histories.gametime_duration_ms = 0 OR segment_histories.gametime_duration_ms IS NULL))
+        AND (segments.segment_number = 0 OR
+          (other_histories.attempt_number = segment_histories.attempt_number
+          AND other_histories.segment_number = segments.segment_number - 1)
+        )
+      GROUP BY segment_id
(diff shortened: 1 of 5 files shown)


app/models/run.rb
```diff
@@ -220,28 +216,42 @@ def possible_timesave(timing)
...
+
+    sql = %Q{
+      SELECT
+        segment_id,
+        STDDEV_POP(segment_histories.#{timing_string}time_duration_ms) AS standard_deviation,
+        AVG(segment_histories.#{timing_string}time_duration_ms) AS mean,
+        PERCENTILE_DISC(.5) WITHIN GROUP (ORDER BY segment_histories.#{timing_string}time_duration_ms) AS median,
+        PERCENTILE_CONT(.1) WITHIN GROUP (ORDER BY segment_histories.#{timing_string}time_duration_ms) AS percentile10,
+        PERCENTILE_CONT(.9) WITHIN GROUP (ORDER BY segment_histories.#{timing_string}time_duration_ms) AS percentile90,
+        PERCENTILE_CONT(.99) WITHIN GROUP (ORDER BY segment_histories.#{timing_string}time_duration_ms) AS percentile99 FROM segment_histories
```
glacials: Can we move the `FROM` to its own line please :3 for clarity

app/models/run.rb
```diff
@@ -220,28 +216,42 @@ def possible_timesave(timing)
...
+      INNER JOIN segments ON segments.id = segment_histories.segment_id
+      INNER JOIN runs ON runs.id = segments.run_id
+      LEFT JOIN (
+        SELECT segment_histories.id AS id, segment_histories.attempt_number AS attempt_number, segments.segment_number as segment_number
+        FROM segment_histories
+        INNER JOIN segments ON segments.id = segment_histories.segment_id
+        WHERE segments.run_id = #{id}
```
glacials: I get really really uncomfortable doing interpolation in a SQL string. Can we do constructed statements instead (`'where segments.run_id = :id', {id: id}`)?

Same for interpolation for `timing_string` in fact. I know in this method it's guaranteed to be safe because of the `case` above, but often code gets copied/pasted in smaller parts or changed around slowly over time without remembering how important the guards are, so I always prefer that it be made foolproof from the beginning.

app/models/run.rb
```diff
@@ -220,28 +216,42 @@ def possible_timesave(timing)
...
+        AND NOT ((segment_histories.#{timing_string}time_duration_ms = 0 OR segment_histories.#{timing_string}time_duration_ms IS NULL))
+      ) AS other_histories ON other_histories.attempt_number = segment_histories.attempt_number AND other_histories.segment_number = segments.segment_number - 1
+      WHERE runs.id = #{id}
+        AND NOT ((segment_histories.#{timing_string}time_duration_ms = 0 OR segment_histories.#{timing_string}time_duration_ms IS NULL))
+        AND (segments.segment_number = 0 OR
+          (other_histories.attempt_number = segment_histories.attempt_number
+          AND other_histories.segment_number = segments.segment_number - 1)
+        )
+      GROUP BY segment_id
+    }.squish
```
glacials: I'm going to be 100% honest here, this entire query is way beyond me and my poor SQL experience. I'm physically and mentally exhausted right now so maybe my brain not be do working good, but my eyes are just glazed over from trying to build the mental model for what all the joins are doing.

I'll take a look when I'm in a better mental state, but I do think (assuming this is the simplest query for the job) we could use some comments or somesuch to help readers like myself understand what this is doing and why.

app/models/run.rb
BatedUrGonnaDie: As a note, I'm pretty positive you cannot use constructed statements with the column name, which is why the old version was the same query twice.

moorecp: Yeah, you can't do that with column names, do I just duplicated the query like the old version.

app/models/run.rb
moorecp: If it wasn't clear in my PR description, I'm not really jazzed about this, either. It's tough to read and thus tough to maintain. Even with comments, it still won't be great, but I can definitely add them. Additionally, I'm sure not betting the farm on this being 100% correct. I'd like as many eyes on this as possible.

spec/models/run_spec.rb
```diff
@@ -589,4 +589,39 @@
...
+      r.parse_into_db
+      r.reload
+      r
+    end
+
+    it 'has the correct stats' do
+      stats = run.segment_history_stats(Run::REAL)
+      ids = run.segments.order(:segment_number).map(&:id)
+
+      expect(stats[ids[0]][:standard_deviation].to_f).to eq 48989.79485566
```
BatedUrGonnaDie: We've had problems in the past with using exact decimals to this precision, might be better to use the `be_within` matcher.

BatedUrGonnaDie:
Might be a bit late for this, but what exactly was the problem with the ActiveRecord notation version?

moorecp:
There is no way to write the `LEFT JOIN` in pure AR notation and using `.joins` with a string was generating the exact same SQL, but ActiveRecord wasn't mapping the results correctly. You'd just get `segment.id` and nothing else.

BatedUrGonnaDie:
Did you try calling the methods or just inspecting the returned objects?  Even with my statements before inspecting them only showed the id but with the dot notation the `SELECT ... AS` would still be in the returned object just kinda hidden.

moorecp:
I really thought I did try just calling the methods on the results, but either I didn't or had something else going on. I just tried it again and it worked, so I put the partial AR notation version back in.

BatedUrGonnaDie:
It looks like rails 5 added a `left_join`, but it's a `LEFT OUTER JOIN`.  Is that different from a normal `LEFT JOIN`?   (my sql is not very great)

moorecp:
`LEFT JOIN` and `LEFT OUTER JOIN` are the same thing. I'll see if it can be used to clean things up, but I'm not hopeful due to the complexity of it (a subselect with an inner join with an alias).

BatedUrGonnaDie:
Yeah, this query is getting pretty complex, so I'm not sure that the AR notation version would be more readable or not.  I mostly just wanted to know the difference 👍

moorecp:
I'm thinking it can probably be cleaned up via something like `Arel`, but I don't have a lot of experience with that.

app/models/run.rb
```diff
@@ -177,8 +177,10 @@ def previous_pb(timing)
   # instead of individually for each segment (N queries)
   def segment_history_stats(timing)
     stats = SegmentHistory.joins(segment: :run)
+                          .joins(stats_join_additional_histories_query(timing))
                           .where(segment: {runs: {id: id}})
                           .where.not(Run.duration_type(timing) => [0, nil])
+                          .where("segments_segment_histories.segment_number = 0 OR (other_histories.attempt_number = segment_histories.attempt_number AND other_histories.segment_number = segments_segment_histories.segment_number - 1)")
```
glacials: Which part of the query stops us from getting back segments directly following skipped segments? Are skipped SegmentHistorys absent from the database and the join is enough to filter them out?

moorecp: The `.joins(stats_join_additional_histories_query(timing))` pulls in a second copy of the histories (`other_histories`) that doesn't include histories with a `NULL` or `0` value for the timing we care about so that we can use it for filtering.

`segments_segment_histories` represents the histories we actually care about and use for calculations. `other_histories` (probably not the best name as I sit here and type it out again) are used to be the history immediately prior to `segments_segment_histories`.

The part that stops us from getting back the histories we don't want is the additional `.where`. If the `segment_segment_histories.segment_number` is `0`, then we're the history for the first segment, so we always include that since there's no previous segment to be skipped. Otherwise, we have to look at the segment history of the segment immediately prior to `segments_segment_histories`. We do that with `other_histories.attempt_number = segment_histories.attempt_number AND other_histories.segment_number = segments_segment_histories.segment_number - 1`.

If that history exists in `other_histories`, that means it wasn't skipped, therefore the next history (`segments_segment_histories`) is valid.

moorecp:
> As a general comment, it seems like this bug has revealed that our data is corrupt because of how we track segment times. These frankenstein durations that represent the sum of two segments but are put into a slot for one's duration are confusing, inaccurate, and mostly useless. Eventually I want us to migrate from durations to "run duration at the time the segment ended", which in this case would naturally yield nulls for skipped splits without affecting other splits.

I'm not quite following how that would actually solve this situation. If you have 3 segments in your LSS and skipped split 2, in the database, you'd have a run duration at the time split 1 ended and a run duration at the time split 3 ended. That second run duration would still be too long and you'd have to look to see if there was a hole before the second stored duration to know to ignore it. That being said, it might still be a better way of handling durations.

Does it makes sense to store a value on a `SegmentHistory` that simply states if it should be used (or skipped) for stat calculations? It could/should be set at parse time, I would think.

glacials:
Yeah that's a good point, I guess the data wouldn't be _cleaner_ so much as it would be... more appropriately titled? Since `segment3.duration` should return `nil` because _we don't know segment3's duration_.

> Does it makes sense to store a value on a SegmentHistory that simply states if it should be used (or skipped) for stat calculations? It could/should be set at parse time, I would think.

Yeah this is a question I had floating in my head too. I don't know the answer; no solution is particularly great. You could store the value but then you have duplication of data; basically `segmentN.skipped?` == `segmentN+1.trust_my_duration?`. Always feels off to duplicate data because I get anxious they'll fall out of sync.

But it also feels off _not_ to do that, because imo a segment (or any object) should stand on its own without needing to potentially look up all its siblings to find out how to display it.

I think this is why I brought up "run duration at the segment's end", because then you'd have fields for duration (naturally null if previous was skipped) _and_ for segment end time (NOT null if previous was skipped) and by the nature of what those fields represent, they don't depend on their sibling segments to inform whether or not you can usefully visualize the segment in any given view.

You just use the one that has the information you want; if you're charting segment durations, you chart `segments.map(&:duration)` and the `nil`s naturally appear where you want. If you're filling out the run page's Finished At column, you just fill out `segments.map(&:end)` and previous segments' skips don't bother you.

All that said, what we have in this PR works and it seems like a big overhaul to do it any other way, with little actual user benefit. So I'm happy not changing course here.

app/models/segment_history.rb
```diff
@@ -22,6 +22,41 @@ def self.with_ends
     '.squish)
   end
 
+  def self.without_statistically_invalid_histories_for_run(run, timing)
```
glacials: Can we not use `run` (i.e. `self.segment.run`) here?

moorecp: Not in a class method.

glacials: Oh right duh. Is there any way to avoid needing to pass the run into this method?

moorecp: The join needs the Run ID, so we have to either have it be part of `Run` so we can just use `self.id` or either the run or its ID has to be passed in to whatever method is written. I also don't like passing a `Run` in since it tightly couples the 2 things together, but I think that stems from the fact these things are tightly coupled.

glacials: Okay, that's fair.

PR #606, opened 2019-09-16
Ignore splits after a skipped split in some charts

### LINKED ISSUE #527: Segment Duration over Time graph doesn't ignore skipped splits
When a split is skipped, the following split should not absorb its duration in the context of the Segment Duration over Time graph, but it seems that it is.

e.g. Bianco 3 and Bianco 7 on https://splits.io/3u5z

moorecp:
Closes #527 

This takes the code from `segment.js` that ignored splits following a skipped split and moves it higher up so that all charts can use it easily. The only real complication came from the augmented object hierarchy not being mirrored in the example object used for the landing page. For the time being, I just used the original segment histories as a fallback if the filtered histories were completely missing (as they are on the landing page).

Reset Rate on Individual Segment Stats and Resets Per Split both want the unfiltered histories, so we can't just only use the filtered histories.

### DIFF
--- a/app/javascript/chart_builder.js
+++ b/app/javascript/chart_builder.js
@@ -48,13 +48,40 @@ document.addEventListener('turbolinks:load', function() {
     }
     throw new Error('Request for run from API failed')
   }).then(function(bodies) {
-    const runs = bodies.map(body => body.run)
+    let runs = bodies.map(body => body.run)
     const timing = new URLSearchParams(window.location.search).get('timing') || runs[0].default_timing
 
     document.getElementById('chart-spinner').hidden = true
     document.getElementById('chart-holder').hidden = false
     if (runs[0].histories.length !== 0) {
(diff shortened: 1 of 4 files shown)


app/javascript/chart_builder.js
```diff
@@ -48,13 +48,40 @@ document.addEventListener('turbolinks:load', function() {
...
     const timing = new URLSearchParams(window.location.search).get('timing') || runs[0].default_timing
 
     document.getElementById('chart-spinner').hidden = true
     document.getElementById('chart-holder').hidden = false
     if (runs[0].histories.length !== 0) {
       const skipped = `${timing}time_skipped`
+      const duration = `${timing}time_duration_ms`
+      runs = runs.map((run) => (
+        {
+          ...run,
```
glacials: More syntax I didn't know about! This is so much better than `_.merge` or whatever I was using.

PR #607, opened 2019-09-17
Improve split timeline comparisons

### LINKED ISSUE #582: Make comparing later splits easier
Once upon a time, on the compare page you could drag the shorter timeline left and right to line up the two splits you wanted to compare / focus your attention around. We removed it because it was too awkward a UX and too hard to even discover it existed, but it did serve a need that now we have again.

We should find a better UX to implement this feature. Maybe something like, if you hover over a segment in one timeline, the other timeline moves horizontally to align its equivalent segment. Still brainstorming though.

moorecp:
Closes #582 

I feel like this PR is serving as a possibility, but probably not the ultimate solution. It implements @glacials's idea in #582. It works, but it feels sort of jarring as just moving your mouse around causes timelines to jump a bunch.
timelines

### DIFF
--- a/app/assets/stylesheets/timeline.sass
+++ b/app/assets/stylesheets/timeline.sass
@@ -56,6 +56,7 @@
 .gold.timeline
   background: $green
   font-size: .5em
+  position: relative
 
 .gold-split
   background-color: $light-black
@@ -66,8 +67,14 @@
   padding: 0
   width: .45em
 
+.timeline-animation
+  left: 0px
+  transition-property: left
+  transition-duration: 0.2s
+
 .timeline-background
   background: repeating-linear-gradient(-45deg, rgba(255, 255, 255, .1), rgba(255, 255, 255, .1), 6px, rgba(0, 0, 0, .1) 6px, rgba(0, 0, 0, .1) 12px)
+  position: relative
 
 .video-progress-line
   height: 6em
--- a/app/assets/stylesheets/timeline.sass
+++ b/app/assets/stylesheets/timeline.sass
@@ -69,8 +69,8 @@
 
 .timeline-animation
   left: 0px
-  transition-property: left
   transition-duration: 0.2s
+  transition-property: left
 
 .timeline-background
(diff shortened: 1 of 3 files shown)


glacials:
Nice! Hmm yeah it definitely seems a bit jumpy trying it out, do you think something like smoothing the jump animation would help or do you think we need a whole different idea?

moorecp:
> Nice! Hmm yeah it definitely seems a bit jumpy trying it out, do you think something like smoothing the jump animation would help or do you think we need a whole different idea?

I was looking at maybe trying to add some sort of animation to smooth it out a little, but I haven't tried that, yet. Something I had tried was adding a delay before the movement took place of ~100ms, such that if you moved back off of the timeline before the 100ms, the movement wouldn't occur. That seemed to help some, but I needed to do it a little better as it caused some display bugs.

I might spend some more time on it and try to flesh out one or both of those ideas and see how it feels.

moorecp:
I've added a little delay to when the timelines move and they also animate a little so things aren't as jarring.
timeline_mk2

moorecp:
I looked into CSS animations at one point. I didn't think they'd work in this situation because they kick in when you transition between states via classes and we're hardcoding the positions, but I didn't actually try. I'll give it a quick shot and see what happens.

moorecp:
I found some CSS animation stuff that was simpler and worked like I would have hoped.

I also found some issues if comparing 2 runs with very different number of splits that I fixed.

glacials:
Sweet, looks great! I just noticed one thing I'd like to ask before merging -- is it possible to have the timeline inspector hover trigger immediately as before, and have the slight delay only apply to the timeline motion?

I'm concerned that if people don't witness anything happen in those <100ms that their cursor quickly passes over the timeline, they won't realize that anything can happen at all on hover. Having the timeline inspector element show immediately might make them keep their cursor there long enough to also see that the comparison timeline adjusts.

moorecp:
Sounds good. The timeline inspector now pops/hides immediately again.

PR #609, opened 2019-09-17
Fix the positioning of the ISS chart on the Landing page

### LINKED ISSUE #571: Logged out front page graph isn't positioned correctly
firefox_2019-07-24_06-04-40

moorecp:
Closes #571 

On a Run page, the ISS chart is contained within a `div` that has `position: relative`. On the landing page, its parent has no positioning. This causes the axes to be rendered in the wrong place. This is only an issue on the ISS chart because it is the only one that has a separate `div` for its axes due to the fact you can scroll its contents.

### DIFF
--- a/app/assets/stylesheets/landing.sass
+++ b/app/assets/stylesheets/landing.sass
@@ -1,0 +1,2 @@
+.relative
+  position: relative
--- a/app/assets/stylesheets/landing.sass
+++ b/app/assets/stylesheets/landing.sass
@@ -1,2 +1,0 @@
-.relative
-  position: relative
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/landing.sass
```diff
@@ -0,0 +1,2 @@
+.relative
```
BatedUrGonnaDie: I believe bootstrap has a `.position-relative` class that does this already.

moorecp: Sure does. I just grepped in the `app/assets/stylesheets` directory. I didn't check the gems or anywhere else.

BatedUrGonnaDie: Since we use the custom paid theme for bootstrap we have it included minified somewhere.  Just be aware that it's still version 4.0.0.

PR #611, opened 2019-09-22
Funnel chart for showing progress of runs

### LINKED ISSUE #470: Funnel chart for attempts
Highcharts has a [funnel chart type](https://api.highcharts.com/highcharts/plotOptions.funnel), which would be nice to use for visualizing historically how far your runs get before you reset (each layer of the funnel being a segment you completed).

moorecp:
Closes #470 

This adds a funnel chart to show how far the user's runs progress before resetting.


glacials:
It looks like the shape/size of the funnel doesn't change based on the data, is that correct?

For instance, in [this run][1] the percentage of runs which reach Bianco 2 is the same as the percentage that have reached Bianco 3 (65.6% / 82 runs), but the Bianco 3 slice of the chart is still skinnier than the Bianco 2 slice.

Haven't looked too deeply into the Highcharts funnel chart so I'm not sure if that's expected or not, but atm it looks like the shape of the chart itself doesn't itself convey any information until you mouseover and look at the numbers.

[1]: https://splits.io/996

moorecp:
That's how the chart works. It *generally* funnels down. It's possible to have segments with the same values be different sizes and segments with different values be the same size.

The [example](https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/demo/funnel/) from the Highcharts website has different values for the final 3 segments, but they're all the same size.

I modified that example [here](https://jsfiddle.net/yo3a7hn2/) to show the same values as different sizes (as you saw yourself with that run).

So, the chart seems to be mostly used as a general, best-fit-line, sort of estimation. It's not 100% to scale or anything like that. The question then becomes, is that useful/interesting enough?

glacials:
Ohh I didn't even realize the _heights_ of the segments were changing based on # of times that segment was reached. Serves me right for only testing it on my own splits, I am reset-averse 😅Was expecting the width to be the thing that changes based on completions. Yeah this whole thing where the width is artificial so that it can fit into this static funnel shape is super weird, not sure why they went that direction.

Browsing around others' splits, the height differences of the segments are really hard to notice. The example you linked is a great demo of the vast differences in different portions of the funnel, but for what we're charting, the differences aren't big enough (and the number of segments is often too high) to make the chart convey any information.

Hell probably even a bar chart would be better at describing the falloff of each split:
```
█▆▅▄▃▁▃▁▂▂▂▁
```
(generated from https://splits.io/4krr with [spark][1])

Thanks for going through the effort of implementing this, but yeah it seems like right now the funnel chart isn't a useful enough visualization for our use case. I say either close #470 as wontfix or see if it can work with another chart type, your choice.

[1]: https://github.com/holman/spark

moorecp:
I'll take a look at other chart options. I think you're right that even a column chart is better. There is a pyramid chart that appears to be an inverse funnel chart, but is more intuitive, but still probably not interesting.

I'll take a look at a few things before throwing in the towel.

moorecp:
@glacials I added a line chart and a column chart locally to compare.

image

The problem with the line chart is if you are comparing to a run that has different/more splits than the source run, there are problems since the x-axis is shared.

The column chart doesn't have this problem as it just adds the next chart next to it.

image


With the line chart, you could use source segment names until you run out, then use whatever additional names are required, but the whole concept gets a little weird. Even ignoring the weirdness introduced by additional segments, do either of these charts provide anything interesting/useful?

BatedUrGonnaDie:
I feel like the column chart does provide some value, but it feels redundant with the reset graph we have already.  Maybe a 2 button toggle at the top to change it between the 2?  That way people could swap between the 2 without needing to take up the extra vertical space.

The comparison thing seems fine for me personally.  LiveSplit IIRC tries to compare on segment name and # of segments, and if they don't match then it just gives up.  So I think forsaking comparisons stuff for one graph would probably be fine.

glacials:
Yeah I agree it has some redundancy with the reset pie chart, but this is also way better at conveying the flow of the run from one split to the next. I like the idea of putting them in the same card with a toggle -- the big slices in the pie chart still have a more immediate visual impact than the cliffs do in this chart, so both have some distinct value.

And yeah +1 to what @BatedUrGonnaDie said: most charts and stats on the site get broken and/or useless when you're comparing two runs with different splits. Rather than fuss about getting everything on the site looking good for dissimilar data, we've learned to build for the default case of identical splits, and build other features to make it easier for communities to standardize long-term.

As for bar vs line, for me it's a wash -- both have small advantages and small disadvantages, so choose whichever you think will work better. One comment I have in general is that the y-axis could convey more information if it were raw # runs rather than % of runs, as it would still be obvious from the guide lines on the chart where 50% etc. is.

moorecp:
Locally, I've changed the y-axis to be based on # of runs. It works fine for non-comparison versions, but if you're comparing 2 runs with a significant difference in number or histories you have a compression problem. If one run has a max number of histories of 10 and the other run has a max of 150, the first run's bars (or line points) will be rendered over a very small vertical portion of the chart.

I'm not sure if we want to scale the bars based on percent, but label the y-axis with raw # based on the primary run and just display the real numbers in the tooltips.

glacials:
Ah nice catch, I hadn't thought of that. In that case, percent axis with raw # in the tooltips sounds good. I wouldn't label the y-axis with raw numbers at all, that could get confusing when looking at multiple runs.

moorecp:
I added buttons to swap between Resets and Run Progress. I did notice that when comparing runs, if the runs share some splits with the same name, they'll be grouped together on the Run Progress chart, which can make the comparison chart look a little wonky.

app/javascript/chart_builder.js
```diff
@@ -87,6 +88,7 @@ document.addEventListener('turbolinks:load', function() {
       buildBoxPlot(runs, chartOptions)
       buildSegmentChart(runs, chartOptions)
       buildResetChart(runs, chartOptions)
+      buildRunProgressVariwideChart(runs, chartOptions)
```
glacials: I'm getting errors that this function doesn't exist

moorecp:
I somehow just missed a commit. I'm pretty sure I've done software development before, but now I'm not so sure.

PR #613, opened 2019-09-24
Parser upgrade

BatedUrGonnaDie:
The new and improved parser!  Short benchmarking on my usual insanely huge run (60k+ run history elements alone) shows a large drop in memory usage.

Previous: Jump to 800MB, peaking ~900MB while waiting for GC.  21 requests in 10 minute span.
New: Jump to 200MB, peaking at 250MB while waiting for GC.  22 requests in 10 minute span.

Requests were measured with a simple `ab -t 600` on the runs controller with parsing happening on every request.

On another note, this same run when hitting V4 historic sits ~550MB, so realistically we save a little over 300MB of RAM from this change since the endpoint will get hit whenever the run page is loaded for the graphs.


app/models/concerns/unparsed_run.rb
```diff
@@ -13,154 +18,286 @@ def parse_into_db
...
-          filesize_bytes:          size
+          total_playtime_ms:       run_data[:total_playtime_ms],
+          default_timing:          run_data[:default_timing],
+          filesize_bytes:          run_data[:size]
         )
 
         HighlightSuggestion.from_run(self)
       end
+    rescue UnparsableRun
+      destroy
```
BatedUrGonnaDie: This is probably not the behavior we want atm, but wasn't sure exactly how we want to handle this right now since we could receive a request to parse before the hour is up in the presigned request.

app/models/concerns/unparsed_run.rb
```diff
@@ -13,154 +18,286 @@ def parse_into_db
...
+      run_data[:metadata][:variables]     = {}
+
+      run_data[:attempts]  = run.attempt_count
+      run_data[:offset_ms] = run.offset.total_seconds * 1000
+
+      # Grab all run metadata
+      # .with will clean up the iterator when we are done
+      run.metadata.variables.with do |iter|
+        until (variable = iter.next).nil?
+          run_data[:metadata][:variables][variable.name] = variable.value
```
BatedUrGonnaDie: Variables still aren't being saved, but since it's such a small part of the parser it didn't make much sense to remove it now.

app/models/concerns/unparsed_run.rb
```diff
@@ -13,154 +18,286 @@ def parse_into_db
...
+          started_at:           dates[0] && DateTime.parse(dates[0].to_rfc3339),
+          ended_at:             dates[1] && DateTime.parse(dates[1].to_rfc3339)
         )
-      end)
+
+        # See https://github.com/glacials/splits-io/pull/474/files#r241242051
+        dates.compact.each(&:dispose)
+
+        container << history
+        next if container.size < MAX_CONTAINER_SIZE
```
BatedUrGonnaDie: Do you know a better way to do this?  It seems hacky to check this and then also have to do it outside the loop in case we never hit the limit.

app/models/concerns/unparsed_run.rb
```diff
@@ -13,154 +18,286 @@ def parse_into_db
...
+
+      return if container.blank?
+
+      write_run_histories(container)
     end
 
-    def write_segments(parsed_segments)
-      segs = []
+    def write_run_histories(histories)
+      result = RunHistory.import(histories)
```
BatedUrGonnaDie: Originally the plan was to move away from `activerecord-import`, but turns out they have a good results object when using PG, so I ended up sticking with it since it allowed us to dump large batches while still checking for missing records easily.

app/models/concerns/unparsed_run.rb
```diff
@@ -13,154 +18,286 @@ def parse_into_db
...
+      return result unless result.failed_instances.any?
+
+      raise UnparsableRun
+    end
 
-      parsed_segments.each do |parsed_segment|
-        segs << Segment.new(
+    def process_segments(run)
+      container = {}
+      previous_segment = OpenStruct.new(realtime_end_ms: 0, gametime_end_ms: 0)
```
BatedUrGonnaDie: I use an OpenStruct here so that we don't need special first loop iteration handing.

app/models/concerns/unparsed_run.rb
```diff
@@ -13,154 +18,286 @@ def parse_into_db
...
-          gametime_start_ms:             parsed_segment[:gametime_start_ms], # starts can be 0 (for the first segment)
-          gametime_end_ms:               zero_to_nil(parsed_segment[:gametime_end_ms]),
-          gametime_duration_ms:          zero_to_nil(parsed_segment[:gametime_duration_ms]),
-          gametime_shortest_duration_ms: zero_to_nil(parsed_segment[:gametime_best_ms]),
-
-          gametime_skipped: parsed_segment[:gametime_skipped],
-          gametime_reduced: false,
-          gametime_gold:    parsed_segment[:gametime_gold]
+          segment_number: i,
+          name:           lsc_segment.name || '',
```
BatedUrGonnaDie: This should fix segments with no name being rejected from the DB.

app/models/concerns/unparsed_run.rb
```diff
@@ -13,154 +18,286 @@ def parse_into_db
...
+          icon:           lsc_segment.icon.presence # do something with this
         )
+
+        %w[real game].each do |timing|
```
BatedUrGonnaDie: This was originally used so that we don't have the exact same block of code repeated once for gametime and once for realtime, but it looks ugly and is hard to follow.  Do you have any good solutions for this?

app/models/concerns/unparsed_run.rb
```diff
@@ -13,154 +18,286 @@ def parse_into_db
...
 
-    def write_segment_histories(segs)
-      ids = segments.order(:segment_number).pluck(:id)
-      histories = []
+    def write_segments(segments)
+      segments.each do |seg|
+        %w[real game].each do |timing|
+          %w[time_end_ms time_duration_ms time_shortest_duration_ms].each do |field|
+            value = seg.send("#{timing}#{field}")
+            seg.send("#{timing}#{field}=", zero_to_nil(value))
```
BatedUrGonnaDie: I needed the value to be zero for the durations in the original loop because I calculate time a little differently, which is why we `zero_to_nil` them here instead.  The new calculation should stop runs from being seen as done if any split has a time (which I believe happens when you reset on your first run, I still need to test this one).

app/models/concerns/unparsed_run.rb
```diff
@@ -13,154 +18,286 @@ def parse_into_db
...
+
+      lsc_segments.each_with_index do |seg, i|
+        seg.segment_history.iter.with do |iter|
+          until (history_element = iter.next).nil?
+            next unless history_element.index.positive?
+
+            container << SegmentHistory.new(
+              segment_id:           segment_ids[i],
+              attempt_number:       history_element.index,
+              realtime_duration_ms: history_element.time.real_time&.total_seconds&.*(1000) || 0,
```
BatedUrGonnaDie: I wasn't sure if these two needed to be `zero_to_nil`'d or not.

db/schema.rb
```diff
@@ -322,6 +323,7 @@
     t.boolean "gametime_gold", default: false, null: false
     t.boolean "gametime_reduced", default: false, null: false
     t.boolean "gametime_skipped", default: false, null: false
+    t.string "icon"
```
BatedUrGonnaDie: Currently just saving the icon to the DB.  Is it preferable to use ActiveStorage for this instead?  They're encoded as base something strings in LiveSplit but I'm sure the conversion is easy enough.

@CryZe I remember you saying something about shrinking the images, is that something we should be doing before we save them?

CryZe: The latest released livesplit-core automatically shrinks the images when parsing the splits, so when you get them from the Run, they are already shrinked. However the base64 storage is not the case anymore, you get the raw image bytes now (I think it still is in 0.11, which is the latest version. The change is only on the master branch atm)

BatedUrGonnaDie: Hm, ok.  Will let @glacials decide one what he wants to do with the images.

Thanks for the heads up appreciate it.

app/models/concerns/unparsed_run.rb
```diff
@@ -13,154 +18,286 @@ def parse_into_db
...
+      # See https://github.com/glacials/splits-io/pull/474/files#r241242051
+      analysis.compact.each(&:dispose)
+
+      # Set run duration
+      last_seg = run.segment(run.len - 1)
+      run_data[:realtime_duration_ms] = last_seg.personal_best_split_time.real_time&.total_seconds&.*(1000) || 0
+      run_data[:gametime_duration_ms] = last_seg.personal_best_split_time.game_time&.total_seconds&.*(1000) || 0
+
+      run_data[:default_timing] = Run::REAL
+      if run_data[:realtime_duration_ms].zero? && run_data[:gametime_duration_ms].positive?
```
glacials: Unrelated to this branch, but we might want to convert this with `zero_to_nil` and check for `nil?` instead of `zero?`. Not sure if it will make any difference in behavior as I'm not sure if this value can ever be `nil` before we touch it, but it would be more consistent with the code above now that times (at least on our end) can be `nil`.

cc @jamacanbacn who noticed some runs with default timings of `realtime` when it felt like it should be `gametime`, perhaps this could be the cause.

app/models/concerns/unparsed_run.rb
glacials: Yeah see [`each_slice`][1] :D e.g.

```ruby
(0...run.attempt_history_len).each_slice(MAX_CONTAINER_SIZE) do |indexes|
  write_run_histories(indexes.map do |i|
    attempt = run.attempt_history_index(i)
    RunHistory.new(
      # ...
    )
  end)
end
```

[1]: https://apidock.com/ruby/Enumerable/each_slice

app/models/concerns/unparsed_run.rb
glacials: 👍 we might want to default this to something readable if we're not going to allow nulls in the db, like `segment_number` or `"???"` or something. Allowing blanks seems prone to UI bugs, e.g. a coaching situation where we say "You should improve your ."

I believe what we do for `run.user` is allow nulls but have `run.user.to_s` render as `"???"` so the UI still functions, but I wouldn't blame you if a migration for that felt out of scope for this PR.

app/models/concerns/unparsed_run.rb
glacials: 🤔 yeah I agree this is pretty hard to grok. For realtime/gametime stuff I tend to go with the "1, 2, n" philosophy of DRY code -- write it for realtime, then copy paste it for gametime. If we ever add a third (nth) timing mechanism for god knows what reason, we can worry about cleanly reusing code then.

app/models/concerns/unparsed_run.rb
glacials: I'm not sure if they _need_ to be, but I do think they should be just in case -- I don't see any situation where a segment was literally completed in zero time, but I do see situations where (a) livesplit-core reports zero times for absent segments or (b) it doesn't currently but will in the future, so afaict there's no harm in `zero_to_nil`ing it.

config/environments/profiling.rb
```diff
@@ -3,7 +3,7 @@
 Rails.application.configure do
   config.public_file_server.enabled = true
 
-  logger           = ActiveSupport::Logger.new(STDOUT)
+  logger           = ActiveSupport::Logger.new('log/profiling.log')
```
glacials: afaik we moved to stdout because we configured Docker (or Docker Compose?) to capture stdout and log it all to a file on the host, which has log rotation set up. Is that still the case?

db/schema.rb
glacials: I strongly prefer storing in Active Storage; I'm permanently scarred from storing non-text and/or variable-length stuff in the database after doing it early on in Splits.io. Even if livesplit-core guarantees some max size, it'll still be a long-term pain to fight against Rails's opinions on what a database field should be.

Active Storage also has some nice [convenience methods for displaying and transforming images][1]. Even if they're already shrunken to "not ridiculous size", we can use these to shrink them even more depending on how they're being displayed. It will also make it easier for us to let users modify images if we ever decide we want to.

But really, most of the benefit is from not needing to twist Rails's arm to behave sanely.

[1]: https://guides.rubyonrails.org/active_storage_overview.html#transforming-images

spec/models/run_spec.rb
```diff
@@ -118,33 +118,33 @@
...
-        [22, "Popcorn ceiling's revenge", 108_322],
+        [22, "Popcorn ceiling's revenge", 108_323],
         [23, 'Jump onto the flying staple', 112_977],
         [24, 'Ride motorbike for 9 seconds', 447_303],
-        [25, 'Final fight phase 1', 59_785],
-        [26, 'Final fight phase 2', 185_817],
-        [27, 'Final fight phase 3', 28_333]
+        [25, 'Final fight phase 1', 59_786],
+        [26, 'Final fight phase 2', 185_818],
+        [27, 'Final fight phase 3', 28_334]
```
glacials: What caused the slight difference in durations here?

app/models/concerns/unparsed_run.rb
glacials: Can you add a comment saying so here?

app/models/concerns/unparsed_run.rb
BatedUrGonnaDie: Do you think we should allow for nulls in the db for the name instead?  I don't really have a preference one way or the other, was just trying to keep things consistent with current behavior.

app/models/concerns/unparsed_run.rb
BatedUrGonnaDie: That's a good way of looking at it.  For now I'll put it back to repeating since with only 2 it's not that much.

spec/models/run_spec.rb
BatedUrGonnaDie: Previous parser added all the segment times on top of each other using the run's total time.  At the end we use that sum to be the full timing for the run, but that was incorrect if you didn't actually finish the run making the last finished split before resetting the final time.
New parser takes each segment and calculates the durations using only the previous segment, then gets the final time from the last segments time.  This is why the open struct was used, to give us 0's for the first segment.

TL;DR: adding all segments vs just looking at each one compared to the last.

app/models/concerns/unparsed_run.rb
glacials: I do think allowing nulls would be the cleanest, yeah. `null` represents not present, and that's the best description of what the user gave us for this field. Then override `Segment#name` to be like
```ruby
class Segment
  def name
    read_attribute(:name) || '???' # or '', or segment_number + 1, or 'Unnamed segment'
  end
end
```
but again, feel free to punt and hardcode something `'???'`-esque here in the parsing code if you feel that's out of scope for this PR.

app/models/concerns/unparsed_run.rb
```diff
@@ -4,163 +4,290 @@ module UnparsedRun
...
+      end
+      result
+    end
+
+    def process_segment_histories(lsc_segments, segment_ids)
+      container = []
+
+      lsc_segments.each_with_index do |seg, i|
+        seg.segment_history.iter.with do |iter|
+          until (history_element = iter.next).nil?
```
BatedUrGonnaDie: I wasn't sure how to use `each_slice` with this one.  Any idea/other ideas to clean this one up?

config/environments/profiling.rb
BatedUrGonnaDie: Yes, but when you run derailed stuff the output from rails basically removes all the information that it gives you.  I'm pretty sure you can have rails do log rotation as well if you think it would be beneficial.

app/models/concerns/unparsed_run.rb
BatedUrGonnaDie: Sounds good for the future.  @jamacanbacn if you have a link to those runs that were being classified incorrectly that would be nice to find the root cause.

db/schema.rb
BatedUrGonnaDie: This was the right call.  The new variant methods allowed very easy transformations of the icons.

config/environments/profiling.rb
glacials: Did we determine that it's required? Otherwise it will just keep increasing in size forever until it crashes the box right?

db/migrate/20190921232640_add_missing_run_fields_safety.rb
```diff
@@ -0,0 +1,7 @@
+class AddMissingRunFieldsSafety < ActiveRecord::Migration[6.0]
+  def change
+    change_column_null :run_histories, :pause_duration_ms, false
```
glacials: I think we might need this to allow nulls here for timers / formats that don't report pause durations on run histories (such as the exchange format!).

app/views/runs/_split_table.slim
```diff
@@ -3,6 +3,8 @@
   .table-responsive: table.card-body.table.mb-0
     thead: tr
       th.align-right #
+      - if run.segments.map(&:icon).any?(&:attached?)
```
glacials: 🚨 n+1 alert 🚨

app/views/runs/_split_table.slim
```diff
@@ -13,6 +15,8 @@
         - compare_segment = compare_segments.compact.min_by { |s| s.duration(timing) }
         tr
           td.align-right.align-middle = segment.segment_number + 1
+          - if run.segments.map(&:icon).any?(&:attached?)
```
glacials: 🚨same 🚨

app/models/concerns/unparsed_run.rb
glacials: Ah with these weird unidiomatic iterators I wouldn't worry about it too much.

app/models/concerns/unparsed_run.rb
```diff
@@ -4,163 +4,290 @@ module UnparsedRun
...
+    MAX_CONTAINER_SIZE = 1000
+
     def parsed?
       parsed_at.present?
     end
 
     def parse_into_db
       with_lock do
-        segments.delete_all
+        segments.each(&:destroy)
```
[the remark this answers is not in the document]
BatedUrGonnaDie: I forgot it existed, does that count?

db/migrate/20190921232640_add_missing_run_fields_safety.rb
BatedUrGonnaDie: The parser defaults this to 0, which I think is better than accepting null because then you can perform operations on pause time without needing null checks.  What do you think about that?

config/environments/profiling.rb
BatedUrGonnaDie: Well this only changes the profiling environment, which should never be ran long term on a production box, if ever.

app/views/runs/_split_table.slim
BatedUrGonnaDie: This should be covered by the includes on the run class that is being passed in.  Is it still N+1'ing?

glacials:
On second look my top-level comment might be unrelated to this branch, feel free to ignore for now and we can look at it independently later.

db/migrate/20190921232640_add_missing_run_fields_safety.rb
glacials: If I had the choice I'd choose for it to report null when pause time is not recorded (and 0 when it is but there were no pauses).

Similar situation to how we used to use 0 for absent run durations, segment durations, etc. -- it definitely makes the code easier for the reason you said, but that is a double-edged sword that also causes a lot of bugs that we've experienced over the years. e.g. summing a set containing 0 is fine, but averaging a set containing 0 throws the average way off. Charting 0 makes the chart wrong. And if we straight-up say "No pauses" on the run page, we'd be maybe lying.

But if the parser reports 0 for null then it sounds like we have no choice, so it's fine.

app/views/runs/_split_table.slim
glacials: Ah ok, I only ran the line in isolation from a Rails console. If it doesn't look like n+1 when loading the page, then disregard.

config/environments/profiling.rb
glacials: Ah, I missed the filename and thought this was for production! Nevermind then.

db/migrate/20190921232640_add_missing_run_fields_safety.rb
BatedUrGonnaDie: No no no, I'm defaulting it to be 0, I think LSC reports it as null if it doesn't exist.  Your argument makes a lot of sense, so I will stop defaulting to 0 and allow nulls in the db.

app/views/runs/_split_table.slim
BatedUrGonnaDie: Yeah, it's a bit misleading because I need to attach the includes to the Run class call, but that's in the controller I believe, so in the future this may become an N+1 if someone forgets it.

db/migrate/20190921232640_add_missing_run_fields_safety.rb
glacials: Ah ok gotcha, I thought you meant livesplit-core.

BatedUrGonnaDie:
Note to self: add dummy test file with blank split names that gets partially through a run then resets.

- [ ] Test split count is correct
- [ ] Test split times are accurate, but no run time is present

BatedUrGonnaDie:
Looked into the top level comment because I'm seeing the same behavior, `collapsed_segments(timing)` is returning wrong data.  Looks like its sending back the collapsed segments, and then the rest of the regular segments.  That function is extremely confusing to me, so I will file an issue for it so it can be revisited.

PR #614, opened 2019-09-24
Effort-to-PB chart should use RunHistory for the x-axis

### LINKED ISSUE #473: Change Effort-to-PB graph data source for playtime
Right now the Effort-to-PB graph uses a sum of segment histories to determine playtime at the moment a PB happens, e.g. to get total playtime that happened before attempt 452, it sums all segment histories (for all segments) that had an attempt number <= 452.

When segment histories are missing because the runner changed their splits or cleared their golds, this graph is nonsensical (has a vertical line at x=0). On top of that, it feels weird that missing segment history could make a graph not having anything to do with segments flip out.

We should switch this graph over to use `(endedAt - startedAt)` from the RunHistory entries, which are not as easily accidentally reset by housekeeping. As a bonus we don't need as much effort to calculate them. These values aren't passed through to the application from our livesplit-core wrapper, which means they are not passed from the API to JavaScript when it is building the graph. We need to add support for it in these three places to depend on them.

TODO:
- [ ] Find out when startedAt / endedAt were added, so we know if we're forsaking some runs by doing this
- [ ] Switch over if above is fine

moorecp:
Closes #473

The Effort-to-PB chart now uses `started_at` and `ended_at` from `RunHistory` to calculate playtime until a PB was reached in order to not bunch up attempts at time=0 if a user clears their splits.

I updated the API documentation, but I'm not sure how much of that is autogenerated. Additionally, I'm not sure if you want to add new fields to an existing API or not or wait until the next API version.

### DIFF
--- a/app/blueprints/api/v4/run_history_blueprint.rb
+++ b/app/blueprints/api/v4/run_history_blueprint.rb
@@ -1,6 +1,21 @@
 class Api::V4::RunHistoryBlueprint < Blueprinter::Base
-  fields :id, :attempt_number, :realtime_duration_ms, :gametime_duration_ms
+  fields :id, :attempt_number, :realtime_duration_ms, :gametime_duration_ms, :started_at, :ended_at
 
-  field(:started_at_ms) { |history, _| (history.started_at || 0).to_f * 1000 }
-  field(:ended_at_ms) { |history, _| (history.ended_at || 0).to_f * 1000 }
+  field(:started_at) do |history, _|
+    if history.started_at
+      history.started_at
+    else
+      DateTime.new(2013, 11, 13, 0, 0, 0)
+    end
+  end
+
+  field(:ended_at) do |history, _|
+    if history.ended_at
+      history.ended_at
+    elsif history.realtime_duration_ms
+      DateTime.new(2013, 11, 13, 0, 0, 0) + (history.realtime_duration_ms / 1000.0).seconds
+    else
+      DateTime.new(2013, 11, 13, 0, 0, 0)
+    end
+  end
 end
--- a/app/blueprints/api/v4/run_history_blueprint.rb
+++ b/app/blueprints/api/v4/run_history_blueprint.rb
@@ -1,21 +1,3 @@
 class Api::V4::RunHistoryBlueprint < Blueprinter::Base
   fields :id, :attempt_number, :realtime_duration_ms, :gametime_duration_ms, :started_at, :ended_at
-
-  field(:started_at) do |history, _|
-    if history.started_at
-      history.started_at
-    else
-      DateTime.new(2013, 11, 13, 0, 0, 0)
-    end
-  end
-
-  field(:ended_at) do |history, _|
-    if history.ended_at
-      history.ended_at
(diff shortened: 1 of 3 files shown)


moorecp:
From what I can tell, `started_at` and `ended_at` were added to `RunHistory` and merged into `master` on Dec 27, 2018 via #474. See commit ddb6011fe099779e75a17998c57ede9fdfbb4ee0.

BatedUrGonnaDie:
I believe the original issue meant when they were added to LiveSplit.  We reparse runs when new features or parser changes get added.

Also all api docs are currently made by hand, it's something we're hoping to automate in the future because of how hard it is to make sure everything is kept up to date.

app/blueprints/api/v4/run_history_blueprint.rb
```diff
@@ -1,3 +1,6 @@
 class Api::V4::RunHistoryBlueprint < Blueprinter::Base
   fields :id, :attempt_number, :realtime_duration_ms, :gametime_duration_ms
+
+  field(:started_at_ms) { |history, _| (history.started_at || 0).to_f * 1000 }
```
BatedUrGonnaDie: Do these procs take 2 or 3 arguments now?  I remember there being a depreciation notice from blueprinter about something related to this, but can't remember if it was just the `if:` procs.

moorecp: According to the [CHANGELOG](https://github.com/procore/blueprinter/blob/master/CHANGELOG.md#0180----20190529), that's only for `if/unless`. Thanks, though. I wouldn't have thought to have checked.

moorecp:
> I believe the original issue meant when they were added to LiveSplit.


It looks like it was added to the codebase on May 1, 2015 when they added the `Attempt` class that contained it. It was mentioned in the release notes for version 1.6.0, which was released on Aug 30, 2015
> Dates and times of attempts are now stored in the Attempt History in the Splits

app/blueprints/api/v4/run_history_blueprint.rb
glacials: > Also shouldn't these be DateTime's?

Yeah unless I'm missing something, it seems like `RunHistory#started_at`/`RunHistory#ended_at` should be rendered in the API more similar to how `created_at` and `updated_at` fields are.
```ruby
> RunHistory.last.started_at
=> Thu, 26 Apr 2018 22:34:12 UTC +00:00
> Segment.last.end_ms(Run::REAL)
=> 3185758
```
It's a little confusing because `Segment#start_ms`/`Segment#end_ms` do use milliseconds to represent the elapsed time _in that run_ at the moment the segment started/ended, but for `RunHistory` we use world time for run starts and ends (and in the API we always render world times as ISO 8601 timestamps).

moorecp: Yup, `DateTime` makes way more sense. The only thing I wasn't sure about is if it's possible to have NULL values in the DB for those fields. I set it as such in the API documentation and added appropriate checks in the JS, but it might not be necessary.

PR #625, opened 2019-10-06
Clear fileless runs, and add job to delete new ones after an hour

BatedUrGonnaDie:
I'm not sure why the GitHub actions are failing, since it works perfectly fine locally and the error doesn't make a lot of sense to me.

### DIFF
--- a/db/migrate/20191006212943_cleanup_runs_with_no_files.rb
+++ b/db/migrate/20191006212943_cleanup_runs_with_no_files.rb
@@ -1,6 +1,6 @@
 class CleanupRunsWithNoFiles < ActiveRecord::Migration[6.0]
   def up
-    Run.find_each do |run|
+    Run.where(parsed_at: nil).find_each do |run|
       file = $s3_bucket_internal.object("splits/#{run.s3_filename}")
       next if file.exists?
 
--- a/db/migrate/20191006212943_cleanup_runs_with_no_files.rb
+++ b/db/migrate/20191006212943_cleanup_runs_with_no_files.rb
@@ -1,0 +1,10 @@
+class CleanupRunsWithNoFiles < ActiveRecord::Migration[6.0]
+  def up
+    Run.find_each do |run|
+      file = $s3_bucket_internal.object("splits/#{run.s3_filename}")
+      next if file.exists?
+
+      run.destroy
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


db/migrate/20191006212943_cleanup_runs_with_no_files.rb
```diff
@@ -0,0 +1,10 @@
+class CleanupRunsWithNoFiles < ActiveRecord::Migration[6.0]
+  def up
+    Run.find_each do |run|
+      file = $s3_bucket_internal.object("splits/#{run.s3_filename}")
+      next if file.exists?
+
+      run.destroy
+    end
```
glacials: triple-check that this method and `.exists?` work as expected! one small line preventing every run from being deleted

in fact out of an overabundance of caution, we might want to do `Run.where(parsed_at: nil).find_each` instead in case there happen to be runs out there that successfully parsed but for one reason or another have no file.

BatedUrGonnaDie: Do we have any runs that do not have a file backing them?  I thought even when uploading as exchange we saved the JSON payload so that every run would have a file backing it.

glacials: Yeah as far as I know we do not have any runs without a file backing them. But since this code is one malformed S3 response away from deleting our entire database of runs, I just mean let's add that condition for extra extra safety.

BatedUrGonnaDie: Fair enough.  I was just trying to make sure the next time we do a full reparse we would be confident that all runs are backed by a file and ready to go.  Fairly certain we're going to need to do a full reparse in the near-ish future and wasn't sure if any missing files would break things.

For now I'll add the extra condition and do a triple check on `.exists?` and we can worry about the full reparse when we get there.

glacials: Gotcha, didn't realize you were preparing for a full reparse. Keep in mind it's still possible we run into runs-without-files if we run the full reparse during the 1-hour grace period between someone's create and upload.

BatedUrGonnaDie: Honestly I keep forgetting about that, so it makes even less sense to prep by going through all runs.

Yeah, with the parser changes getting icons and pause times, and the plans for other features to leverage runs from other people I feel like a full reparse will be needed in the next month or so.

BatedUrGonnaDie:
Triple checked and scoped, merging.

PR #633, opened 2019-10-15
Use fuzzy matching for timeline comparisons when there are a differing number of segments

### LINKED ISSUE #619: Skipping a split makes comparison aligns off
When you skip a split, we roll it into the following segment for display purposes, renaming it `"<segment1-name> + <segment2-name>"`.

Now that we have comparisons lining up to each other, an existing issue comes to light where one timeline thinks its splits are one ahead of the other's.

See an example here: https://splits.io/4nqv?compare=4nch
- The top timeline skipped "Darkness 60 Mats", the 9th split
- Therefore hovering over any segment after that one, in either timeline, causes the other timeline to align to the wrong split

This also exposes itself in having the split colors be off from each other.

moorecp:
Closes #619 

This uses fuzzy matching to line up timelines when there are differing numbers of splits between timelines. It's currently holding the position of the timelines when mousing over a segment without a good enough match as opposed to returning to the origin to cut down on timelines snapping about.

It doesn't currently display any sort of message on a non-match like @BatedUrGonnaDie talked about. Clearly that can happen; I wanted to get this in front of people first and then iterate if needed.

### DIFF
--- a/app/views/runs/_timeline.slim
+++ b/app/views/runs/_timeline.slim
@@ -4,12 +4,13 @@
 - else
   - if run.video&.supports_embedding?
     div style="height: 1em": span.video-progress-line.timeline-animation id="video-progress-line-#{run.id36}"
-  .timeline-background.timeline-animation data={run_id: run.id36, segment_names: run.collapsed_segments(timing).map(&:name)}
+  - collapsed_segments = run.collapsed_segments(timing)
+  .timeline-background.timeline-animation data={run_id: run.id36, segment_names: collapsed_segments.map(&:display_name), total_segments: collapsed_segments.length}
     .timeline.shadow style="width: #{run.duration(timing) / scale_to * 100.0}%"
-      - run.collapsed_segments(timing).each.with_index do |segment, index|
+      - collapsed_segments.each.with_index do |segment, index|
         .pure-u.split(
           id="#{run.id36}-split-#{index}"
-          data={start_ms: segment.start(timing).to_ms, run_id: run.id36, segment_name: segment.name, segment_number: index+1}
+          data={start_ms: segment.start(timing).to_ms, run_id: run.id36, segment_name: segment.display_name, segment_number: index+1, total_segments: collapsed_segments.length}
           class="#{next_timeline_color(run.id36)} #{'progress-bar-striped progress-bar-animated' if segment.duration(timing).nil?}"
           style="width: #{segment.proportion(timing) * 100.0}%; z-index: #{index}; #{'cursor: pointer;' if run.video}"
         )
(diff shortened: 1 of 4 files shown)


app/views/runs/_timeline.slim
```diff
@@ -4,12 +4,13 @@
 - else
   - if run.video&.supports_embedding?
     div style="height: 1em": span.video-progress-line.timeline-animation id="video-progress-line-#{run.id36}"
-  .timeline-background.timeline-animation data={run_id: run.id36}
+  - collapsed_segments = run.collapsed_segments(timing)
```
glacials: Is there a benefit to caching this manually in a variable here vs. letting Rails cache the query for us?

app/views/runs/_timeline.slim
```diff
@@ -4,12 +4,13 @@
...
+  .timeline-background.timeline-animation data={run_id: run.id36, segment_names: collapsed_segments.map(&:display_name), total_segments: collapsed_segments.length}
     .timeline.shadow style="width: #{run.duration(timing) / scale_to * 100.0}%"
-      - run.collapsed_segments(timing).each.with_index do |segment, index|
+      - collapsed_segments.each.with_index do |segment, index|
         .pure-u.split(
           id="#{run.id36}-split-#{index}"
-          data={start_ms: segment.start(timing).to_ms, run_id: run.id36, segment_number: index+1}
+          data={start_ms: segment.start(timing).to_ms, run_id: run.id36, segment_name: segment.display_name, segment_number: index+1, total_segments: collapsed_segments.length}
```
glacials: Seems like we're duplicating information here:
- The array of segment names is stored on `.timeline`, but it can be fetched with 
    ```javascript
    Array.from(timeline.getElementsByClassName('split')).map(el => el.dataset.segment_name)
    ```
- The number of segments is being stored on both `.timeline` and every `.split`, but it can be fetched with
    ```javascript
    timeline.getElementsByClassName('split').length
    ```

app/views/runs/_timeline.slim
moorecp: `ForgetfulPersonsRun.collapsed_segments` does do a non-zero amount of work apart from just a SQL query. Since I was adding additional calls to it, it seemed sensible to go ahead and just cache the result. It does look like some of those calls will be going away, though.

glacials: Oh that's right, I forgot that method is not just a query!

app/views/runs/_timeline.slim
moorecp: Sounds good. I hate redoing work over and over, but storing all the names and counts multiple times is also bad.

app/views/runs/_timeline.slim
moorecp: 2 of the uses of `collapsed_segments` were removed leaving just 2. I'm fine going back to 2 calls of the original if you'd prefer. The performance difference is probably negligible.

BatedUrGonnaDie:
In my head it was basically a tippy style message that would just fade after a second or 2, but not having one is fine.  If we do end up getting a lot of feedback about it then it can always be added later.

app/views/runs/_timeline.slim
glacials: Yeah I understand the struggle. In general we prefer extra work over extra storage on Splits.io, if it's not clear-cut which is better.

PR #643, opened 2019-10-26
Add splitting to races

glacials:
Adds a Split button to races. This button will split if there is at least one segment left in the run, or create a segment then split if not. It will also create a run if one isn't yet attached to the race entry. These two features together can be used to build splits adhoc during a blind or semi-blind races.

The API allows you to choose whether to enable this smart behavior by passing `more=1`. When not included, the API will only split in the classical sense, completing the run if you're on the last segment.

Includes several other race improvements:

- Adds timelines to races for each runner splitting
    Screen Shot 2019-10-25 at 16 52 13
- Timelines (both in & out of races) now have smarter segment widths for in-progress runs (they were previously all equal widths)
    localhost_3000_54
- Adds a "Syncing" spinner to the race page that shows whenever you've performed an action but it hasn't reflected on the page yet due to the worker and/or ActionCable taking an extra bit
    Screen Shot 2019-10-25 at 16 54 18
- The race page is now more aggressive about hiding itself (and showing a spinner) while loading / while navigating away from the page, rather than showing outdated information cached by Turbolinks
- The Create Race button no longer shows an error symbol after being clicked & transitioning to the new race page, even though there was no error
- Can no longer start editing a race after it has started (previously you could start, but not save the changes)

This may change after https://github.com/LiveSplit/livesplit-core/issues/260 is addressed. However I don't want to block on it.

Although the timeline & segment width calculations for in-progress runs is better, it still really sucks when viewing multiple in-progress timelines together & would love suggestions. I can't find a way to pick a `scale_to` value that makes sense before the race is over.  Segments from different runners that have wildly different durations can be the same or similar width, or vice-versa, based on which segment they're on & how they did on other segments.

cc @jhobz


jhobz:
Wow, that was a fast implementation! Looks great from the screenshots. Is this somewhere on beta atm? If so, I can use it with a real race on Friday to get some real-life testing on it.

glacials:
@jhobz: Cool, yeah I've just put this on bet.splits.io. Feel free to use it, however I'm about to hop on a plane to BlizzCon so I can't help with any issues until next week.

app/controllers/api/v4/runs/splits_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+        status: 400,
+        message: @run.errors.full_messages.to_sentence,
+      }
+    end
+  end
+
+  private
+
+  def split_params
+    params.require(:split).permit(
```
BatedUrGonnaDie: Does `params.require` gracefully fix it's error?  If not then we should catch it and then return a 4XX over a 500.

app/models/run.rb
```diff
@@ -255,6 +255,170 @@ def segments_with_groups
     @segments_with_groups
   end
 
+  # split tries to split this run by filling in the first unfilled segment with a time. If more is true and this split
+  # would cause the run to complete, a new segment will be created at the end of the run and left blank for the next
+  # split to fill. This can be used to perform blind runs where the runner doesn't know how many splits there will be.
+  # Just make sure to eventually call this with more=false in order to actually finish the run.
+  def split(more: false, realtime_end_ms:, gametime_end_ms:)
```
BatedUrGonnaDie: This function is going to throw a bunch of rubocop warnings.

app/javascript/vue/race-nav.js
```diff
@@ -88,6 +89,57 @@ export default {
         this.loading.ready = false
       }
     },
+    setSyncing: async function() {
```
BatedUrGonnaDie: We're defining this in multiple places, is there any way to have it be shared?

app/models/run.rb
glacials: Damn wow you were right. Just pushed a commit that fixed a lot of Rubocop stuff, but this function and the controller function are still too long,  too high AbcSize, and too high complexity. For now I put a Rubocop TODO comment on it, I'm about to get on a plane and will think about if refactoring should be done now or punted on so we can unblock other things first.

glacials:
^ This commit piggybacks the pricing overhaul we've been talking about offline onto this branch, since they're launching together. Re-requesting review for this stuff. (Sorry, I rebased onto master instead of merging, without thinking about it)

app/models/user.rb
```diff
@@ -112,9 +116,30 @@ def has_advanced_comparisons?
...
+    patron?(tier: 1, before: STRIPE_MIGRATION_DATE) || subscriptions.tier1.active.any? || admin?
+  end
+
+  def has_advanced_analytics?
+    patron?(tier: 1, before: STRIPE_MIGRATION_DATE) || subscriptions.tier1.active.any? || admin?
+  end
+
+  # patron? returns something truthy if the user has been a patron at or above the given tier since the given `before`
+  # time. Not passing a `before` is the same as not caring how long the user has been a patron.
+  def patron?(tier: 0, before: Time.now.utc)
```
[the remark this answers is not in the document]
glacials: It works fine on console, is there something different about how it works in a full Rails app?

```ruby
> def f(t: Time.now.utc)
  t
end
=> :f
> f
=> 2019-11-25 21:24:23 UTC
> f
=> 2019-11-25 21:24:25 UTC
> f
=> 2019-11-25 21:24:27 UTC
```

app/models/segment.rb
```diff
@@ -32,9 +32,11 @@ def self.with_ends
   # Returns the Segment the runner would be on at the given run time. A run time is a Duration since the start of the
   # run.
   #
-  # If the Duration takes place outside the run (i.e. is greater than Run#duration), nil is returned.
+  # If the Duration takes place outside the run (i.e. is greater than Run#duration), the last segment is returned. This
+  # is to support blind runs where the number of segments is not known, and new segments can be added on to the end of
+  # the run as they are reached.
   def self.at_run_time(run_time)
-    with_ends.order(segment_number: :asc).where('realtime_end_ms > ?', run_time.to_ms).first
+    with_ends.order(segment_number: :asc).where('realtime_end_ms > ?', run_time.to_ms).first || last
```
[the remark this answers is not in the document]
glacials: Took this comment for  me to realize we're actually not using this function anywhere anymore, I think since we replaced the "Current split" column of the race entries table with a full timeline. I think I made this change so that runners who were finished would still show their final split rather than nothing, before replacing the whole thing with a timeline.

I'll undo the change, might keep the function around though just because it could come in handy for auto-coaching.

app/models/user.rb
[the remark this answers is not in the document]
BatedUrGonnaDie: I was under the impression that default args were evaluated once when the function was defined, but that might not be covered.  There are some really weird rules related to date times in ruby and when they need special handling to be dynamic.

PR #646, opened 2019-11-04
Don't show comparison timeline as In Progress if it doesn't use the selected timing method

### LINKED ISSUE #615: Comparisons can make splits appear in-progress
Sometimes a comparison might make one run appear incorrectly as in-progress, e.g.

Completed run A (correct): https://splits.io/262b
Completed run B (correct): https://splits.io/oqt

Comparing B to A (correct): https://splits.io/oqt?compare=262b
Comparing A to B (incorrect - B appears in-progress): https://splits.io/262b?compare=oqt

moorecp:
Closes #615 

#615 was caused by the logic checking to see if `This run doesn't have #{timing}time splits.` should be displayed as opposed to showing the timeline. It was checking if a run had a duration for the current timing method and if that run was not complete. If a run had no duration for that timing method, it just fell through and all splits showed as In progress.

This was introduced with #550 and racing related stuff. My fix might not be the right thing to do based on racing. I'm actually not sure why `!run.completed?(timing)` is a reason for displaying `This run doesn't have #{timing}time splits.`. I would think `run.duration(timing).blank?` would be all you would want and then a segment would display as `In progress` based on the later check of `segment.duration(timing).present?`, but I have no experience with the racing side of the site.


app/views/runs/_timeline.slim
```diff
@@ -1,5 +1,5 @@
 - scale_to ||= run.duration(timing).to_ms
-- if run.duration(timing).present? && !run.completed?(timing)
+- if run.duration(timing).blank? || !run.completed?(timing)
```
glacials: On second look this line seems like it was always redundant (or contradictory), as `Run#completed?` checks `duration(timing).present?` also.

moorecp:
What should be displayed when comparing 2 runs where 1 is only RTA and complete and the other is only IGT and in progress? Based on your 2 conditions up there, the second run should be displayed as In Progress even though the timelines can't be sensibly compared.

It seems like `This run doesn't have #{timing}time splits.` is a valid thing to display unless the run actually has at least 1 split of `#{timing}time`. If you get past that check, then you can get into the Complete vs In Progress checks.

glacials:
Very good point, I agree!

PR #648, opened 2019-11-08
Upgrade to Bootstrap 4.3, fresh look

glacials:
Tired of waiting for our old theme to catch up to the modern Bootstrap version (2 years out of date atm), so I switched us to the main Bootstrap branch and updated us to 4.3.

In doing so we lose all our styling information, so I built us our own dark mode by modifying a few key Bootstrap variables. Rather than switching to another theme that will eventually fall out of date, this will let us stay on mainline Bootstrap indefinitely.

localhost_3000_56
localhost_3000_56 (1)
localhost_3000_search_q=sunshine


app/views/games/categories/_title.slim
```diff
@@ -1,4 +1,4 @@
-.card.mb-3
+.card.mb-3.bg-darkish
```
BatedUrGonnaDie: Is every single card going to have `.bg-darkish`?  If so is there a way to just set this once so we don't need the remember to add it everytime?

glacials: I thought I'd have to do some weird SASS `@extend`ing that wouldn't be friendly to upgrades, but this prompted me to do some experimenting and I found an apparently undocumented variable that works great!

PR #651, opened 2019-12-05
Tweak design to be more mobile-friendly

glacials:
Tweaks a lot of the site design to be more friendly when viewing on mobile. On the whole, the desktop experience should be unchanged except for a couple of sidegrades like the run table.

There's still some more I want to do, including probably replicate what I did to the run table to all other similar tables, but this is a good cut-off point for this branch both in terms of reviewability and ability to go to production on its own before it gets too big.

**Notes**:
- I also added Speedrun.com & SRL sync buttons to the game edit page (admins only) as a temporary measure, since the new run table uses game cover images it's exposed that a lot of our cover images use broken URLs that get fixed on a re-sync. (mostly due to case changes like `Portal` -> `portal`, I guess Speedrun.com changed their cover URLs at some point)
- You can't have an `a` tag inside an `a` tag, so for the linkified `list-group-item`s in the new run table I had to ditch the user badge, the delete/disown menu, and the rewind button. I added the rewind button to the run show page, and the others are accessible from there as well.

Couple examples of visual difference, though a lot of the changes are in usability as well:

<table><tr><td>Old</td><td>New</td><tr><td>

</td><td>

</td></tr><tr><td>

</td><td>

</td></tr><tr><td>

</td><td>

</td></tr><tr><td>

</td><td>

</td>

</tr></table>

### DIFF
--- a/app/javascript/packs/application.js
+++ b/app/javascript/packs/application.js
@@ -7,11 +7,6 @@
 // To reference this file, add <%= javascript_pack_tag 'application' %> to the appropriate
 // layout file, like app/views/layouts/application.html.erb
 
-import 'turbolinks-animate'
-document.addEventListener('turbolinks:load', function() {
-  TurbolinksAnimate.init({element: document.querySelector('body.turbolinks-animate'), animation: 'fadeinup'})
-})
-
 import "jquery"
 require("@rails/activestorage").start()
 require("@rails/ujs").start()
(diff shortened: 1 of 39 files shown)


app/javascript/packs/application.js
```diff
@@ -7,6 +7,11 @@
 // To reference this file, add <%= javascript_pack_tag 'application' %> to the appropriate
 // layout file, like app/views/layouts/application.html.erb
 
+import 'turbolinks-animate'
+document.addEventListener('turbolinks:load', function() {
+  TurbolinksAnimate.init({element: document.querySelector('body.turbolinks-animate'), animation: 'fadeinup'})
```
BatedUrGonnaDie: This is going to call init on every single page load, is this intended behavior?

glacials: Oops, not sure but this was leftover code from a library I ended up not using. Removed!

PR #653, opened 2019-12-11
Add affiliate links

glacials:
Adds an affiliate links feature to support upcoming partners. Navigate to splits.io/test, get redirected to the subscriptions page with a special code. If you sign up, the referring partner (test) will get a kickback.

This works by applying a Stripe Coupon, which will be tied 1-to-1 to a partner. We can later query against coupons to determine payouts.

Stripe Checkout, the pre-baked hosted checkout flow we use, apparently doesn't yet support Coupons. So we have to go through the Checkout flow without it then apply the Coupon afterwards.

This is decent if the Coupon isn't actually applying a meaningful discount (right now it's the minimum, $0.01), but if we long-term want to give users discounts as an incentive to use affiliate links, they'll still see full price in the Checkout flow (and be charged full price for the first month) so we'd probably want to find a better solution. Mostly hoping Stripe Checkout implements Coupon support before then, because I'm not all about building our own forms to take in credit card numbers.

Separately from that, using Coupons for this at all is very much MVP to test the idea; payouts will be manual vs. a solution like Stripe Connect (which _does_ integrate with Checkout, but we have to be accepted into the program) or a third party platform that integrates with Stripe to provide an affiliates product.

### DIFF
--- a/app/controllers/subscriptions_controller.rb
+++ b/app/controllers/subscriptions_controller.rb
@@ -1,10 +1,17 @@
 class SubscriptionsController < ApplicationController
   before_action :set_subscription, only: %i[update destroy]
 
+
   def index
   end
 
   def show
+    if params[:coupon]
+      Stripe::Subscription.update(
+        Subscription.find_by(stripe_session_id: params[:session_id]).stripe_subscription_id,
+        {coupon: params[:coupon]}
+      )
+    end
   end
 
   def update
--- a/app/javascript/stripe.js
+++ b/app/javascript/stripe.js
@@ -6,10 +6,12 @@ document.addEventListener('click', async function(event) {
 
   event.preventDefault()
 
+  const coupon = new URL(document.URL).searchParams.get('coupon')
+
   const stripe = Stripe(process.env.STRIPE_PUBLISHABLE_KEY)
   const result = await stripe.redirectToCheckout({
     items: [{plan: subscribeButton.dataset.planId, quantity: 1}],
-    successUrl: `${location.origin}/subscriptions/success?session_id={CHECKOUT_SESSION_ID}`,
+    successUrl: `${location.origin}/subscriptions/success?session_id={CHECKOUT_SESSION_ID}&coupon=${coupon}`,
     cancelUrl: `${location.origin}/subscriptions/cancel`,
     clientReferenceId: gon.user.id,
     customerEmail: gon.user.email,
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -114,6 +114,13 @@ Rails.application.routes.draw do
   delete '/settings/applications/:application',      to: 'applications#destroy'
   delete '/settings/authorizations/:application',    to: 'authorized_applications#destroy',                      as: :authorization
 
+  # partners is a map of partner special URLs to Stripe coupon IDs
+  {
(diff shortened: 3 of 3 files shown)


jhobz:
When you say you're using the "pre-baked hosted checkout flow," do you mean the version Stripe hosts, or the version that's built-in to pages after an action (like clicking a "checkout with Stripe" button)? I have a little experience with Stripe, and for the method we use with RPGLB's site, the payments db table is hosted on our end, so any discounts could theoretically be saved to the user's data model and just applied during the next checkout (or the next subscription period) automatically. I'm unfamiliar with Stripe Coupons, but is it maybe an unnecessary step in this situation?

glacials:
We're using the Stripe-hosted page that you get redirected to; if you're talking about the in-page modal, that is Stripe Checkout v1 and is considered legacy by Stripe now. However there are two ways to integrate with the new Checkout, a client-only easy way and a server-side more powerful way, and we're using the former, so maybe I should look into the latter.

None of our payments info is stored on-site, we just listen to Stripe webhooks for subscriptions getting activated or cancelled. They handle the rest in-between.

jhobz:
Ah, ok, yeah we use Stripe Checkout v1 then and my experience there probably won't be of much use. Sorry for the misdirection.

config/routes.rb
```diff
@@ -114,6 +114,13 @@
   delete '/settings/applications/:application',      to: 'applications#destroy'
   delete '/settings/authorizations/:application',    to: 'authorized_applications#destroy',                      as: :authorization
 
+  # partners is a map of partner special URLs to Stripe coupon IDs
```
BatedUrGonnaDie: Putting the partners hash in the routes file seems a bit weird, hard to know where to look for it in the future.  What do you think about putting it in something like an initializer instead?

app/javascript/stripe.js
```diff
@@ -6,10 +6,12 @@ document.addEventListener('click', async function(event) {
 
   event.preventDefault()
 
+  const coupon = new URL(document.URL).searchParams.get('coupon')
+
   const stripe = Stripe(process.env.STRIPE_PUBLISHABLE_KEY)
   const result = await stripe.redirectToCheckout({
     items: [{plan: subscribeButton.dataset.planId, quantity: 1}],
-    successUrl: `${location.origin}/subscriptions/success?session_id={CHECKOUT_SESSION_ID}`,
+    successUrl: `${location.origin}/subscriptions/success?session_id={CHECKOUT_SESSION_ID}&coupon=${coupon}`,
```
BatedUrGonnaDie: This will still function correctly with no parameter, yeah?

app/controllers/subscriptions_controller.rb
```diff
@@ -1,10 +1,17 @@
...
 
+
   def index
   end
 
   def show
+    if params[:coupon]
+      Stripe::Subscription.update(
+        Subscription.find_by(stripe_session_id: params[:session_id]).stripe_subscription_id,
+        {coupon: params[:coupon]}
```
BatedUrGonnaDie: How come this is in curly braces?

glacials:
fyi this is on hold as we just got accepted into Stripe Connect, so that seems like a much more proper way of doing this and being able to track (and automate) payouts.

PR #660, opened 2020-01-12
Add automatic submitting to SRDC

### LINKED ISSUE #127: Autosubmit times to speedrun.com

BatedUrGonnaDie:
Adds a stream-lined (and in most cases automatic) upload process for getting your run onto SRDC.  Assuming you set the variables in the splits file, it will pre-populate all required fields for you, and you should be able to double check and hit submit.

Non-exhaustive changelog:
- Adds new rake task that will sync all games, categories, platforms, regions, and variables from SRDC.
- Properly saves run variables to the db now, instead of parsing and throwing them away
- Allows SRDC ID to be cleared from a run
- Removes old sync code based around runs being uploaded

Things punted to a later date:
- Extracting SRDC ID to its own table like video (I would like to namespace all the SRDC models, and will be taken as part of that)
- Optimizing the import process (with how the relations are structured, I'm not confident this will be super easy and this has already been in the works for almost 2 full months.)

Closes #127


BatedUrGonnaDie:
@glacials I believe you also still need to add in the upsell information for anyone who is not a supporter.

app/controllers/runs_controller.rb
```diff
@@ -56,11 +56,7 @@ def update
 
     if params[@run.id36][:srdc_url]
       srdc_id = SpeedrunDotCom::Run.id_from_url(params[@run.id36][:srdc_url])
-      if !srdc_id
-        redirect_params = {
-          alert: 'Your Speedrun.com URL must have the format https://www.speedrun.com/tfoc/run/6yjoqgzd.'
-        }
```
glacials: Now when I enter `6yjoqgzd` into this field, my Speedrun.com link gets set to `https://www.speedrun.com/run/f`.

app/models/concerns/unparsed_run.rb
```diff
@@ -58,10 +59,13 @@ def parse_into_db
...
           default_timing:          run_data[:default_timing],
-          filesize_bytes:          run_data[:size]
+          filesize_bytes:          run_data[:size],
+
+          platform:                SpeedrunDotComPlatform.find_by(name: run_data[:metadata][:platform_name]),
+          region:                  SpeedrunDotComRegion.find_by(name: run_data[:metadata][:region_name])
         )
 
-        HighlightSuggestion.from_run(self)
+        # HighlightSuggestion.from_run(self)
```
glacials: Was this left accidentally?

BatedUrGonnaDie: Yep, probably throwing an error so I fixed it the old fashioned way.

app/controllers/runs_controller.rb
BatedUrGonnaDie: Looks like it was saving the string representation of false.  Should be fixed now good catch.

PR #665, opened 2020-01-17
Make game/category selectors Vue components

glacials:
Migrates the three areas on the site where you can autocomplete a game/category -- search, edit run, and edit race -- to Vue components with a decent amount of reusability.

This should incidentally fix the bug where the search bar gets ludicrously tall, since we're not powering it with that old library anymore.


app/views/layouts/application.slim
```diff
@@ -18,147 +18,146 @@ html
     = render partial: 'application/favicon'
     = render partial: 'shared/twitter_card'
     = content_for(:head)
-  body data-controller=controller_name data-action=action_name
     = Gon::Base.render_data
+  body data-controller=controller_name data-action=action_name
```
glacials: Do yourself a favor and turn off whitespace changes in the options menu at the top of GitHub when reviewing this file

app/controllers/runs/likes_controller.rb
```diff
@@ -25,12 +25,12 @@ def destroy
...
 
   def set_run
-    @run = Run.find_by(id: params[:run].to_i(36))
+    @run = Run.find_by(id: params[:run_id].to_i(36))
     head :not_found if @run.nil?
   end
 
   def set_run_like
-    @run_like = RunLike.find_by(user: current_user, run: Run.find_by(id: params[:run].to_i(36)))
+    @run_like = RunLike.find_by(user: current_user, run: Run.find_by(id: params[:run_id].to_i(36)))
```
glacials: I made these changes because I've been wanting to revert us back to conventional Rails style for `params` for a while, and the rest of this branch got easier if I went ahead and did parts of it now.

wooferzfg: Is there a reason you can't do this here:
```
Run.find(params[:run_id].to_i(36))
```

BatedUrGonnaDie: `.find` throws an error vs `.find_by` returning nil on no record found.

app/javascript/vue/Search.vue
```diff
@@ -0,0 +1,70 @@
...
+      gameResults: [],
+      input: '',
+      loading: false,
+      selectedItem: null,
+      userResults: [],
+    }
+  },
+  methods: {
+    debouncedFetch: _.debounce(async function(newQuery) {
+      this.gameResults = (await fetch(`/api/v4/games?search=${newQuery}`).then(response => response.json())).games
```
BatedUrGonnaDie: Should these be `await`'d since they seem like they can be done in parallel?

app/javascript/vue/race-title.js
```diff
@@ -94,7 +88,7 @@ export default {
           method: 'PATCH',
           headers: headers,
           body: JSON.stringify({
-            category_id: this.categoryId,
+            category_id: (this.category && this.category.id) || null,
```
BatedUrGonnaDie: Do we know if we can do the safe navigation operator yet?

app/views/layouts/application.slim
BatedUrGonnaDie: Can you double check that this does not impact the contents of the global `gon` variable across page loads?  I remember something with turbolinks wanting this to not be in the head tag.

app/views/races/_create.slim
```diff
@@ -1,7 +1,7 @@
 - game = local_assigns.fetch(:game, Game.new)
 - category = local_assigns.fetch(:category, Category.new)
 
-// Make sure to have #vue-race be defined before calling this partial
+// Make sure to have #vue defined before calling this partial
```
BatedUrGonnaDie: Is this comment relevant now that `vue` is defined in `application.slim`?

config/routes.rb
```diff
@@ -84,9 +84,9 @@
   post   '/rivalries', to: 'rivalries#create', as: :rivalries
   delete '/rivalries', to: 'rivalries#destroy'
 
-  get   '/games/:game',      to: 'games#show',  as: :game
-  get   '/games/:game/edit', to: 'games#edit',  as: :edit_game
-  patch '/games/:game',      to: 'games#update'
+  get   '/games/:game',      to: 'games#show',  as: :game, format: false # disable format so requests like /ads.txt don't render a run
```
BatedUrGonnaDie: Can we update this comment to be a bit more accurate?

app/javascript/vue/race-title.js
glacials: Not yet -- we could [manually enable][1] "experimental support" for it, but I'd rather wait for full support which is supposed to be this year anyway: https://github.com/tc39/proposals/blob/master/finished-proposals.md

[1]: https://github.com/babel/babel/tree/master/packages/babel-plugin-proposal-optional-chaining

app/views/layouts/application.slim
glacials: Ah crap yeah you're right. It technically works but it's just appending more `window.gon=...` calls to `<head>`, growing the page source with every load. Moved back.

app/views/races/_create.slim
glacials: Good point, no it is not!

app/javascript/vue/race-title.js
BatedUrGonnaDie: I agree, waiting for full support sounds like the best option.

PR #672, opened 2020-01-30
Remove alpine from the web/worker/webpacker images

glacials:
Stops using the alpine images for Ruby since they've been causing various small issues over time. I also decided to combine the prod and dev Dockerfiles into one while I was at it. Not quite sure I understand the builder vs. runner thing, so tell me if I messed that up somehow, but atm this boots fine in development.


BatedUrGonnaDie:
Do we want to stick with one dockerfile going forward?  If we want to condense down to one dockerfile then I think we should drop the builder vs runner.  It doesn't make a whole lot of sense with a one file set up as the bonus is a smaller final image at the cost of cutting **everything** that isn't 100% essential, and in a one file setup we want to be able to compile the gems and npm packages native extensions.

PR #674, opened 2020-02-07
Add an autosplitter badge

glacials:
Adds a badge to runs that were performed using an autosplitter:

Screen Shot 2020-02-07 at 13 04 30

I had trouble with the tooltip even after hard loads, so I dove a bit into solving them globally -- now that our whole site is in a Vue app container, even if we're not using Vue in a given place we can still take advantage of the Vue tippy tooltips, which show consistently regardless of soft/hard loads, so I moved the surrounding tooltips over to that style.

### DIFF
--- a/app/views/runs/_timing_badge.slim
+++ b/app/views/runs/_timing_badge.slim
@@ -3,7 +3,7 @@
     a.badge.badge-dark(
       href=run_path(run, params.permit(:compare).merge(timing: 'game'))
       v-tippy=true
-      :title='"Recorded in realtime & gametime—click to switch"'
+      title='Recorded in realtime & gametime—click to switch'
     )
       span.text-success => icon('fas', 'globe-americas')
       = icon('fas', 'gamepad')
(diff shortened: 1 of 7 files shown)


CryZe:
I believe using the auto splitter settings is potentially going to cause true positives and false negatives. You may reuse the same splits for a different game and the settings might not get cleared out. And you may use an auto splitter that doesn't have settings.

glacials:
Hm, is there any better way of detecting autosplitters?

CryZe:
Not really. But detecting the auto splitter settings should work in a lot of cases.

app/views/runs/_timing_badge.slim
```diff
@@ -1,20 +1,24 @@
 - if run.completed?(Run::REAL) && run.completed?(Run::GAME)
   - if timing == Run::REAL
-    a.badge.badge-dark.tip(
-      title='Recorded in realtime & gametime — click to switch'
+    a.badge.badge-dark(
       href=run_path(run, params.permit(:compare).merge(timing: 'game'))
+      v-tippy=true
+      :title='"Recorded in realtime & gametime—click to switch"'
```
BatedUrGonnaDie: I believe if you don't put the colon before the attribute you won't have to double wrap the strings.  From my understanding the colon tells view to interpret it as JS first.

glacials: Could have sworn `title` was a special case with this, but now that I try it it works fine and I can't find anything in the docs about it not working. ¯\_(ツ)_/¯

glacials:
@CryZe Alright, I'll go ahead and merge this and accept that there will be some false positives/negatives. Seems like the common case should be fine, but I'll keep an ear out and remove if it's causing too much confusion. Thanks for the input!

PR #694, opened 2020-05-08
Swap ubuntu back in for alpine

BatedUrGonnaDie:
This should unblock the builds.  Requires a pipeline change before merging this in.

I ended up re-using the old `splitsio-build` code deploy project from before, and modified it with some enhancements we picked up with the alpine builds.  Overall we lose a couple of minutes and the container size is a lot larger, but this should solve almost if not all of the compile and dependency issues that we have been seeing with alpine.  We also re-unify the dev and prod builds to use the same dockerfile.

@glacials if you can please review this PR, the code deploy project called out, and the beta site briefly to make sure the frontend looks correct.  I double checked the JS process.env calls and those look to be good, but I don't know if there are more gotchas in there.


BatedUrGonnaDie:
Hmm, on the last test it looks like the gem caching portion is still flakey.  I've disabled it for the time being so it's not blocking, so builds will be ~2 minutes slower.  I plan to take another look at the dependency caching this weekend to hopefully get gem caching reliable and see if I can get node modules working at all.

BatedUrGonnaDie:
Some failure reasons: If for some reason it can't connect to ECR (credential issue, connection issue), no image marked as latest, repo doesn't exist.

If there is an error docker will just build the image from scratch.  In fact the first time we run each of these build stages this command fails as there is no image in that repo, so it will build it from scratch and then as part of `post_build` it will tag and upload it for future runs.

glacials:
Sounds good to me, thanks for explaining!

PR #699, opened 2020-06-06
Mass dependency update

BatedUrGonnaDie:
This should take care of most of the dependabot updates, as well as updating rails to 6.0.3 which cleans up 95% of the depreciation notices and rspec to v4 proper.


BatedUrGonnaDie:
@glacials build is failing due to strong migrations complaining about an old migration.  Do you think it's ok to either wrap the offending statements in `safety_assured` or bump the migration starting point up to the latest one?  The one it's complaining about the solution is to break it into multiple migrations which I weary to do now that it's already live.

BatedUrGonnaDie:
Ended up just bumping up the strong migrations starting point.  Like you said, since we're really just trying to prevent issues in production it doesn't make sense to keep inspecting all of those old migrations.  And as they add new rules and triggers I feel this would crop up again with some of the older ones.  Maybe we should periodically update this in the future as well when we have new migrations.

PR #708, opened 2020-07-11
Fix Twitch Player Issues

### LINKED ISSUE #707: Timeline bar and seeking by segment is broken for Twitch videos
This was discussed briefly in Discord, so this is just to track the issue. For whatever reason, JS calls involving the player object created for a Twitch video don't work, so seeking and the progress bar don't function.

If you manually create a player in the JS console and use that, everything works as expected. So, it doesn't appear the Twitch JS API has changed.

moorecp:
Closes #707 

So, this probably isn't the permanent fix, but it does seem to work and might give some insight as to the real problem.


moorecp:
I can't get that event to fire unless I leave in the `setTimeout`. I was doing some searching around and [someone](https://discuss.dev.twitch.tv/t/twitch-player-events-wont-fire-with-vue/18234) was having some Vue related issues with things not working quite right. I don't have any Vue experience, but I tried, to no avail, to do some stuff in the `application.js` pack to see if that would make a difference. It's not like the current Twitch stuff is in a Vue component, so I don't see the virtual DOM affecting it, but I'm not sure.

glacials:
Alright, thanks for trying. I'll accept this for now and make an issue to revisit.

PR #711, opened 2020-07-16
Address YouTube and Turbolinks issues

### LINKED ISSUE #710: Videos wonky after pressing back button
"when going back to a run page via the back button, the seek via click on the splits bar doesn't work"

Reported from this Discord message: https://discordapp.com/channels/361957211318124555/535862834261655593/733219640796381216

moorecp:
Closes #710 

This fixes the issue where we were having to poll to see if YouTube was ready due to Turbolinks issues. Additionally, it addresses #710 and fixes problems where a YouTube player wouldn't be seen as attached to the DOM when not performing a full page load.

### DIFF
--- a/app/javascript/youtube.js
+++ b/app/javascript/youtube.js
@@ -1,11 +1,6 @@
 const async = require('async')
 
 window.onYouTubeIframeAPIReady = () => {
-  setupYouTubeVideos()
-  return true
-}
-
-function setupYouTubeVideos() {
   Array.from(document.querySelectorAll('[data-youtube]')).forEach(el => {
     if (el.dataset.loaded === '1') {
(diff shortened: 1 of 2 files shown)


app/javascript/youtube.js
```diff
@@ -1,6 +1,7 @@
 const async = require('async')
 
 window.onYouTubeIframeAPIReady = () => {
+  setupYouTubeVideos()
   return true
 }
```
glacials, line 6: Are we able to collapse `onYouTubeIframeAPIReady` and `setupYouTubeVideos` into a single function?

moorecp: I don't see why not.

PR #712, opened 2020-07-16
Properly scale the video progress bar on the timelines for comparisons

moorecp:
If you are comparing 2 runs and the base run is shorter than the other run, the base run video progress bar won't scale properly. When you click on a segment in the shorter (faster) run, the progress bar should go to the beginning of the segment. The actual video goes to the correct place, but the bar appears ahead of where it should due to scaling off of its duration and not the duration of the longer run.

output

### DIFF
--- a/app/controllers/runs_controller.rb
+++ b/app/controllers/runs_controller.rb
@@ -133,7 +133,7 @@ class RunsController < ApplicationController
                         {id: @run.user.id, name: @run.user.name}
                       end
 
-    gon.scale_to = [@run.duration(timing).to_ms, @compare_runs.map { |run| run.duration(timing).to_ms }].flatten.max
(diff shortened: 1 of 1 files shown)


app/controllers/runs_controller.rb
```diff
@@ -133,7 +133,7 @@ def set_run
                         {id: @run.user.id, name: @run.user.name}
                       end
 
-    gon.scale_to = @run.duration_ms(timing)
+    gon.scale_to = [@run.duration(timing).to_ms, @compare_runs.map { |run| run.duration(timing).to_ms }].flatten.max
```
glacials, line 136: I'd prefer the splat operator over building an array that needs to be `flatten`ed:
```ruby
gon.scale_to = [@run.duration(timing).to_ms, *@compare_runs.map { |run| run.duration(timing).to_ms }].max
```

moorecp: Sounds good.

PR #718, opened 2020-07-27
Fix duplication when filtering by nonempty categories

MichaelJBerk:
In the current implementation, filtering by nonempty returns duplicate categories for each run associated with it. By joining the runs from the category's game, it will not return these duplicates, and each category should only appear once

### DIFF
--- a/app/models/category.rb
+++ b/app/models/category.rb
@@ -15,7 +15,7 @@ class Category < ApplicationRecord
 
   validates :name, presence: true
 
-  scope :nonempty, -> {joins(:runs)}
+  scope :nonempty, -> {joins(:game).group('categories.id').joins(:runs)}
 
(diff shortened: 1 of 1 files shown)


app/models/category.rb
```diff
@@ -15,7 +15,7 @@ class Category < ApplicationRecord
 
   validates :name, presence: true
 
-  scope :nonempty, -> {joins(:runs)}
+  scope :nonempty, -> {joins(:game).group('categories.id').joins(:runs)}
```
glacials, line 18: Would a `joins(:runs).distinct` suffice? Just a bit simpler and more efficient if so.

glacials: `:runs`, not `:game` :P Think we lost our original goal

Issue #721, opened 2020-08-18
Authorization URL does not support Google login


MichaelJBerk:
The URL for granting access to a third-party application (https://splits.io/oauth/authorize) redirects the user to the Twitch login page, if they are not already signed in to Splits.io. If a user uses their Google account to log in (and does not have a Twitch account connected), then they will not be able to grant access to a third-party application unless they have already logged in to Splits.io.

It would be best if the auth URL took a logged-out use to a page giving them the choice for how to sign in.

glacials:
I agree, I think this is an artifact from a time we only supported Twitch. I think our current sign in modal could just be extracted to a partial and reused in a page. I'd gladly accept a PR for this, otherwise I can get around to it eventually. Thanks for the report!

PR #722, opened 2020-08-19
Add support for authorizing applications via Google

MichaelJBerk:
If the user is not currently signed in, the authorization URL no longer redirects to the Twitch login page (see #721).
The user is now presented with a page where they can choose how they would like to sign in. This page is located at `/loginpage`, and will redirect to the homepage unless the user is trying to authorize an application.


app/views/pages/login.slim
```diff
@@ -0,0 +1,15 @@
+- if !cookies['return_to'] 
+	head
+		meta http-equiv="refresh" content="0; URL='/'"
```
glacials, line 3: What's the purpose of this redirect? We may not be explicitly sending people here without a `return_to` cookie, but afaict we can still handle them fine if they happen to arrive without one for whatever reason. I personally sometimes type sign-in URLs directly into my address bar, or bookmark them.

config/routes.rb
```diff
@@ -36,6 +36,7 @@
   get '/privacy',   to: 'pages#privacy_policy', as: :privacy_policy
   get '/partners',  to: 'pages#partners',       as: :partners
   get '/hot',       to: 'pages#hotsauce',       as: :hotsauce
+  get '/loginpage', to: 'pages#login',          as: :login
```
glacials, line 39: This is my bad for not writing this down anywhere, but we canonically use the term "sign in" over "log in". I'd also add that having "page" in the path is redundant, so I think `/signin` and `:signin` would be the best fit.

app/views/pages/login.slim
MichaelJBerk: I hear what you're saying, and I agree that it makes more sense for the page to be accessible even if the user isn't currently authorizing an application. Should we still redirect a user who's already signed in, or display some other content?

glacials: Sure, signing in while signed in is unexpected so we can redirect if that's the case. Instead of a meta tag though, the controller tends to be the best place for redirects. In this case `app/controllers/pages_controller.rb` would be changed to something like
```ruby
class PagesController < ApplicationController
  def login # or whatever the "to: pages#login" part ends up being in config/routes.rb
    redirect_to(root_path) if current_user.present?
  end
end
```

PR #727, opened 2020-09-13
Add username/password sign in

glacials:
Adds username/password sign in, as a third option after Twitch and Google sign-in. Finally!

This is a simple top-level goal, but required overall more code on supporting features:

1. By allowing linkless accounts we need to implement a forgot password flow (meaning we need to implement our first mailers), which is most of this branch.
2. We also need to add a canonical `user.email` field and allow users to change it, as opposed to relying on `user.twitch.email` and `user.google.email`.

There are still more supporting features this feature begs for like 2FA, CAPTCHA, Gravatar, and general security/anti-spam/anti-fraud that Twitch and Google used to give us for free, but I didn't include these on this first pass.

I tried this branch in two other ways, one that completely replaced our auth with Devise, and one that tried to string our existing auth code up to Devise and use the missing parts it had that we didn't, but it ended up being overall a lesser headache to implement these things ourselves with the help of some vanilla Rails constructs like `has_secure_password` and mailers, which have gotten pretty good recently.

There are a couple of changes piggybacked on this branch as well:

1. Cleaned up many "Speedrun.com" spellings to "speedrun.com", which their branding prefers
2. Started moving a couple of routes to vanilla Rails `resources` definitions
3. Made WSL support in the Makefile a little better (cc @jamacanbacn)

_This is live on beta so the emailing part can be tried out._


app/views/settings/_avatar.slim
```diff
@@ -1,10 +1,15 @@
...
+      ' back in to refresh it.
+    - elsif current_user.google
+      ' Your avatar is linked to your Google account and can be updated
+      ' there. After any changes, you may need to log out of Splits.io and then
+      ' back in to refresh it.
     - else
-      ' Google
-    ' account and can be updated there. After any changes, you may need to log out of Splits.io and then back in to
-    ' refresh it.
+      / TODO: Add Gravatar support
```
BatedUrGonnaDie, line 13: Have you tested out how the rest of the site will handle null avatars?

config/environments/development.rb
```diff
@@ -33,7 +33,7 @@
 
   # Don't care if the mailer can't send.
   config.action_mailer.raise_delivery_errors = false
-
+  config.action_mailer.delivery_method = :file
```
BatedUrGonnaDie, line 36: Just in case can we add this to `profiling.rb` as well?  It inherits from production normally which would default to `ses`.

BatedUrGonnaDie (approved):
Just some minor questions.  Overall looks like a really good change!

app/mailers/application_mailer.rb
```diff
@@ -1,2 +1,4 @@
 class ApplicationMailer < ActionMailer::Base
+    default from: 'hi@splits.io'
```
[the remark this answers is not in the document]
glacials: It is! I'm a little worried about spam/unactionable/accidental replies, but I figured let's roll the dice and see, and change it later if it's bad.

app/models/user.rb
```diff
@@ -41,9 +41,23 @@ class User < ApplicationRecord
...
+  #
+  # Although it's theoretically bad to have max lengths on passwords, BCrypt
+  # truncates at 72 bytes, so we should enforce that here rather than let users
+  # pass in 73 bytes and be surprised that we only use the first 72.
+  has_secure_password(validations: false)
+  validates_confirmation_of :password, allow_blank: true
+  validates_length_of       :password, maximum: ActiveModel::SecurePassword::MAX_PASSWORD_LENGTH_ALLOWED
 
   validates :name, presence: true, uniqueness: true, format: {with: NAME_REGEX}
+  validates :email, presence: true, uniqueness: true # TODO: Add a not-null to database
```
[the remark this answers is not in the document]
glacials: Yeah, it's because I wanted to deploy this branch to beta to test mailers, but when it's deployed to beta migrations are run against the production database. So if this column was marked not null in the production database before the branch was deployed to production, production rails would hit errors about trying to insert nulls.

app/views/settings/_avatar.slim
glacials: I'm fairly confident Twitch can (used to?) report null avatars already, so I believe the code already knows how to handle it. But yeah I've done a little manual testing on this branch to be sure, it displays a fallback.

app/models/user.rb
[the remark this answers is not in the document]
BatedUrGonnaDie: Makes perfect sense, thanks for the explanation.

app/views/settings/_avatar.slim
BatedUrGonnaDie: Sounds good, just wanted to make sure random pages wouldn't start throwing 500's.

PR #730, opened 2020-11-01
Split off friendly race UUIDs into its own gem

glacials:
Recently I've been wanting to split some of our more general-purpose code into gems so others can benefit from the work, and this one was a pretty straightforward candidate.

This splits off the logic that shortens race UUIDs into reversible, stateless slugs for use in URLs. It's now a gem that can be used in any Rails application on any model that has UUID primary keys. It is located at [glacials/friendly_uuid](https://github.com/glacials/friendly_uuid).

Another split I might do later is our `Duration` implementation, as it's more precise than `ActiveSupport::Duration` and has been battle tested.

There are some formatting changes included here because I'm thinking of switching us over to [ruby-formatter/rufo](https://github.com/ruby-formatter/rufo) so I'm letting it leak some changes in as I test it out.

### DIFF
--- a/app/controllers/races_controller.rb
+++ b/app/controllers/races_controller.rb
@@ -1,9 +1,9 @@
 class RacesController < ApplicationController
-  before_action :set_races,        only: [:index]
-  before_action :set_race,         only: [:show, :update]
+  before_action :set_races, only: [:index]
+  before_action :set_race, only: [:show, :update]
   before_action :check_permission, only: [:show]
-  before_action :set_race_gon,     only: [:show]
-  before_action :shorten_url,      only: [:show]
+  before_action :set_race_gon, only: [:show]
+  before_action :shorten_url, only: [:show]
 
   def index
   end
@@ -33,7 +33,7 @@ class RacesController < ApplicationController
 
   def check_permission
     return unless @race.secret_visibility?
-    return if     @race.joinable?(token: race_params[:join_token], user: current_user)
+    return if @race.joinable?(token: race_params[:join_token], user: current_user)
 
     render :unauthorized, status: :unauthorized
   end
@@ -43,13 +43,13 @@ class RacesController < ApplicationController
 
     token = @race.entries.find_for(current_user).present? ? @race.join_token : nil
     gon.race = {
-      id:         @race.id,
-      join_token: token || race_params[:join_token]
+      id: @race.id,
+      join_token: token || race_params[:join_token],
     }
   end
 
   def set_race
-    @race = Race.friendly_find!(params[:id])
+    @race = Race.find(params[:id])
   rescue ActiveRecord::RecordNotFound
     render "application/not_found"
   end
--- a/Gemfile
+++ b/Gemfile
@@ -71,15 +71,16 @@ gem 'cancancan'
 gem 'doorkeeper'
 
 # db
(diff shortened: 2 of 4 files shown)


app/controllers/races_controller.rb
```diff
@@ -43,13 +43,13 @@ def set_race_gon
...
-      id:         @race.id,
-      join_token: token || race_params[:join_token]
+      id: @race.id,
+      join_token: token || race_params[:join_token],
     }
   end
 
   def set_race
-    @race = Race.friendly_find!(params[:id])
+    @race = Race.find(params[:id])
```
BatedUrGonnaDie, line 52: 2 concerns that I have with overloading the base `find` and `find_by` methods that may not be substantiated.

1. How does this play compatibility wise with other gems that may modify these methods?
2. Is there ever a situation where we would want to call the base `find` or `find_by` methods instead of the friendly version?

And as a side note, does it also cover `find_by!` as well?  I took a look at the gem and didn't see it but I'm not sure how it works in Rails to know if just overloading `find_by` is enough to have it implemented for free.

BatedUrGonnaDie (commented):
2 1/2 questions for you on this.

app/controllers/races_controller.rb
glacials: 1. I did not think to test compatibility with other gems that override these methods, but it appears to work great!
    ```ruby
    # test.rb
    class ApplicationModel
      def self.find(id)
        "[base record '#{id}']"
      end
    end

    module FriendlyUUID
      def find(id)
        "#{super(id)} via FriendlyUUID"
      end
    end

    module AnotherGem
      def find(id)
        "#{super(id)} via AnotherGem"
      end
    end

    class User < ApplicationModel
      extend FriendlyUUID
      extend AnotherGem
    end

    puts User.find("test") # [base record 'test'] via FriendlyUUID via AnotherGem
    ```

2. I can't think of a use case where you'd want to access the original functions, but let me know if I'm missing something. Both of them are logically lightweight, and become noops when given a full ID. The only reason you might not want to use one is efficiency, but I can't imagine someone both wanting that efficiency _and_ wanting the gem to be installed, as the inefficiency is baked in. If I'm missing a use case though, happy to refactor.

2.5. We don't explicitly define a `find_by!`, but I tested it manually and `find_by!` works perfectly in races. **But** looking at the source of [`find_by!`](https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-find_by-21) it does **not** call out to `find_by` 🤔 Maybe a version difference? I've added https://github.com/glacials/friendly_uuid/issues/2 on the gem to investigate more, since if I add an override to `find_by!` "just to be sure" and it turns out to call out to `find_by` in  the end, we could end up running unnecessary queries from calling `expand` twice.

BatedUrGonnaDie: Ah, I didn't realize they became noops when given full ID's.  Then just throw that one away since I didn't have a specific scenario in mind, more just thinking about the future.

2.5 As long as calling `find_by!` will work correctly, then I'm ok to merge this in.  My concern was someone coming along to create a PR and potentially getting hit with differences in the lookups that they are not expecting.

BatedUrGonnaDie (approved):

PR #733, opened 2020-11-14
Use timesync instead of new Date() built in for count down/up

BatedUrGonnaDie:
The default behavior of moment for object creation is to use new Date() to create the timestamp [[1]](https://momentjs.com/docs/#/parsing/now/), which pulls from the system clock.  If 2 computers are out of sync then the count down/up would be out of sync.  Switch over to calling `moment(ts.now)` which provides moment with a unix timestamp of the server's time, which is synced across all clients [[2]](https://momentjs.com/docs/#/parsing/unix-timestamp-milliseconds/).

I wasn't able to verify locally that this will fix the problem since I couldn't get the 2 devices to connect properly, but I'm fairly confident that this should fix the issue reported.

### DIFF
--- a/app/javascript/count.js
+++ b/app/javascript/count.js
@@ -2,6 +2,8 @@ import {setDriftlessInterval} from 'driftless'
 const moment = require('moment')
 require('moment-duration-format')(moment)
 
+import {ts} from 'time.js'
+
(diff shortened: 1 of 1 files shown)


glacials (approved):
Nice find!

PR #737, opened 2020-11-23
Allow trials without credit cards

### LINKED ISSUE #732: Allow starting a trial without entering a credit card number
We use Stripe's built-in trial feature right now, which requires a credit card number to start. We don't want to require this, so we should build our own on-site trial separate from this that allows a trial to be started manually via button-press, and does not require a credit card number or any Stripe interaction to start.

glacials:
Currently trial subscriptions are handled completely by Stripe, and they don't have an option to allow them without entering payment information first.

This branch moves trial subscriptions to be handled by us, so that we don't need to have that requirement in place.

It also refactors the code that decides which users have which features, because it was getting pretty muddy having so many features with so many possible grantors of that feature. Now, each model that can grant a feature (Subscription, PatreonUser, and now SubscriptionTrial) has the same interface and can be easily queried en masse by User.

Also letting some more [`rufo`][1] lint leak through. I'm pretty confident I want us to switch to it.

Fixes #732.

[1]: https://github.com/ruby-formatter/rufo

### DIFF
--- a/app/models/patreon_user.rb
+++ b/app/models/patreon_user.rb
@@ -6,7 +6,7 @@ class PatreonUser < ApplicationRecord
   #    (grandfathered)
   # 2. Anyone paying at least the price of Gold via Patreon
   def has_predictions?
-    patron?(tier: 2, before: STRIPE_MIGRATION_DATE) || patron?(tier: 4)
+    active?(tier: 2, since_before: STRIPE_MIGRATION_DATE) || active?(tier: 4)
   end
 
   # Redirectors used to be a tier 3 Patreon feature, so they're available to:
@@ -14,7 +14,7 @@ class PatreonUser < ApplicationRecord
   #    (grandfathered)
   # 2. Anyone paying at least the price of Gold via Patreon
   def has_redirectors?
-    patron?(tier: 3, before: STRIPE_MIGRATION_DATE) || patron?(tier: 4)
+    active?(tier: 3, since_before: STRIPE_MIGRATION_DATE) || active?(tier: 4)
   end
 
   # Advanced comparisons used to be a tier 3 Patreon feature, so they're
@@ -22,35 +22,35 @@ class PatreonUser < ApplicationRecord
   # 1. Early patrons (grandfathered as a thank-you)
   # 2. Anyone paying at least the price of Gold via Patreon
   def has_advanced_comparisons?
-    patron?(tier: 3, before: STRIPE_MIGRATION_DATE) || patron?(tier: 4)
+    active?(tier: 3, since_before: STRIPE_MIGRATION_DATE) || active?(tier: 4)
   end
 
   # Sum-of-best leaderboards used to be free, so they're available to:
   # 1. Early patrons (grandfathered as a thank-you)
   # 2. Anyone paying at least the price of Gold via Patreon
   def has_sum_of_best_leaderboards?
-    patron?(tier: 1, before: STRIPE_MIGRATION_DATE) || patron?(tier: 4)
+    active?(tier: 1, since_before: STRIPE_MIGRATION_DATE) || active?(tier: 4)
   end
 
   # Hiding used to be free, so it's available to:
   # 1. Early patrons (grandfathered as a thank-you)
   # 2. Anyone paying at least the price of Gold via Patreon
   def has_hiding?
-    patron?(tier: 1, before: STRIPE_MIGRATION_DATE) || patron?(tier: 4)
+    active?(tier: 1, since_before: STRIPE_MIGRATION_DATE) || active?(tier: 4)
   end
 
   # Advanced video support used to be free, so it's available to:
   # 1. Early patrons (grandfathered as a thank-you)
   # 2. Anyone paying at least the price of Gold via Patreon
   def has_advanced_video?
-    patron?(tier: 1, before: STRIPE_MIGRATION_DATE) || patron?(tier: 4)
+    active?(tier: 1, since_before: STRIPE_MIGRATION_DATE) || active?(tier: 4)
   end
 
   # Auto-highlighting used to be free, so it's available to:
   # 1. Early patrons (grandfathered as a thank-you)
   # 2. Anyone paying at least the price of Gold via Patreon
   def has_autohighlight?
-    patron?(tier: 1, before: STRIPE_MIGRATION_DATE) || patron?(tier: 4)
+    active?(tier: 1, since_before: STRIPE_MIGRATION_DATE) || active?(tier: 4)
   end
 
   # Advanced analytics used to be lower-tier paid Patreon features, so they're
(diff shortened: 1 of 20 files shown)


app/models/patreon_user.rb
```diff
@@ -1,3 +1,92 @@
...
+  # equivalent at Patreon.
+  def has_srdc_submit?
+    patron?(tier: 4)
+  end
+
+  # patron? returns something truthy if the user has been a patron at or above
+  # the given tier since the given `before` time. Not passing a `before` is the
+  # same as not caring how long the user has been a patron. Tier 4 is an
+  # imaginary tier equal to the price of Gold.
+  def patron?(tier: 0, before: Time.now.utc)
```
BatedUrGonnaDie, line 76: Naming wise this seems a bit confusing to me.  Although this is more a nitpick than anything else.

app/models/subscription_trial.rb
```diff
@@ -0,0 +1,51 @@
+class SubscriptionTrial < ApplicationRecord
+  SUBSCRIPTION_LENGTH = 14.days
+
+  belongs_to :user
+
+  validates :user_id, presence: true, uniqueness: true
+
+  def ended_at
+    created_at + SUBSCRIPTION_LENGTH
```
BatedUrGonnaDie, line 9: Since this is going to be completely static maybe we should store this in the DB instead of calculating it every time?

app/models/user.rb
```diff
@@ -117,102 +112,64 @@ def to_s
...
+  # feature_grantors returns an array of objects that can give the user
+  # features, like a subscription, a grandfathered patreon subscription, or a
+  # trial. Every feature grantor must implement all `has_FEATURE?`-type methods
+  # below.
+  #
+  # If the feature grantor resolves to nil, it is assumed that it grants no
+  # features.
+  def feature_grantors
+    # TODO: Convert subscriptions from has-many to has-one
+    [patreon, subscriptions, trial].compact
```
BatedUrGonnaDie, line 124: Is there any way to preload this so we're not making potentially 3 calls anytime we want to check features?

app/views/layouts/application.slim
```diff
@@ -127,22 +127,16 @@ html lang='en'
               => icon('fas', 'flag-checkered')
               | Return to race
         header#header.m-1.p-0 = yield(:header)
+        - if current_user && current_user.email.nil?
+          .alert.alert-danger
```
BatedUrGonnaDie, line 131: I feel like the majority of users do not have a splits.io backed account, and this might be alarming with the current styling for something that was how the site was designed just a few weeks (months?) ago.

app/views/layouts/application.slim
```diff
@@ -127,22 +127,16 @@ html lang='en'
...
+            ' You don't have an email attached to your Splits.io account. If
+            ' you become locked out your account will be irrecoverable. You
+            ' will also experience bugs around the site. Please add an email
```
BatedUrGonnaDie, line 134: I think blatantly stating that you will experience bugs if you don't have a splits.io backed account is pretty strong language.  Maybe can tone this down a bit?

BatedUrGonnaDie (changes requested):
I'm not sure if I'm missing something, but I see the schema.rb file is changed but I don't see any migration file as part of this PR.

app/models/subscription_trial.rb
glacials: I disagree here actually, for the same reason we calculate (rather than store) `Segment#realtime_end_ms`/`Segment#gametime_end_ms` -- it guarantees data integrity because we only have one source of data truth. We can never have a segment end that doesn't make sense, and if we do anyway because of a logic bug, when we fix it we are not left over with a bunch of incorrect data we need to fix in a migration. We are trading some computation time for some data guarantees.

In the case of trials it's an immeasurably small computation and a small-ish data guarantee, but I think the principle still holds.

I can see two advantages of storing this in the database separately:

1. if we decide to later change the subsciption trial length, it would only change for new trials, which might be preferable depending on the change.
2. if we have a special situation where an admin wants to extend someone's trial (or perhaps make it indefinite), they can edit `ended_at` which "makes sense", rather than edit `created_at` which, while it would work, wouldn't really "make sense", or they would have to add a new feature grantor for an admin-granted free subscription or somesuch.

But IMO they are not big enough advantages.

app/models/user.rb
glacials: I've tried, but it seems that Rails keeps optimizing the queries to be separate even with `includes`:

```rails
> User.includes(:patreon, :trial, :subscriptions).first.feature_grantors
  User Load (0.4ms)  SELECT "users".* FROM "users" ORDER BY "users"."id" ASC LIMIT $1  [["LIMIT", 1]]
  PatreonUser Load (0.3ms)  SELECT "patreon_users".* FROM "patreon_users" WHERE "patreon_users"."user_id" = $1  [["user_id", 1]]
  SubscriptionTrial Load (0.3ms)  SELECT "subscription_trials".* FROM "subscription_trials" WHERE "subscription_trials"."user_id" = $1  [["user_id", 1]]
  Subscription Load (0.4ms)  SELECT "subscriptions".* FROM "subscriptions" WHERE "subscriptions"."user_id" = $1  [["user_id", 1]]
```

I'm not super torn up about it, it's a maximum of 3 extra queries per page.

app/views/layouts/application.slim
glacials: Very fair. I'll remove this part.

app/views/layouts/application.slim
glacials: This is actually not Splits.io-backed vs. non-Splits.io-backed, as when we added username/password sign in in #727 we migrated `TwitchUser` / `GoogleUser` emails to `User`. So this will only include users who had a null Twitch and Google email. AFAIK it's impossible to have a null Google email, so this was effectively only for users with null Twitch emails and no Google link. Although you have needed an email address to sign up for Twitch for years, the Twitch API will return a null email if you have not verified your email with them.

I checked the production database and about 14% of users (~6k) have null emails and will trigger this warning. I can still tone down this styling since that's not nothing.

glacials (commented):
Oops, forgot to `git add` the migration and the controller 😅

app/models/subscription_trial.rb
BatedUrGonnaDie: Makes perfect sense when I think about it that way, thanks for the explanation!

app/models/user.rb
BatedUrGonnaDie: Huh, that's weird.  I agree since it's a constant amount of queries (and not likely to grow) this isn't a big deal.

app/views/layouts/application.slim
BatedUrGonnaDie: Oh, didn't realize this was covering a different use case.  It's less of a issue then in my mind, but that is still a large # of users so thanks for making that change.

BatedUrGonnaDie (approved):
Looks good once that one test is fixed.

PR #738, opened 2020-11-29
Periodically sync with speedrun.com

### LINKED ISSUE #690: Automatically sync games to speedrun.com
The speedrun.com game sync is much improved now, but doesn't happen automatically anymore. We should run it occasionally to get new games.

glacials:
Our sync job with speedrun.com games needs to be kicked off manually atm. This is leading to users being confused about why their new game isn't present on Splits.io.

This branch calls the job automatically once per 7 days. Fixes #690.


config/initializers/scheduler.rb
```diff
@@ -0,0 +1,8 @@
+require 'rake'
+require 'rufus-scheduler'
+
+Rufus::Scheduler.singleton.every '7d' do
```
BatedUrGonnaDie, line 4: How does this work for multiple server environments?  I've looked at this in the past and my understanding was that every 7 days, every server that is alive would then run this job.

Also we should probably add the guard for console/rake tasks https://github.com/jmettraux/rufus-scheduler#avoid-scheduling-when-running-the-ruby-on-rails-console

glacials: Nice catch, added a tiny bit of duct tape to use Redis as a sort of locking mechanism.

BatedUrGonnaDie (approved):

PR #747, opened 2020-12-23
Update show.slim

Hey-Im-Drek:
On the races page, the embeded twitch streams say
h58jteScGB
Adding '&parent=splits.io' to the end of the embed link fixes the issue.


glacials (approved):
Thanks for the fix!

PR #776, opened 2021-07-15
Added 'uses_autosplitter' to run API

bigfoott:
Followed guidance from [this message](https://i.imgur.com/MFDMcie.png) in Discord.

I was unable to test fully as I was having issues getting Docker to work, although as stated in the message above it should work regardless.

One concern I have is that I'm not sure if this field can return null under normal curcumstances. In the half-working version of the site I was running in Docker, I noticed that the field returned `null` for the generated run, but I wasn't sure if that was a special/weird case.


glacials (changes requested):
Nice job getting everything running! This field can indeed return null -- the [code that parses runs](https://github.com/glacials/splits-io/blob/a28b42fe7d95cafe860d6914df6d962933e1e0a5/app/models/concerns/unparsed_run.rb#L53) never can, but when we [added the field](https://github.com/glacials/splits-io/blob/main/db/migrate/20200207203633_add_uses_autosplitter_to_runs.rb) we never migrated old runs or set a default, so many runs do have null values for this.

So if we can update the API schema and the documentation to say null is possible, then this will be perfect 👍🏻

bigfoott:
I wasn't able to test these changes locally as I had some issues with Docker when trying to upload a test run that used an autosplitter, but they should be correct.

glacials (approved):
Seems good, thanks for addressing the feedback!

BatedUrGonnaDie (approved):

PR #780, opened 2021-08-15
Exclude timesync and SRL endpoints from tracking

BatedUrGonnaDie:
Each time these are called we're creating a job to track them, but these are called automatically during loading of other pages so that tracking information is not really all that useful.


glacials (approved):
Nice catch, LGTM!

PR #857, opened 2024-10-17
Upgrade to Ruby 2.7.8 → 3.0.7, Rails 6.0 → 6.1

glacials:
Had a tangled mess of a mass update to Rails 7, but had no confidence in its correctness, so taking it slow making incremental updates instead.

This branch's goal is only to update Ruby, but it seems Rails needs a bump to reasonably support that as well. There were a few explicit updates required along with that, and the rest of the gem updates were caused (so required) directly by the `bundle update rails ...` command.

The migrations were generated by Authie, which required them for its own upgrade.


BatedUrGonnaDie:
I don't see it mentioned, but did you run the Rails specific [update task](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#moving-between-versions) and then make any changes we would need to based on the [6.0 -> 6.1 breaking changes](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-6-0-to-rails-6-1)?

glacials:
Oh thanks, I didn't know about that! Have pushed those updates, and some test fixes.

config/application.rb
```diff
@@ -8,6 +8,14 @@
...
+    # Configuration for the application, engines, and railties goes here.
+    #
+    # These settings can be overridden in specific environments using the files
+    # in config/environments, which are processed later.
+    #
+    # config.time_zone = "Central Time (US & Canada)"
+    # config.eager_load_paths << Rails.root.join("extras")
+
     # Initialize configuration defaults for originally generated Rails version.
     config.load_defaults 6.0
```
BatedUrGonnaDie, line 20: We should load the defaults for 6.1 now (see other comment).

config/initializers/new_framework_defaults_6_1.rb
```diff
@@ -0,0 +1,67 @@
+# Be sure to restart your server when you modify this file.
```
BatedUrGonnaDie, line 1: This file is only needed to enable 6.1 defaults one at a time. If possible can you test each one and then remove this and enable the 6.1 defaults?

BatedUrGonnaDie (approved):
If we wanna move to the 6.1 defaults, see my 2 comments. If not, LGTM send it.

config/initializers/new_framework_defaults_6_1.rb
glacials: Want to ship this and let it run for a bit first, so we have a smaller search space for any issues we hit. Then will swap over the defaults one by one!

